PR #6, opened 2026-02-09
Implement the Logger in the Deploy Command

logan-scully-arm:
## Changes

- Changes warning prints to warning logger entries.
- Returns these entries from `SetupExitCleanup`
- Printing warnings is now done entirely at the cobra/command level

Due to the way `Run` and `DryRun` work at the moment, they've maintained their current output method.

### DIFF
--- a/cmd/topo/deploy.go
+++ b/cmd/topo/deploy.go
@@ -97,7 +97,7 @@ Use --dry-run to see what commands would be executed without actually running th
 		}
 
 		deployment, cleanup := docker.NewDeployment(composeFile, deployOpts)
-		stop := goperation.SetupExitCleanup(os.Stderr, cleanup, os.Exit)
+		stop := goperation.SetupExitCleanup(os.Stdout, cleanup, os.Exit)
 
 		defer func() {
 			entries := stop()
@@ -105,9 +105,9 @@ Use --dry-run to see what commands would be executed without actually running th
 		}()
 
 		if deployDryRun {
-			return deployment.DryRun(os.Stderr)
+			return deployment.DryRun(os.Stdout)
 		}
-		return deployment.Run(os.Stderr)
+		return deployment.Run(os.Stdout)
 	},
 }
 
--- a/cmd/topo/deploy.go
+++ b/cmd/topo/deploy.go
@@ -11,6 +11,9 @@ import (
 	"github.com/arm-debug/topo-cli/internal/deploy/docker/operation"
 	goperation "github.com/arm-debug/topo-cli/internal/deploy/operation"
 	checks "github.com/arm-debug/topo-cli/internal/deploy/project_checks"
+	"github.com/arm-debug/topo-cli/internal/output/console"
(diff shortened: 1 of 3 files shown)


internal/deploy/operation/operation.go
```diff
@@ -16,20 +18,25 @@ type Operation interface {
 }
 
 // SetupExitCleanup sets up a handler to run an operation once when the program exits due to an interrupt signal.
-func SetupExitCleanup(operation Operation, w io.Writer, exit func(int)) func() {
+func SetupExitCleanup(w io.Writer, operation Operation, exit func(int)) func() []logger.Entry {
```
yejseo01, line 21: Was there s reason for the order of the parameter change? or is this just unnecessary change?

cmd/topo/deploy.go
```diff
@@ -43,10 +46,14 @@ Use --dry-run to see what commands would be executed without actually running th
...
 		cmd.SilenceUsage = true
+		c := console.NewLogger(os.Stderr, term.Plain)
 
 		portChanged := cmd.Flags().Changed("port")
 		if portChanged && noRegistry {
-			_, _ = fmt.Fprintln(os.Stderr, "WARN: --port has no effect when --no-registry is set. Define SSH port in your SSH config instead.")
+			c.Log(logger.Entry{
+				Level:   logger.Warning,
+				Message: "--port has no effect when --no-registry is set. Define SSH port in your SSH config instead.",
+			})
```
awphi, line 56: Muuuuch better

cmd/topo/deploy.go
```diff
@@ -83,17 +90,24 @@ Use --dry-run to see what commands would be executed without actually running th
...
+			entries := stop()
+			c.Log(entries...)
+		}()
 
 		if deployDryRun {
-			return deployment.DryRun(os.Stdout)
+			return deployment.DryRun(os.Stderr)
 		}
-		return deployment.Run(os.Stdout)
+		return deployment.Run(os.Stderr)
```
awphi: Are we now printing deployment output to stderr? That doesn't seem right 🤔

awphi (commented):
Nice! Question on deploy print output - happy to approve once answered

internal/deploy/operation/operation.go
logan-scully-arm: There's no strong reason. Whenever I've noticed it, I've tried to move the logger or io.writer to be the first variable. Across the repo it was in various orders throughout.

cmd/topo/deploy.go
logan-scully-arm: I put this in err since the deployment output is just logging rather than a specific command result, but happy to change it back if you disagree.

awphi: I think for consistency with other commands we should stick to stdout even if it's not quite right

awphi (approved):

yejseo01 (approved):
LGTM

PR #7, opened 2026-02-09
feat: Implement --output flag in log

logan-scully-arm:
## Changes

- Moves the output to be var-less. Is now accessed through a consolidated resolve function
- This replaces defaulting to plain for functions without the output flag

### DIFF
--- a/cmd/topo/clone.go
+++ b/cmd/topo/clone.go
@@ -4,7 +4,6 @@ import (
 	"os"
 
 	"github.com/arm-debug/topo-cli/internal/arguments"
-	"github.com/arm-debug/topo-cli/internal/output/console"
 	"github.com/arm-debug/topo-cli/internal/output/term"
 	"github.com/arm-debug/topo-cli/internal/project"
 	"github.com/arm-debug/topo-cli/internal/template"
@@ -43,6 +42,10 @@ Some projects require build arguments. Supply them on the command line or answer
 `,
 	Args: cobra.MinimumNArgs(2),
 	RunE: func(cmd *cobra.Command, args []string) error {
+		c, err := GetLogger(cmd)
+		if err != nil {
+			return err
+		}
 		cmd.SilenceUsage = true
 		path := args[0]
 		src := args[1]
@@ -72,7 +75,6 @@ Some projects require build arguments. Supply them on the command line or answer
 
 		logs, err := project.Clone(path, projectSource, argProvider)
 
-		c := console.NewLogger(os.Stderr, term.Plain)
 		c.Log(logs...)
 		return err
 	},
--- a/cmd/topo/clone.go
+++ b/cmd/topo/clone.go
@@ -4,6 +4,7 @@ import (
 	"os"
 
 	"github.com/arm-debug/topo-cli/internal/arguments"
+	"github.com/arm-debug/topo-cli/internal/output/console"
 	"github.com/arm-debug/topo-cli/internal/output/term"
 	"github.com/arm-debug/topo-cli/internal/project"
 	"github.com/arm-debug/topo-cli/internal/template"
@@ -42,10 +43,11 @@ Some projects require build arguments. Supply them on the command line or answer
 `,
 	Args: cobra.MinimumNArgs(2),
 	RunE: func(cmd *cobra.Command, args []string) error {
-		c, err := GetLogger(cmd)
+		outputFormat, err := resolveOutput(cmd)
 		if err != nil {
 			return err
 		}
+		c := console.NewLogger(os.Stderr, outputFormat)
 		cmd.SilenceUsage = true
 		path := args[0]
 		src := args[1]
--- a/cmd/topo/extend.go
+++ b/cmd/topo/extend.go
@@ -49,6 +49,8 @@ or interactively when prompted:
 		composeFilePath := args[0]
 		sourceArg := args[1]
 
+		c, err := GetLogger(cmd)
+
 		src, err := template.NewSource(sourceArg)
 		if err != nil {
 			return err
@@ -74,7 +76,6 @@ or interactively when prompted:
 
 		logs, err := project.Extend(composeFilePath, src, argProvider)
 
-		c := console.NewLogger(os.Stderr, term.Plain)
 		c.Log(logs...)
 		return err
 	},
--- a/cmd/topo/extend.go
+++ b/cmd/topo/extend.go
@@ -4,7 +4,6 @@ import (
 	"os"
 
 	"github.com/arm-debug/topo-cli/internal/arguments"
-	"github.com/arm-debug/topo-cli/internal/output/console"
 	"github.com/arm-debug/topo-cli/internal/output/term"
 	"github.com/arm-debug/topo-cli/internal/project"
 	"github.com/arm-debug/topo-cli/internal/template"
@@ -50,6 +49,9 @@ or interactively when prompted:
 		sourceArg := args[1]
 
 		c, err := GetLogger(cmd)
+		if err != nil {
+			return err
+		}
 
 		src, err := template.NewSource(sourceArg)
 		if err != nil {
--- a/cmd/topo/extend.go
+++ b/cmd/topo/extend.go
@@ -4,6 +4,7 @@ import (
 	"os"
 
 	"github.com/arm-debug/topo-cli/internal/arguments"
+	"github.com/arm-debug/topo-cli/internal/output/console"
 	"github.com/arm-debug/topo-cli/internal/output/term"
 	"github.com/arm-debug/topo-cli/internal/project"
 	"github.com/arm-debug/topo-cli/internal/template"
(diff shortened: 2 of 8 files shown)


cmd/topo/clone.go
```diff
@@ -71,13 +73,21 @@ Some projects require build arguments. Supply them on the command line or answer
...
+		}
 
-		c := console.NewLogger(os.Stderr, term.Plain)
+		logFmt, err := resolveOutput(cloneLogOutFmt)
+		if err != nil {
+			return err
+		}
+		c := console.NewLogger(os.Stderr, logFmt)
 		c.Log(logs...)
 		return err
```
th3james, line 81: err can never be non-nil, we should clarify this is expected to be the non-error path
```suggestion
		return nil
```

cmd/topo/extend.go
```diff
@@ -73,13 +75,21 @@ or interactively when prompted:
...
+		}
 
-		c := console.NewLogger(os.Stderr, term.Plain)
+		logFmt, err := resolveOutput(extendLogOutFmt)
+		if err != nil {
+			return err
+		}
+		c := console.NewLogger(os.Stderr, logFmt)
 		c.Log(logs...)
 		return err
```
th3james, line 88: Same comment
```suggestion
		return nil
```

cmd/topo/clone.go
```diff
@@ -71,13 +73,21 @@ Some projects require build arguments. Supply them on the command line or answer
...
 
 		logs, err := project.Clone(path, projectSource, argProvider)
+		if err != nil {
+			return err
+		}
 
-		c := console.NewLogger(os.Stderr, term.Plain)
+		logFmt, err := resolveOutput(cloneLogOutFmt)
+		if err != nil {
+			return err
```
th3james: The error from `resolveOutput` doesn't indicate which flag is invalid, which is a problem because both `--output` and `--log-output` share the same function. We could add a `resolveLogOutput(cmd)` helper in `root.go` that reads the flag value directly from the command by name:

```go
func resolveLogOutput(cmd *cobra.Command) (term.Format, error) {
    v, err := cmd.Flags().GetString("log-output")
    if err != nil {
        return term.Plain, err
    }
    return resolveOutput("log-output", v)
}

func resolveOutput(flagName, flagValue string) (term.Format, error) {
    v := strings.TrimSpace(strings.ToLower(flagValue))
    switch v {
    case "plain":
        return term.Plain, nil
    case "json":
        return term.JSON, nil
    default:
        return term.Plain, fmt.Errorf("invalid --%s value %q: must be 'plain' or 'json'", flagName, flagValue)
    }
}
```

This removes the package-level `cloneLogOutFmt` / `extendLogOutFmt` / `deployLogOutFmt` vars, keeps the flag name as a single source of truth between registration and error reporting, and simplifies call sites to just `resolveLogOutput(cmd)`.

cmd/topo/main.go
```diff
@@ -8,7 +8,15 @@ import (
...
 )
 
+var logOutput string
+
 func main() {
+	rootCmd.PersistentFlags().StringVar(
+		&logOutput,
+		"log-output",
+		"plain",
+		"Log output: stderr, stdout, or file:/path",
```
tgonzalezorlandoarm: Should this be plain or json instead?

logan-scully-arm: Yeah it should, was copying these lines :sweat_smile: Thanks for spotting this

cmd/topo/main.go
```diff
@@ -8,7 +8,15 @@ import (
 	"github.com/arm-debug/topo-cli/internal/output/term"
 )
 
+var logOutput string
+
 func main() {
+	rootCmd.PersistentFlags().StringVar(
```
th3james: I believe it's more idiomatic to define this in the rootCmd package in an `init()` function. That keeps all the flag binding inside `root.go`

cmd/topo/main.go
```diff
@@ -8,7 +8,15 @@ import (
...
+		&logOutput,
```
th3james: You're reading from `GetString`
cmd/topo/root.go:71
```go
	v, err := cmd.Flags().GetString("log-output")
```
So this var isn't even used. IMHO `PersistentFlags().String` seems generally superior to using a package level var.

```suggestion
func main() {
	rootCmd.PersistentFlags().String(
```

cmd/topo/root.go
```diff
@@ -1,21 +1,42 @@
...
+var loggerCtxKey = loggerCtxKeyType{}
+
 var rootCmd = &cobra.Command{
 	Use:           "topo",
 	Short:         "Topo CLI",
 	Version:       fmt.Sprintf("%s (commit: %s)", version.Version, version.GitCommit),
 	SilenceErrors: true,
 	SilenceUsage:  true,
+
+	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
```
th3james: I gotta be honest, I do not love `PersistentPreRunE` - this is basically constructor inheritance, with all the problems that come with it. Most notably, if any sub commands implement `PersistentPreRunE`, the parent's is silently skipped - and unlike typical inheritance, Cobra doesn't have a super() mechanism. You'd have to manually call the parent's function yourself, which is error-prone and easy to forget.

My preference is for explicit rather than implicit, and we could consider a more functional approach to this problem:

```go
  func withLogger(fn func(*cobra.Command, []string, *console.Logger) error) func(*cobra.Command, []string) error {
      return func(cmd *cobra.Command, args []string) error {
          format, err := resolveLogOutput(cmd)
          if err != nil {
              return err
          }
          log := console.NewLogger(os.Stderr, format)
          return fn(cmd, args, log)
      }
  }
```

Then each command explicitly opts in:
```go
  var cloneCmd = &cobra.Command{
      Use:  "clone",
      RunE: withLogger(func(cmd *cobra.Command, args []string, c *console.Logger) error {
          // c is right here as a parameter
          ...
      }),
  }
```

IMHO, this makes it much easier to follow the flow of context building, in a way which magical inheritance chains obfuscate.

th3james (changes requested):
`withLogger` is optional, the unused var is a blocker

cmd/topo/root.go
logan-scully-arm: Okay that sounds good :+1: Might do it a bit differently, but will get rid of persistent pre-run

cmd/topo/main.go
logan-scully-arm: Having looked at examples of Cobra a bit more, It might be worth transitioning to `NewXCommand` implementations in the future -- I've put this in init for now

cmd/topo/root.go
```diff
@@ -38,7 +43,12 @@ func resolveTarget(flagValue string) (string, error) {
 	return "", fmt.Errorf("target not specified: provide --target or set TOPO_TARGET env var")
 }
 
-func resolveOutput(flagValue string) (term.Format, error) {
+func resolveOutput(cmd *cobra.Command) (term.Format, error) {
+	flagValue, err := cmd.Flags().GetString("output")
+	if err != nil {
+		return term.Plain, nil
+	}
```
th3james, line 50: This is a programmer error, not an expected situation. We should crash
```suggestion
	if err != nil {
		panic(fmt.Sprintf("bug: output flag not registered: %v", err))
	}
```

cmd/topo/main.go
```diff
@@ -10,8 +10,16 @@ import (
 
 func main() {
 	if err := rootCmd.Execute(); err != nil {
-		c := console.NewLogger(os.Stderr, term.Plain)
-		c.Log(logger.Entry{Level: logger.Err, Message: err.Error()})
+		outputFormat, e := resolveOutput(rootCmd)
+		if e != nil {
```
th3james: Nitpicking, but the potential for var shadowing keeps me awake at night
```suggestion
		outputFormat, outputFormatError := resolveOutput(rootCmd)
		if outputFormatError != nil {
```

th3james (approved):

cmd/topo/clone.go
```diff
@@ -43,6 +43,11 @@ Some projects require build arguments. Supply them on the command line or answer
 `,
 	Args: cobra.MinimumNArgs(2),
 	RunE: func(cmd *cobra.Command, args []string) error {
+		outputFormat, err := resolveOutput(cmd)
+		if err != nil {
+			return err
+		}
+		c := console.NewLogger(os.Stderr, outputFormat)
```
federicobozzini, line 50: I would wonder if the fact that this code is basically repeated vertabitm multiple times for different commands means it might be time to extract it into its own function.

logan-scully-arm: If I recall I discussed w/ @th3james about this. I think we came to the conclusion that there wasn't enough here for it to really make a benefit.

```
		outputFormat, err := resolveOutput(cmd)
		if err != nil {
			return err
		}
		c := console.NewLogger(os.Stderr, outputFormat)
```
vs

```
		c, err := GetNewLogger(cmd)
		if err != nil {
			return err
		}
```
The first version here is clearer for a new viewer of the code and is only 1 line longer.

federicobozzini: Fair enough. Let's keep an eye on it in case it grows bigger though.

PR #8, opened 2026-02-09
fix: always drops features prefix from list

awphi:
## Changes

- Spotted that `ProbeHardware` was not properly filtering the "Features" prefix from the relevant line in ` /proc/cpuinfo` as it's actually in the format `Features        : fpu asimd` rather than `Features: fpu asimd`. Fixed this and updated the test.
- Also moved `ExtractArmFeatures` to a method on `HardwareProfile` and gave the LUT it uses a clearer name


awphi:
Closing because I think this is going to be rendered obsolete by our choice to go with `lscpu` for HW checks

PR #9, opened 2026-02-09
feat: simple describe command

awphi:
## Changes

- Adds an extremely naive `topo describe` command which
  - Takes a `--target`
  - Runs a subset of the current HW-related health checks on it
  - Munges them into a struct
  - Writes that struct to a file in yaml, always overwriting what is already there
- Note this PR features no new tests as it features basically no new business logic - if you disagree lmk what you think needs explicit extra testing

There is still a fairly significant work to do in the form of:
- Adding CPU model + core count to the `host` description by replacing current probe with `lscpu` - done in #10 
- Adding accessible memory to the `host` description
- Refactoring to decouple the HW checks from the `health` package


Luke-Parkin (approved):

cmd/topo/describe.go
```diff
@@ -0,0 +1,65 @@
...
+		wd, err := os.Getwd()
+		if err != nil {
+			return err
+		}
+
+		outputFile := filepath.Join(wd, targetDescriptionFilename)
+		err = os.WriteFile(outputFile, reportBytes, 0o0644)
+		if err != nil {
+			return err
+		}
```
federicobozzini, line 44: Can we move this logic to a helper function?

internal/describe/describe.go
federicobozzini, line 1: I would add some tests for this new package.

cmd/topo/describe.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/arm-debug/topo-cli/internal/output/term"
+	"github.com/spf13/cobra"
+	"go.yaml.in/yaml/v4"
+)
+
+var describeTarget string
+
+const targetDescriptionFilename = "target-description.yaml"
+
+var describeCmd = &cobra.Command{
```
federicobozzini, line 18: If you add a new command you should also update the README and the changelog.

cmd/topo/describe.go
```diff
@@ -0,0 +1,65 @@
...
+	Use:   "describe",
+	Short: "Describe the hardware characteristics of the target host (CPU architecture, ISA features, etc.)",
```
federicobozzini: Technically so far it's only the CPU architecture and the ISA features. Nothing else.

internal/describe/describe.go
```diff
@@ -0,0 +1,47 @@
+package describe
+
+import (
+	"github.com/arm-debug/topo-cli/internal/health"
+	"github.com/arm-debug/topo-cli/internal/ssh"
+)
+
+type HostCPU struct {
+	Features []string
+	// TODO enrich with more details like CPU model and accessible memory
```
federicobozzini: I'm not a big fan of TODOs across the codebase.

internal/describe/describe.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+type RemoteProcCPU struct {
```
federicobozzini: Nit but it should be RemoteprocCPU. `remoteproc` is a standalone name.

cmd/topo/describe.go
```diff
@@ -0,0 +1,65 @@
...
+		if err != nil {
+			return err
+		}
+
+		report, err := describe.Generate(sshTarget)
+		if err != nil {
+			return err
+		}
+
+		reportBytes, err := yaml.Marshal(report)
```
federicobozzini: Unless I'm missing something, I think we would normally use a `Printer` here, with a `asYaml` method to define the implementation (that can just call `yarm.Marshal`).

internal/describe/describe.go
```diff
@@ -0,0 +1,47 @@
...
+	conn := health.NewConnection(sshTarget, ssh.ExecSSH)
+	hwProfile, err := conn.ProbeHardware()
+	if err != nil {
+		return TargetHardwareReport{}, err
+	}
+
+	return generateReport(hwProfile), nil
+}
+
+func generateRemoteProcReport(remoteCPUs []string) []RemoteProcCPU {
```
federicobozzini: Same observation here about how this should be `generateRemoteprocReport`.

internal/describe/describe.go
```diff
@@ -0,0 +1,47 @@
...
+	res := make([]RemoteProcCPU, len(remoteCPUs))
+	for i, cpu := range remoteCPUs {
+		res[i] = RemoteProcCPU{Name: cpu}
+	}
+	return res
+}
+
+func generateReport(hwProfile health.HardwareProfile) TargetHardwareReport {
+	return TargetHardwareReport{
+		Host: HostCPU{
```
federicobozzini: This might be fine for now, but a target might have multiple clusters (e.g. big.LITTLE or dynamicQ) so onl y one host won't be enough.

awphi: Interesting - @Luke-Parkin and I were thinking of swapping to use `lscpu` for the CPU probe in the future so I'd be curious to see how it presents such a target

internal/describe/describe.go
awphi: https://github.com/arm/topo/pull/9/commits/bda211018352a303e36a46e67a8dd728b2497c4d

internal/describe/describe.go
awphi: https://github.com/arm/topo/pull/9/commits/bda211018352a303e36a46e67a8dd728b2497c4d

internal/describe/describe.go
awphi: https://github.com/arm/topo/pull/9/commits/bda211018352a303e36a46e67a8dd728b2497c4d

cmd/topo/describe.go
awphi: https://github.com/arm/topo/pull/9/commits/bda211018352a303e36a46e67a8dd728b2497c4d

internal/describe/describe.go
Luke-Parkin: a big.Little cluster or dynamicQ cluster would still be under "host". It'd just be under a seperate `model:`

cmd/topo/describe.go
```diff
@@ -0,0 +1,65 @@
...
+		outputFile := filepath.Join(wd, targetDescriptionFilename)
+		err = os.WriteFile(outputFile, reportBytes, 0o0644)
+		if err != nil {
+			return err
+		}
+
+		c := console.NewLogger(os.Stderr, term.Plain)
+		c.Log(logger.Entry{
+			Level:   logger.Info,
+			Message: fmt.Sprintf("Target description written to %s", outputFile),
```
Luke-Parkin: Im not sure how this aligns with our best practices and/or ci guidelines- but I almost feel like this should just return the path to the file.

internal/describe/describe.go
th3james: Agreed - you're adding new behaviour, there should be some tests. Baking testing in early forces you to think about how to make your design testable.

internal/describe/describe.go
awphi: So do you think `host` should be a list of cpus like the following?

```yaml
host:
  - model: Cortex-A725 (big)
  features:
    - a
    - b
    - c
  - model: Cortex-A55 (little)
  features:
    - a
    - b
    - c
```

currently it's just 

```yaml
host:
  model: Cortex-A55
  features:
    - a
    - b
    - c
```

th3james (changes requested):
I'd like to see some test coverage

internal/describe/describe.go
awphi: Fair, was thinking the actual of logic of `describe.Generate` is so minimal atm as it just calls out to health.Probe but there's no harm in adding some extra tests

internal/describe/describe.go
federicobozzini: What is a `model`?

awphi: CPU model name, it's not included in this PR but added in the above example for clarity on what each array item is

federicobozzini:
> redacted

Not sure if we should link to private repos from a (soon-to-be) public one.

internal/describe/describe.go
federicobozzini: @awphi 

I've run the command on a target and what I get is :

```yaml
host:
    features:
        - Features
        - ':'
        - fp
        - asimd
        - evtstrm
        - aes
remoteprocs:
    - name: imx-rproc
```

No model. I also don't see it defined in the code anywhere.

Extra thing, I don't know where the `':'` comes from but it doesn't look right. 


You also wrote:
```yaml
host:
  - model: Cortex-A725 (big)
  features:
    - a
    - b
    - c
  - model: Cortex-A55 (little)
  features:
    - a
    - b
    - c
```

Nothing terrible wrong here, but I think the features would be the same between the two `model`s here. so there is no need to repeat them.

awphi: > No model. I also don't see it defined in the code anywhere.

Yes sorry, was getting my wires crossed - this is added by the upstack PR #10

> Extra thing, I don't know where the ':' comes from but it doesn't look right.

This is a bug in the current `ProbeHardware` that is currently mitigated by the fact that when displaying the features in `topo health` we filter for specific arm features. This should be addressed separately imo (and indeed is in #10 by virtue of using `lscpu`)

> Nothing terrible wrong here, but I think the features would be the same between the two models here. so there is no need to repeat them.

Okay cool so perhaps the schema should be something like

```yaml
host: 
  procs:
    - name: Cortex-A725 (big)
       cores: 20
    - name: Cortex-A55 (small)
       cores: 2
  features:
    - a
    - b
    - c
```

(again, neither name nor cores are implemented here, just trying to hammer down if there's anything in this base schema that needs to change to support multi-cluster targets as you originally alluded to)

cmd/topo/describe.go
awphi: The problem is `Printer` only deals with `Printable`s which atm have to define `AsJSON` and `AsPlain`. So not only would this mean we'd have to implement those two methods for the hardware report (which we'd never use) we'd also have to implement `AsYAML` on all the other printables which we'd also never use.

I'm open to a wider refactor but not sure it belongs in this first pass?

cmd/topo/describe.go
awphi: I agreed but as discussed in person, agreed to veto for now since we don't have a central abstraction for logging without prefix atm, will re-visit if this becomes a paint point

cmd/topo/describe.go
awphi: https://github.com/arm/topo/pull/9/commits/656abd1fb7234b1887fa5512957f3cf56d122d76 - will remember going forward!

cmd/topo/describe.go
awphi: I've pulled out a separate writer here https://github.com/arm/topo/pull/9/commits/8c7f825a24267cd57b1261ff6ed2c157fc4df394 - imo this suffices until we want YAML output across the codebase at which point `TargetHardwareReport` can become a printable - WDYT?

internal/describe/describe.go
awphi: https://github.com/arm/topo/pull/9/commits/7fb2990bc91bfb19eeca32cb7ca10109f5150dc8

CHANGELOG.md
```diff
@@ -5,4 +5,4 @@
 ### Added
 
 - Initial release
-- Operations supported: `help`, `clone`, `completion`, `deploy`, `extend`, `health`, `init`, `service`, `stop`, and `templates`.
+- Operations supported: `help`, `clone`, `completion`, `deploy`, `extend`, `health`, `describe`, `init`, `service`, `stop`, and `templates`.
```
tgonzalezorlandoarm: nit: should be alphabetically ordered except for `help` IMO

cmd/topo/describe.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/arm-debug/topo-cli/internal/output/term"
+	"github.com/arm-debug/topo-cli/internal/ssh"
+	"github.com/spf13/cobra"
+)
+
+var describeTarget string
+
+var describeCmd = &cobra.Command{
+	Use:   "describe",
+	Short: "Describe the hardware characteristics of the target host including CPU ISA features and remoteproc capabilities",
```
tgonzalezorlandoarm: Short: Describe the hardware characteristics of the target host
Long: Generates a target.yaml file that describes the hardware characteristics of the target host including CPU ISA features and remoteproc capabilities

cmd/topo/describe.go
```diff
@@ -0,0 +1,58 @@
...
+	Args:  cobra.ExactArgs(0),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		cmd.SilenceUsage = true
+		sshTarget, err := resolveTarget(describeTarget)
+		if err != nil {
+			return err
+		}
+
+		conn := health.NewConnection(sshTarget, ssh.ExecSSH)
```
tgonzalezorlandoarm, line 30: probably not for this PR but I wonder if NewConnection should be in `health`

README.md
```diff
@@ -52,6 +52,14 @@ The `--target` flag accepts SSH config host aliases or `user@host` destinations.
 ./topo deploy --target my-board
 ```
 
+### Generate a description of your target hardware
+
+```sh
+./topo describe --target my-board
```
tgonzalezorlandoarm: According to the manual flow in https://confluence.arm.com/x/sf0Nmw `describe` should happen after `health` and before `templates`

README.md
```diff
@@ -52,6 +52,14 @@ The `--target` flag accepts SSH config host aliases or `user@host` destinations.
...
+```
+
+This creates a `target-description.yaml` in the current directory.
```
tgonzalezorlandoarm: Maybe explain what it can be used for?

internal/describe/describe.go
```diff
@@ -0,0 +1,68 @@
...
+	"os"
+	"path/filepath"
+
+	"github.com/arm-debug/topo-cli/internal/health"
+	"go.yaml.in/yaml/v4"
+)
+
+const targetDescriptionFilename = "target-description.yaml"
+
+type HostCPU struct {
```
tgonzalezorlandoarm: TargetHostCPU sounds better to me, but maybe that's just me

internal/describe/describe.go
```diff
@@ -0,0 +1,68 @@
...
+type RemoteprocCPU struct {
+	Name string
+}
+
+type TargetHardwareReport struct {
+	Host        HostCPU
+	RemoteProcs []RemoteprocCPU
+}
+
+func Generate(conn health.Connection) (TargetHardwareReport, error) {
```
tgonzalezorlandoarm: GenerateDescription?

internal/describe/describe.go
```diff
@@ -0,0 +1,68 @@
...
+
+	hwProfile, err := conn.ProbeHardware()
+	if err != nil {
+		return TargetHardwareReport{}, err
+	}
+
+	return generateReport(hwProfile), nil
+}
+
+func WriteTargetDescriptionFile(dir string, report TargetHardwareReport) (string, error) {
```
tgonzalezorlandoarm: WriteTargetDescriptionToFile

tgonzalezorlandoarm (commented):
Some nit comments, some opinion comments, and some valid comments

tgonzalezorlandoarm:
> > https://github.com/Arm-Debug/topo-team-docs/blob/master/CODING_STYLE.md#8-new-code-is-accompanied-by-tests
> 
> Not sure if we should link to private repos from a (soon-to-be) public one.

There's an argument to be made for making the CODING style public as part of the contributing process? If we are already taking an opinionated approach to our code, we should ask our contributors to follow this approach as well.

cmd/topo/describe.go
awphi: Agreed it should definitely not, planning to do a refactor to better decouple health and describe as a follow up

internal/describe/describe_test.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+func TestGenerate(t *testing.T) {
+	t.Run("returns hardware profile for given target", func(t *testing.T) {
+		mockExecSSH := func(target ssh.Host, command string) (string, error) {
+			if strings.Contains(command, "cpuinfo") {
+				return "Features: feature1 feature2\n", nil
+			}
+			if strings.Contains(command, "remoteproc") {
+				return "remoteproc1\nremoteproc2\n", nil
```
tgonzalezorlandoarm: I notice the difference in feature separation vs remoteproc separation. Just out of curiosity, are we dependent on this exact format? I'd make this insensitive to the separator if so

internal/describe/describe_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+		outputFile, err := describe.WriteTargetDescriptionFile(dir, report)
+		require.NoError(t, err)
+		content, err := os.ReadFile(outputFile)
+		require.NoError(t, err)
+		err = yaml.Unmarshal(content, &reportOut)
+		require.NoError(t, err)
+
+		require.FileExists(t, outputFile)
+		assert.Equal(t, report, reportOut)
```
yejseo01, line 78: this bit seems like "assert" stage of AAA - https://semaphore.io/blog/aaa-pattern-test-automation

so breaking the lines accordingly would be good

```suggestion

		content, err := os.ReadFile(outputFile)
		require.NoError(t, err)
		err = yaml.Unmarshal(content, &reportOut)
		require.NoError(t, err)
		require.FileExists(t, outputFile)
		assert.Equal(t, report, reportOut)
```

internal/describe/describe_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+		outputFile1, err := describe.WriteTargetDescriptionFile(dir, report1)
+		require.NoError(t, err)
+		outputFile2, err := describe.WriteTargetDescriptionFile(dir, report2)
+		require.NoError(t, err)
+		content, err := os.ReadFile(outputFile2)
+		require.NoError(t, err)
+
+		require.Equal(t, outputFile1, outputFile2)
+		assert.Contains(t, string(content), "feature1")
```
yejseo01, line 102:
```suggestion

		content, err := os.ReadFile(outputFile2)
		require.NoError(t, err)
		require.Equal(t, outputFile1, outputFile2)
		assert.Contains(t, string(content), "feature1")
```

yejseo01 (commented):
minor comments regarding line breaks

internal/describe/describe_test.go
```diff
@@ -0,0 +1,100 @@
...
+		err = yaml.Unmarshal(content, &reportOut)
+		require.NoError(t, err)
+
+		require.FileExists(t, outputFile)
+		assert.Equal(t, report, reportOut)
+	})
+
+	t.Run("overwrites existing file", func(t *testing.T) {
+		dir := t.TempDir()
+		report1 := describe.TargetHardwareReport{}
```
tgonzalezorlandoarm: very small comment, feel free to ignore: I'd make report1 write 2 features.

internal/describe/describe_test.go
awphi: It's the format of the commands we're executing as probes atm - `cat /proc/cpuinfo` and `ls`-ing the remoteproc dir

internal/describe/describe.go
awphi: Went with `GenerateTargetDescription` for consistency with the write fn https://github.com/arm/topo/pull/9/commits/25655766d4984ba78b668a488b7f5af146494bb7

internal/describe/describe.go
awphi: https://github.com/arm/topo/pull/9/commits/25655766d4984ba78b668a488b7f5af146494bb7

README.md
awphi: https://github.com/arm/topo/pull/9/commits/af07906aee3c6de25b0dbeab029d27aa362ddf7e

README.md
awphi: Okay moved health to the top with describe following it - https://github.com/arm/topo/pull/9/commits/af07906aee3c6de25b0dbeab029d27aa362ddf7e

cmd/topo/describe.go
awphi: https://github.com/arm/topo/pull/9/commits/2b779f1c4ecf1fb9b0ea15dd4e9319cbbf5c43d1

CHANGELOG.md
awphi: https://github.com/arm/topo/pull/9/commits/1fbed6978422c361782398bd1e4b336917ca61c9

internal/describe/describe_test.go
awphi: https://github.com/arm/topo/pull/9/commits/48c6d0183eccdfb4f5ba2786f14688dba78efa10

internal/describe/describe_test.go
awphi: https://github.com/arm/topo/pull/9/commits/48c6d0183eccdfb4f5ba2786f14688dba78efa10 - good call. Always get tripped up when writing tests on what to classify as action vs assertion in go when we have to mix in loads of `require.NoError` in the action stage 😵

internal/describe/describe_test.go
awphi: https://github.com/arm/topo/pull/9/commits/2c5049ba269c690c20af3bb7e246a83d407a8847

internal/describe/describe.go
awphi: https://github.com/arm/topo/pull/9/commits/42e18226f354ea1858da8497e1d21313b67dff6c sounds reasonable, not sure if this struct is going to be the long term solution though given there's so much overlap between it and health's `HardwareProfile`

cmd/topo/describe.go
awphi: https://github.com/arm/topo/pull/9/commits/8c7f825a24267cd57b1261ff6ed2c157fc4df394

internal/describe/describe_test.go
tgonzalezorlandoarm: Maybe worth making it insensitive to the separator then?

tgonzalezorlandoarm (approved):
LGTM 🚀 
Whatever is left from my side are nits

internal/describe/describe_test.go
awphi: The actual logic already is - will swap the separators in the test for clarity! https://github.com/arm/topo/pull/9/commits/0aa2d35657ef389250a48f63191d504e0ac39f5c

tgonzalezorlandoarm (approved):
LGTM 🚀

internal/describe/describe_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+		outputFile1, err := describe.WriteTargetDescriptionToFile(dir, report1)
+		require.NoError(t, err)
+		outputFile2, err := describe.WriteTargetDescriptionToFile(dir, report2)
+		require.NoError(t, err)
+
+		content, err := os.ReadFile(outputFile2)
+		require.NoError(t, err)
+		require.Equal(t, outputFile1, outputFile2)
+		assert.Contains(t, string(content), "feature1")
```
th3james, line 102: You need to assert the negative case here, right? A no-op would pass this test?
```suggestion
		assert.Contains(t, string(content), "feature1")
		assert.NotContains(t, string(content), "feature2")
```

internal/describe/describe.go
```diff
@@ -0,0 +1,68 @@
...
+	f, err := os.Create(outputFile)
+	if err != nil {
+		return "", err
+	}
+	defer func() {
+		_ = f.Close()
+	}()
+
+	encoder := yaml.NewEncoder(f)
+	return outputFile, encoder.Encode(report)
```
th3james: Throwing away errors a bit here - not sure `defer` is necessary given there's only one operation to handle errors in
```suggestion
  encoder := yaml.NewEncoder(f)
  if err := encoder.Encode(report); err != nil {
      f.Close()
      return "", err
  }

  return outputFile, f.Close()
```

th3james (changes requested):

internal/describe/describe.go
awphi: Yep `defer` not strictly necessary but I'm a usually a fan since it makes the function easy to extend later. In your suggestion the `f.Close` call inside the encode error handler still generates an error we need to account for so went with 
```go
	if err := encoder.Encode(report); err != nil {
		closeErr := f.Close()
		return "", errors.Join(err, closeErr)
	}
```
not used `errors.Join` before but seemed appropriate here - lmk if you disagree https://github.com/arm/topo/pull/9/commits/d473ecfdd46442c64a4e04317f06fc31f5dbad58

internal/describe/describe_test.go
awphi: Ah yes good spot - changed the test data in response to a comment and didn't re-check the assertions 🤦 https://github.com/arm/topo/pull/9/commits/f9c2a00d39cf4ba38532ba3ed5efdaebb69377f6

th3james (approved):
Nice work

PR #10, opened 2026-02-10
feat: lscpu-based describe

awphi:
## Changes

This PR builds on the foundation of #9 to extend the information provided in `target-description.yaml` with useful presentational data:

- Changes the current health check to use `lscpu` over cat-ing out `cpuinfo` - this gives us an easier to parse format (JSON), core count and human-readable CPU model name to write into describe

Fabricated sample `topo describe` output:

```yaml
host:
    - model: Cortex-X925
      cores: 10
      features:
        - fp
        - asimd
        - evtstrm
        - aes
    - model: Cortex-A725
      cores: 10
      features:
        - fp
        - asimd
        - evtstrm
        - aes
remoteprocs: 
    - name: imx-rproc
````

Will follow up with a later PR to decouple the hardware checks from the health package.

### CODE REVIEW

**Comment by awphi** [comment]
Re-drafting as not ready for review given changes on #9

File: internal/health/target.go [hunk 1]
```diff
@@ -81,37 +104,108 @@ func (c *Connection) CheckDependencies(hardware map[HardwareCapability]struct{})
...
 	}
+	hp.RemoteCPU = cpus
 
 	return hp, nil
 }
 
-func (c *Connection) collectFeatures() ([]string, error) {
-	out, err := c.Run("grep -m1 Features /proc/cpuinfo")
+func (c *Connection) collectCPUInfo() ([]HostProcessor, error) {
+	ok, err := c.BinaryExists("lscpu")
```
logan-scully-arm, line 123: You might have a better idea than me on this, but I was under the impression that we shouldn't assume `lscpu` is present. I believe a lot of embedded devices might not have it.

We might be okay having this as a requirement, but I thought I'd get your take on it.

internal/health/target.go
```diff
@@ -81,37 +104,108 @@ func (c *Connection) CheckDependencies(hardware map[HardwareCapability]struct{})
...
+		fields []LscpuOutputField
+	}
+	var coreTypes []coreType
+
+	for _, f := range fields {
+		if f.Field == "Model name:" {
+			coreTypes = append(coreTypes, coreType{name: f.Data})
+			continue
+		}
+		if len(coreTypes) > 0 { // Belongs to the last modelName
```
yejseo01:
```suggestion
		if len(coreTypes) > 0 {
```

internal/health/target_test.go
```diff
@@ -38,66 +50,243 @@ func TestRun(t *testing.T) {
...
+		assert.Equal(t, []string{"fp", "asimd"}, hw.HostProcessor[0].Features)
+	})
+
+	t.Run("returns error when lscpu not found", func(t *testing.T) {
+		mockExec := func(_ ssh.Host, command string) (string, error) {
+			switch {
+			case strings.Contains(command, "command -v"):
+				return "", errors.New("not found")
+			default:
+				return "", errors.New("not found")
```
yejseo01: If both cases are going to return the same thing, the switch case is not necessary
```suggestion
			return "", errors.New("not found")
```

internal/health/target.go
awphi: Yea we thought about this and concluded the benefits we get from `lscpu` (human-readable chip names, sane consistent parseable JSON output) outweigh the negatives of not supporting a subset of distros that don't have it as standard. FWIW we tested it on the RPi 5, imx93, stm32mp, orion and DGX spark and they all had it as standard.

We do check and error appropriately if it doesn't exist so it's always something that could be addressed later if/when we find a board that we want to support where it's not available and/or easy to install.

internal/health/target_test.go
```diff
@@ -38,66 +50,243 @@ func TestRun(t *testing.T) {
...
-				return "", nil // simulate successful initial connection
+			switch {
+			case command == "true":
+				return "", nil
+			case strings.Contains(command, "command -v"):
+				return "/usr/bin/lscpu", nil
+			case command == "lscpu --json":
+				return lsCpuOutputRaw, nil
+			default:
+				return "", errors.New("no remoteproc")
```
th3james:
```suggestion
				return "", errors.New("unexpected command: " + command)
```

internal/health/target_test.go
```diff
@@ -38,66 +50,243 @@ func TestRun(t *testing.T) {
 }
 
 func TestProbe(t *testing.T) {
-	t.Run("probe succeeds and collects features", func(t *testing.T) {
+	t.Run("probe succeeds and collects CPU info", func(t *testing.T) {
```
th3james, line 43: The probe command now bundles together two quite distinct target checks:
- health: only cares about connectivity and whether remoteproc exists
- describe: cares about hardware, doesn't care about dependencies

IMHO, it would be better to replace `Probe` with a health-check focussed probe function, primarily because the hardware probes are potentially quite expensive and we're throwing them away.

Separating these two things will also reduce the amount of irrelevant state mocking required in the tests.

internal/health/target_test.go
```diff
@@ -38,66 +50,243 @@ func TestRun(t *testing.T) {
...
+				return "", errors.New("not found")
+			default:
+				return "", errors.New("not found")
+			}
+		}
+
+		conn := health.NewConnection("hostname", mockExec)
+		_, err := conn.ProbeHardware()
+
+		assert.Error(t, err)
```
th3james:
```suggestion
		assert.ErrorContains(t, err, "lscpu not found")
```

internal/health/target_test.go
```diff
@@ -38,66 +50,243 @@ func TestRun(t *testing.T) {
...
+				return "not json", nil
+			default:
+				return "", errors.New("not found")
+			}
+		}
+
+		conn := health.NewConnection("hostname", mockExec)
+		_, err := conn.ProbeHardware()
+
+		assert.Error(t, err)
```
th3james:
```suggestion
		assert.ErrorContains(t, err, "collecting CPU info")
```

th3james (changes requested):

internal/health/target_test.go
awphi: https://github.com/arm/topo/pull/10/commits/91560838ef7cba4773eb5b593255d565734c9ef6

internal/health/target_test.go
awphi: https://github.com/arm/topo/pull/10/commits/91560838ef7cba4773eb5b593255d565734c9ef6

internal/health/target_test.go
awphi: https://github.com/arm/topo/pull/10/commits/91560838ef7cba4773eb5b593255d565734c9ef6

internal/health/target_test.go
awphi: Totally agree but best left for a follow-up PR if you agree? I don't think it should be too hard to decouple even after merging this - WDYT @Luke-Parkin? Keen to avoid a mega refactor/feature-combo PR

internal/health/target.go
logan-scully-arm: Yeah if it works across all of our test boards then that sounds good.

internal/health/target_test.go
awphi: Not quite ready for review but to proof this exists - https://github.com/arm/topo/compare/awphi/lscpu-describe...awphi/refactor-target?expand=1

Luke-Parkin: Definitely agreed, I'd like to put this in first so that vscode can start working on it, then decouple.

internal/health/target.go
tgonzalezorlandoarm, line 123: Should we add `lscpu` as a `TargetRequiredDependencies` in `internal/health/dependencies.go`

tgonzalezorlandoarm: (I think checking it should be part of the health command)

tgonzalezorlandoarm: Also (perhaps for a next PR) could should we have `install` also install `lscpu` if necessary?

internal/health/target_test.go
tgonzalezorlandoarm: agreed

internal/health/target.go
awphi: I agree with both of these but given we've a) yet to find a board without lscpu pre-installed and b) are planning a plugin system - I'd like to suggest this could be one of our first and perhaps default plugins?  Adding it as a `TargetRequiredDependencies` sounds reasonable for now though so will do that - good shout!

awphi: https://github.com/arm/topo/pull/10/commits/2d1a18cc88780b56c903b902b7d745b9b0ac0348

tgonzalezorlandoarm: I don't think we need `c.BinaryExists("lscpu")` anymore? From what I remember, the health command will check this for you now

tgonzalezorlandoarm: > I'd like to suggest this could be one of our first and perhaps default plugins?

I'm not sure I understand why this would be a plugin and not just a default check?

awphi: > I'm not sure I understand why this would be a plugin and not just a default check?

My understanding is we plan to implement even these sort of core things as plugins that are shipped with topo to dogfood the architecture and given it's not urgent as we've not found a board without lscpu yet I'm suggesting we take this approach for lscpu.

> I don't think we need c.BinaryExists("lscpu") anymore? From what I remember, the health command will check this for you now

Not strictly necessary true but I'd like to keep it in case the user runs describe without running health so you still get a useful error message as we don't don't do a full dependency check in describe

tgonzalezorlandoarm: All of that makes sense to me. :rocket:

tgonzalezorlandoarm (approved):

internal/health/target.go
```diff
@@ -1,17 +1,40 @@
...
 
 	"github.com/arm-debug/topo-cli/internal/ssh"
 )
 
 type execSSH func(target ssh.Host, command string) (string, error)
 
+type HostProcessor struct {
+	ModelName string   `yaml:"model"`
+	Cores     int      `yaml:"cores"`
+	Features  []string `yaml:"features"`
```
federicobozzini, line 18: As already mentioned the features here would be duplicated for all `HostProcessor`s. Do we want this?

internal/health/target.go
```diff
@@ -81,37 +104,108 @@ func (c *Connection) CheckDependencies(hardware map[HardwareCapability]struct{})
...
+			v, err := strconv.Atoi(f.Data)
+			if err != nil {
+				return HostProcessor{}, err
+			}
+			coresPerUnit = v
+		case "Socket(s):", "Cluster(s):":
+			v, err := strconv.Atoi(f.Data)
+			if err != nil {
+				// Some platforms report "-" for sockets when using clusters
+				continue
```
federicobozzini, line 163: Don't we end up making stuff up here? Wouldn't it be better to just leave the field empty in some way?

internal/health/target.go
```diff
@@ -1,17 +1,40 @@
...
+	RemoteCPU     []RemoteprocCPU `yaml:"remoteprocs"`
+}
+
+type LscpuOutputField struct {
+	Field    string             `json:"field"`
+	Data     string             `json:"data"`
+	Children []LscpuOutputField `json:"children,omitempty"`
+}
+
+type LscpuOutput struct {
```
federicobozzini: NIT: this struct can be made private.

internal/health/target.go
```diff
@@ -1,17 +1,40 @@
...
+	Lscpu []LscpuOutputField `json:"lscpu"`
```
federicobozzini, line 37: Minor observation, but this type feels a bit weird. It'also only used by doing `lscpu.lscpu` which to me suggests it might be unnecessary or poorly designed.

awphi: Agree it's ugly but it is the output format of `lscpu --json` and defining a matching type is required for unmarshalling

federicobozzini: Maybe we can just rename the variable where it's used to something like `lscpuOutput`.

internal/health/target.go
awphi: Iirc @Luke-Parkin's conclusion on this was that `lscpu` must be duplicating it per host proc for a reason so unless we can with 100% certainty assert they are always duplicated then we shouldn't touch it.

internal/health/target.go
awphi: https://github.com/arm/topo/pull/10/commits/119737b4b0c32b8e6ecaa92455a01190efa9a898

internal/health/target.go
awphi: https://github.com/arm/topo/pull/10/commits/119737b4b0c32b8e6ecaa92455a01190efa9a898

internal/health/target.go
awphi: You should always get a "Clusters" or a "Sockets" entry e.g.:

imx93:
```
Socket(s):                            -
Cluster(s):                           1
```

DGX spark:
```
Socket(s):                               1
```

That being said I've added a check + corresponding test to enforce this - https://github.com/arm/topo/pull/10/commits/119737b4b0c32b8e6ecaa92455a01190efa9a898

th3james (approved):

PR #3, opened 2026-02-06
chore: remove qlty

awphi:
## Changes

- Removes qlty cloud from our CI actions
- Will add https://goreportcard.com/ in a follow-up in its place

PR #11, opened 2026-02-10
chore: remove straggling qlty files

awphi:
## Changes

- Removes the qlty config that was missed here #3


federicobozzini (approved):
We might also need to remove QLTY from the required checks for PRs. I'll check.

tgonzalezorlandoarm (approved):

awphi:
> We might also need to remove QLTY from the required checks for PRs. I'll check.

I never added it as a required check in this repo 😏

PR #14, opened 2026-02-11
feat: add topo setup-keys command

tgonzalezorlandoarm:
* Add the option to specify the key path or not.
* Add a --dry-run flag

For now, only support:
- Linux
- Creating keys through ssh-keygen
- Sending keys to the target through ssh-copy-id.

### DIFF
--- a/cmd/topo/setup_keys.go
+++ b/cmd/topo/setup_keys.go
@@ -50,5 +50,5 @@ func init() {
 	rootCmd.AddCommand(setupKeysCmd)
 	addTargetFlag(setupKeysCmd, &setupKeysTarget)
 	setupKeysCmd.Flags().BoolVar(&setupKeysDryRun, "dry-run", false, "Show what commands would be executed without actually running them")
-	setupKeysCmd.Flags().StringVar(&setupKeysKeyPath, "key-path", "", "Specify the SSH key path where the generated keys will be stored (default is ~/.ssh/id_ed25519_topo_<target>)")
+	setupKeysCmd.Flags().StringVar(&setupKeysKeyPath, "key-path", "", "Specify the SSH path where the generated key pair will be stored. Default directory: ~/.ssh. Default public key file name: id_ed25519_topo_<target>.pub)")
 }
--- a/cmd/topo/setup_keys.go
+++ b/cmd/topo/setup_keys.go
@@ -1,0 +1,54 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"runtime"
+
+	"github.com/arm-debug/topo-cli/internal/setupkeys"
+	"github.com/spf13/cobra"
+)
+
+var (
+	setupKeysTarget  string
+	setupKeysDryRun  bool
+	setupKeysKeyPath string
+)
+
+var setupKeysCmd = &cobra.Command{
+	Use:   "setup-keys",
+	Short: "Generate SSH keys for the target and install the public key on the target host",
+	Long: `Generate SSH keys for the target and install the public key on the target host.
+
+Use --dry-run to see what commands would be executed without actually running them.`,
+	Args: cobra.ExactArgs(0),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		cmd.SilenceUsage = true
+
+		if runtime.GOOS != "linux" {
+			return fmt.Errorf("topo setup-keys currently supports Linux hosts only")
+		}
+
+		resolvedTarget, err := resolveTarget(setupKeysTarget)
+		if err != nil {
+			return err
+		}
+
+		seq, err := setupkeys.NewKeyCreationAndPlacementOnTarget(resolvedTarget, setupKeysKeyPath)
+		if err != nil {
+			return err
+		}
+
+		if setupKeysDryRun {
+			return seq.DryRun(os.Stdout)
+		}
+		return seq.Run(os.Stdout)
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(setupKeysCmd)
+	addTargetFlag(setupKeysCmd, &setupKeysTarget)
+	setupKeysCmd.Flags().BoolVar(&setupKeysDryRun, "dry-run", false, "Show what commands would be executed without actually running them")
+	setupKeysCmd.Flags().StringVar(&setupKeysKeyPath, "key-path", "", "Specify the SSH key path where the generated keys will be stored (default is ~/.ssh/id_ed25519_topo_<target>)")
+}
--- a/internal/setupkeys/setupkeys.go
+++ b/internal/setupkeys/setupkeys.go
@@ -1,0 +1,96 @@
+package setupkeys
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"unicode"
+
+	goperation "github.com/arm-debug/topo-cli/internal/deploy/operation"
+)
+
+func NewKeyCreationAndPlacementOnTarget(target string, keyPath string) (goperation.Sequence, error) {
+	if keyPath == "" {
+		home, err := os.UserHomeDir()
+		if err != nil {
+			return nil, fmt.Errorf("failed to determine home directory: %w", err)
+		}
+
+		keyName := fmt.Sprintf("id_ed25519_topo_%s", sanitizeTarget(target))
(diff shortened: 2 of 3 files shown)


internal/setupkeys/setupkeys_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"runtime"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestNewSequenceDryRunOutputsCommands(t *testing.T) {
+	if runtime.GOOS != "linux" {
+		t.Skip("only runs on linux because it relies on linux-specific paths")
+	}
```
awphi: nit: could use `testutil.RequireOS(t, "linux")`

internal/setupkeys/setupkeys.go
```diff
@@ -0,0 +1,96 @@
...
+		keyPath = filepath.Join(home, ".ssh", keyName)
+	}
+
+	if err := ensureSSHDir(keyPath); err != nil {
+		return nil, err
+	}
+
+	ops := []goperation.Operation{
+		newSetupKeysOperation("Generate SSH key pair for target", []string{"ssh-keygen", "-t", "ed25519", "-f", keyPath, "-C", target}),
+		newSetupKeysOperation("Copy SSH public key to target", []string{"ssh-copy-id", "-i", keyPath + ".pub", target}),
```
awphi, line 32: Not sure about the naming here - the operation implementation is not related to setting up keys and is more of an arbitrary command executor. 

Imo we should adjust the name or adopt the pattern in the other commands where we have a specific struct tailored for each operation rather than this very flexible abstraction. My preference would be for the latter for consistency and consider moving to a more general abstraction across the wider codebase if we see fit.

internal/setupkeys/setupkeys.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func sanitizeTarget(target string) string {
+	var b strings.Builder
+	for _, r := range target {
+		switch {
+		case unicode.IsLetter(r) || unicode.IsDigit(r):
+			b.WriteRune(r)
+		case r == '-' || r == '_' || r == '.':
+			b.WriteRune(r)
```
awphi: nit: Can combine these cases

internal/setupkeys/setupkeys.go
```diff
@@ -0,0 +1,96 @@
...
+		default:
+			b.WriteRune('_')
+		}
```
awphi: Maybe unimportant but this will make ipv6 hosts with ports quite odd looking 
e.g. `[2001:db8::1]:443` becomes `_2001_db8__1__443`

internal/setupkeys/setupkeys.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	ops := []goperation.Operation{
+		newSetupKeysOperation("Generate SSH key pair for target", []string{"ssh-keygen", "-t", "ed25519", "-f", keyPath, "-C", target}),
+		newSetupKeysOperation("Copy SSH public key to target", []string{"ssh-copy-id", "-i", keyPath + ".pub", target}),
+	}
+	return goperation.NewSequence(ops...), nil
+}
+
+func ensureSSHDir(keyPath string) error {
```
awphi:
```suggestion
func ensureDir(keyPath string) error {
```

nothing SSH specific in this

cmd/topo/setup_keys.go
```diff
@@ -0,0 +1,54 @@
...
+		}
+		return seq.Run(os.Stdout)
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(setupKeysCmd)
+	addTargetFlag(setupKeysCmd, &setupKeysTarget)
+	setupKeysCmd.Flags().BoolVar(&setupKeysDryRun, "dry-run", false, "Show what commands would be executed without actually running them")
+	setupKeysCmd.Flags().StringVar(&setupKeysKeyPath, "key-path", "", "Specify the SSH key path where the generated keys will be stored (default is ~/.ssh/id_ed25519_topo_<target>)")
```
awphi:
```suggestion
	setupKeysCmd.Flags().StringVar(&setupKeysKeyPath, "key-path", "", "Specify the SSH key path where the generated keys will be stored (default is ~/.ssh)")
```

clarify the keys are going in `~/.ssh` not a subdir

internal/setupkeys/setupkeys_test.go
awphi, line 1: Would be good to see some tests for actually running the operation even if only with a custom key path to avoid interfering with other tests

awphi (changes requested):
Handful of comments - mostly small things as per!

internal/setupkeys/setupkeys_test.go
tgonzalezorlandoarm: Will do this in this PR

cmd/topo/setup_keys.go
tgonzalezorlandoarm: I was trying to say that the key name is going to be `id_ed25519_topo_<target>`

internal/setupkeys/setupkeys.go
tgonzalezorlandoarm: done: https://github.com/arm/topo/pull/14/commits/4e8c6e7fdbd616b3c745553e7f7700d43e78fb24

internal/setupkeys/setupkeys.go
tgonzalezorlandoarm: done: https://github.com/arm/topo/pull/14/commits/2f3c62086c6c1a281e85be83d666e9c6752946fb

internal/setupkeys/setupkeys_test.go
tgonzalezorlandoarm: done: https://github.com/arm/topo/pull/14/commits/12ff09dc2821b3ee2f4fe8068a99e32ce0555e35

cmd/topo/setup_keys.go
awphi: Maybe it's fine tbh. I'd still like to see the word file + pair in there as because keys have no extension it's slightly unclear - e.g. `ssh-keygen`:
```sh
Generating public/private ed25519 key pair.
Enter file in which to save the key (/Users/adam/.ssh/id_ed25519): 
```

tgonzalezorlandoarm: Hopefully this is clearer now?
https://github.com/arm/topo/pull/14/commits/540ba60cc0508616ddfce60ac50050950d5cf970

internal/setupkeys/setupkeys.go
tgonzalezorlandoarm: Interesting, hadn't thought of that. I'm keen on leaving it this way but open to alternatives.
Do you think we should change `:` specifically to something else for example?

cmd/topo/setup_keys.go
awphi: Very clear ty!

internal/setupkeys/setupkeys.go
tgonzalezorlandoarm: I agree, this will come in the next PR if that's ok with you?

internal/setupkeys/setupkeys_test.go
tgonzalezorlandoarm: Given that we are creating keys and copying them into target via ssh, probably best to fake out the commands? I've done that here, hopefully that's fine:
https://github.com/arm/topo/pull/14/commits/335aa6df2284d62aef8c75aae9a6aac4658e535a

tgonzalezorlandoarm: we could avoid faking out ssh-keygen

awphi: Looks like a good approach to me in lieu of an e2e test which is probably excessive when this feature is so young

internal/setupkeys/setupkeys.go
awphi: You're probably right - afaik it's a bit of an edge case so if no one else think this is important feel free to ignore!

awphi (approved):
LGTM 🚀

internal/setupkeys/setupkeys_test.go
awphi: > we could avoid faking out ssh-keygen

That would be good as a stretch - it's best to avoid mocking anything we don't absolutely have to and this will give us exposure to each platform's `ssh-keygen` impl

PR #16, opened 2026-02-12
Chore: Varless target flag

logan-scully-arm:
Just a small set of changes to remove the package level vars for handling the target flag.

## Changes

- Removed package level vars for `--target` flag
- Implements a command flag lookup in resolve target to check for the flag
- Resolve target as a result now takes a `cobra.Command` as its argument


cmd/topo/stop.go
```diff
@@ -28,7 +28,7 @@ Use --dry-run to see what commands would be executed without actually running th
 	RunE: func(cmd *cobra.Command, _ []string) error {
 		cmd.SilenceUsage = true
 
-		resolvedTarget, err := resolveTarget(stopTarget)
+		resolvedTarget, err := resolveTarget(cmd)
```
th3james: You haven't removed `stopTarget`

cmd/topo/root.go
```diff
@@ -27,16 +27,22 @@ func init() {
...
-	cmd.Flags().StringVar(target, "target", "", "The SSH destination.")
+func addTargetFlag(cmd *cobra.Command) {
+	cmd.Flags().StringP("target", "t", "", "The SSH destination.")
 }
 
-func resolveTarget(flagValue string) (string, error) {
+func resolveTarget(cmd *cobra.Command) (string, error) {
+	flagValue, err := cmd.Flags().GetString("target")
+	if err != nil {
+		return "", nil
```
th3james: As in the other PR, this is a programmer error, not a system state we anticipate occurring in a well functioning program. We should terminate with an assertion, and avoid the implication that this error should be handled further up the stack.
```suggestion
		panic(fmt.Sprintf("resolveTarget: flag not registered: %v", err))
```

cmd/topo/templates.go
```diff
@@ -21,12 +21,12 @@ var templatesCmd = &cobra.Command{
...
+			return err
+		}
 		if templateFilters.Target != "" {
-			target, err := resolveTarget(templateFilters.Target)
-			if err != nil {
-				return err
-			}
-			templateFilters.Target = target
+			templateFilters.Target = resolvedTarget
 		}
```
th3james: This call has different implications to the others - here the target is optional, rather than required. I'd recommend splitting `resolveTarget` up:
```go
  // lookupTarget returns the target if provided via --target or TOPO_TARGET, or "" if neither.
  func lookupTarget(cmd *cobra.Command) (string, bool) { ... }

  // requireTarget returns the target or errors if not specified.
  func requireTarget(cmd *cobra.Command) (string, error) {
      t, exists := lookupTarget(cmd)
      if !exists {
          return "", fmt.Errorf("target not specified: provide --target or set TOPO_TARGET env var")
      }
      return t, nil
  }
```
then here:
```suggestion
		if target, targetSpecified := lookupTarget(cmd); targetSpecified {
			templateFilters.Target = target
		}
```

th3james (changes requested):

cmd/topo/main.go
```diff
@@ -10,8 +10,8 @@ import (
 
 func main() {
 	if err := rootCmd.Execute(); err != nil {
-		outputFormat, outputFormatError := resolveOutput(rootCmd)
-		if outputFormatError != nil {
+		outputFormat, e := resolveOutput(rootCmd)
+		if e != nil {
```
federicobozzini:
```suggestion
		outputFormat, err := resolveOutput(rootCmd)
		if err != nil {
```
Just for consistency.

cmd/topo/templates.go
```diff
@@ -21,12 +21,9 @@ var templatesCmd = &cobra.Command{
 			return err
 		}
 
-		if templateFilters.Target != "" {
-			target, err := resolveTarget(templateFilters.Target)
-			if err != nil {
-				return err
-			}
-			templateFilters.Target = target
+		resolvedTarget, targetIncl := lookupTarget(cmd)
```
federicobozzini: What does `targetIncl` mean? Also, can't we just check if target is not empty?

logan-scully-arm: `targetIncl` was meant to be target included. 

I guess I can replace that with the error instead of panic.

logan-scully-arm: In retrospect, James made a good point in the other PR, I've removed the unneeded `exists` bool.

cmd/topo/main.go
logan-scully-arm: I realized this wasn't err because that clashes with the command error. I've put it back now

cmd/topo/root.go
```diff
@@ -27,26 +27,40 @@ func init() {
...
+	cmd.Flags().StringP("target", "t", "", "The SSH destination.")
 }
 
-func resolveTarget(flagValue string) (string, error) {
-	const targetEnvVar = "TOPO_TARGET"
+func lookupTarget(cmd *cobra.Command) string {
+	flagValue, err := cmd.Flags().GetString("target")
+	if err != nil {
+		panic(fmt.Sprintf("bug: output flag not registered: %v", err))
+	}
```
th3james, line 40: The point of this lookup function is to tolerate the value not existing. This is the distinction between `lookup` and `require`
```suggestion
```

cmd/topo/root.go
```diff
@@ -27,26 +27,40 @@ func init() {
...
 
-func addTargetFlag(cmd *cobra.Command, target *string) {
-	cmd.Flags().StringVar(target, "target", "", "The SSH destination.")
+func addTargetFlag(cmd *cobra.Command) {
+	cmd.Flags().StringP("target", "t", "", "The SSH destination.")
 }
 
-func resolveTarget(flagValue string) (string, error) {
-	const targetEnvVar = "TOPO_TARGET"
+func lookupTarget(cmd *cobra.Command) string {
```
th3james: Example of the lookup pattern from the standard library:
https://pkg.go.dev/os#LookupEnv
you return a tuple, with the boolean representing whether the value was present
```suggestion
func lookupTarget(cmd *cobra.Command) (string, boolean) {
```

cmd/topo/root.go
```diff
@@ -27,26 +27,40 @@ func init() {
...
-	if env := strings.TrimSpace(os.Getenv(targetEnvVar)); env != "" {
-		return env, nil
+
+	return ""
+}
+
+func requireTarget(cmd *cobra.Command) (string, error) {
+	t := lookupTarget(cmd)
+	if t == "" {
+		return "", fmt.Errorf("target not specified: provide --target or set TOPO_TARGET env var")
```
th3james: You've got the semantics backwards here."Lookup" tolerates absence of value, "require" demands that the value exists. This function exists to upgrade the lookup `ok = false` to a panic in situations where not specifying the value is a programming error.
```suggestion
	t, ok := lookupTarget(cmd)
	if !ok {
		return panic(fmt.Sprintf("bug: command require target flag, but it wasn't registered: %v", err))
```

th3james (dismissed):

cmd/topo/root.go
th3james: Ah, no, I'm wrong about this. This should be an error, as even if a function considers the flag optional, it should be registered, and not registering it is a programming error. There is, however, a copy-pasta typo in the flag name
```suggestion
	if err != nil {
		panic(fmt.Sprintf("bug: target flag not registered: %v", err))
	}
```

cmd/topo/root.go
```diff
@@ -27,26 +27,40 @@ func init() {
...
 	}
-	return "", fmt.Errorf("target not specified: provide --target or set TOPO_TARGET env var")
+	return t, nil
 }
 
 func resolveOutput(cmd *cobra.Command) (term.Format, error) {
 	flagValue, err := cmd.Flags().GetString("output")
 	if err != nil {
-		panic(fmt.Sprintf("bug: output flag not registered: %v", err))
+		return term.Plain, nil
```
th3james: This function should be renamed for consistency. It should either be `requireOutput` and continue to panic, or `lookupOutput`. Either way, it's outside of the scope of this PR

cmd/topo/root.go
logan-scully-arm: Yeah def should be an error. Good spot on the copy-pasta

cmd/topo/root.go
logan-scully-arm: I disagree w/ this; lookup does tolerate it not being present. (the only panic is when there's an unlikely issue in the cobra function to get the flag).

We surely shouldn't be panicking in this situation where there is actually a user mistake. I think you've got the use of OK wrong -- this is `exists` and it's false if there is no target found at lookup.

cmd/topo/root.go
logan-scully-arm: Good shout -- I'll do another once my current two are merged.

cmd/topo/root.go
```diff
@@ -27,26 +27,40 @@ func init() {
...
 
-func addTargetFlag(cmd *cobra.Command, target *string) {
-	cmd.Flags().StringVar(target, "target", "", "The SSH destination.")
+func addTargetFlag(cmd *cobra.Command) {
+	cmd.Flags().StringP("target", "t", "", "The SSH destination.")
 }
 
-func resolveTarget(flagValue string) (string, error) {
-	const targetEnvVar = "TOPO_TARGET"
+func lookupTarget(cmd *cobra.Command) (string, bool) {
```
tgonzalezorlandoarm, line 34: I see a good opportunity here to sanitize the target string.

`len(strings.Fields(cmd.Flags().GetString("target"))) == 1`? Something along the lines of that

tgonzalezorlandoarm: We should basically have a step in which we guarantee that nothing is injected, and this function looks like the ideal place to do so IMO

tgonzalezorlandoarm (changes requested):
I realize that I haven't put the request changes in the comment before

yejseo01 (commented):
LGTM, but make sure you raise a differnt PR/edit this PR for sanitizing the target here before merging so we don't forget

cmd/topo/root.go
logan-scully-arm: yeah sounds good, doing that now

cmd/topo/root.go
```diff
@@ -27,26 +28,53 @@ func init() {
...
 
-func addTargetFlag(cmd *cobra.Command, target *string) {
-	cmd.Flags().StringVar(target, "target", "", "The SSH destination.")
+func addTargetFlag(cmd *cobra.Command) {
+	cmd.Flags().StringP("target", "t", "", "The SSH destination.")
 }
 
-func resolveTarget(flagValue string) (string, error) {
+var sshTargetRgx = regexp.MustCompile(
+	`^(?:[a-zA-Z0-9._-]+@)?[a-zA-Z0-9._-]+(?::\d+)?$`,
```
tgonzalezorlandoarm: In general, I consider sanitizing input with regexp (specially in the cases where it has security implications) is not a good practice.

Reasons:
1. regexp are very complicated to get right
2. regexp are complicated to understand.
3. regexp are very complicated to verify.
4. It's hard to see if a regexp is exploitable or not.

tgonzalezorlandoarm: Thinking more about this, I think sanitizing the input may be worth considering in a separate PR

tgonzalezorlandoarm (approved):

tgonzalezorlandoarm (changes requested):
Sorry, meant to request changes here

cmd/topo/root.go
logan-scully-arm: Yeah I'll do that in a different PR, will not use regex :+1:

tgonzalezorlandoarm (approved):

PR #17, opened 2026-02-12
feat: detect SSH password-based authentication

tgonzalezorlandoarm:
topo does not support password-based authentication (and does not intend to, as key-based authentication is stronger).

Currently, the first command in topo's flow is:
- `topo health --target <target>`.

`topo health` is therefore the command in which we check what type of authentication is being used in the SSH.

 * Add a check in `topo health` that fails early if SSH password-based authentication is the only authentication that is set up.

This approach means that if `topo health` is skipped, the user won't get this early failure and will probably face issues on the rest of the commands.

### DIFF
--- a/cmd/topo/health.go
+++ b/cmd/topo/health.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"fmt"
 	"os"
 
 	"github.com/arm-debug/topo-cli/internal/health"
@@ -9,10 +10,9 @@ import (
 	"github.com/spf13/cobra"
 )
 
-var (
-	healthTarget           string
-	healthAcceptNewHostKey bool
-)
+var healthTarget string
+
+const acceptNewHostFlag = "accept-new-host-keys"
 
 var healthCmd = &cobra.Command{
 	Use:   "health",
@@ -28,7 +28,13 @@ var healthCmd = &cobra.Command{
 		if err != nil {
 			return err
 		}
-		report, err := health.Check(sshTarget, healthAcceptNewHostKey)
+
+		acceptNewHostKeys, err := cmd.Flags().GetBool(acceptNewHostFlag)
+		if err != nil {
+			panic(fmt.Sprintf("internal error: %s flag not registered: %v", acceptNewHostFlag, err))
+		}
+
+		report, err := health.Check(sshTarget, acceptNewHostKeys)
 		if err != nil {
 			return err
 		}
@@ -38,6 +44,6 @@ var healthCmd = &cobra.Command{
 
 func init() {
 	addTargetFlag(healthCmd, &healthTarget)
-	healthCmd.Flags().BoolVar(&healthAcceptNewHostKey, "accept-new-host-keys", false, "Automatically trust and add new SSH host keys for the target")
+	healthCmd.Flags().Bool(acceptNewHostFlag, false, "Automatically trust and add new SSH host keys for the target")
 	rootCmd.AddCommand(healthCmd)
 }
--- a/cmd/topo/health.go
+++ b/cmd/topo/health.go
@@ -9,7 +9,10 @@ import (
 	"github.com/spf13/cobra"
 )
 
-var healthTarget string
+var (
+	healthTarget           string
+	healthAcceptNewHostKey bool
+)
 
 var healthCmd = &cobra.Command{
 	Use:   "health",
@@ -25,7 +28,7 @@ var healthCmd = &cobra.Command{
 		if err != nil {
 			return err
 		}
-		report, err := health.Check(sshTarget)
+		report, err := health.Check(sshTarget, healthAcceptNewHostKey)
 		if err != nil {
 			return err
 		}
@@ -35,5 +38,6 @@ var healthCmd = &cobra.Command{
 
 func init() {
 	addTargetFlag(healthCmd, &healthTarget)
+	healthCmd.Flags().BoolVar(&healthAcceptNewHostKey, "accept-new-host-keys", false, "Automatically trust and add new SSH host keys for the target")
 	rootCmd.AddCommand(healthCmd)
 }
--- a/internal/health/health.go
+++ b/internal/health/health.go
@@ -90,12 +90,13 @@ func GenerateReport(hostDependencies []DependencyStatus, targetStatus Status) Re
 func Check(sshTarget string, acceptNewHostKeys bool) (Report, error) {
 	dependencyStatuses := CheckInstalled(HostRequiredDependencies, BinaryExistsLocally)
 
-	sshExists, err := BinaryExistsLocally("ssh")
-	if err != nil {
-		return GenerateReport(dependencyStatuses, Status{}), err
+	authProbeEnabled := false
+	for _, s := range dependencyStatuses {
+		if s.Dependency.Name == "ssh" {
+			authProbeEnabled = s.Installed
+			break
+		}
 	}
-
-	authProbeEnabled := sshExists
 	opts := target.ConnectionOptions{
 		AuthProbeEnabled:  authProbeEnabled,
 		AcceptNewHostKeys: acceptNewHostKeys,
--- a/internal/health/health.go
+++ b/internal/health/health.go
@@ -9,7 +9,7 @@ import (
 	"github.com/arm-debug/topo-cli/internal/target"
 )
 
-const passwordAuthErrorMessage = `Topo does not support SSH password-based authentication. To connect, either:
+const passwordAuthErrorMessage = `note: Topo does not support SSH password-based authentication. To connect, either:
 - create your own SSH keys for the target, or
 - run 'topo setup-keys --target <target>' to let Topo generate keys and configure passwordless authentication`
 
--- a/internal/health/health.go
+++ b/internal/health/health.go
@@ -1,11 +1,15 @@
 package health
 
 import (
+	"errors"
+	"os"
 	"strings"
 
 	"github.com/arm-debug/topo-cli/internal/ssh"
 	"github.com/arm-debug/topo-cli/internal/target"
 )
+
+var passwordAuthErrorMessage = "SSH password-based authentication is not supported by topo. Please either:\n- create your own SSH keys for the target or\n- run `topo setup-keys --target <target>` to have topo generate keys for you and set up passwordless authentication"
 
 type HealthCheck struct {
 	Name    string
@@ -83,11 +87,29 @@ func GenerateReport(hostDependencies []DependencyStatus, targetStatus Status) Re
 	return report
 }
 
-func Check(sshTarget string) (Report, error) {
+func Check(sshTarget string, acceptNewHostKeys bool) (Report, error) {
 	dependencyStatuses := CheckInstalled(HostRequiredDependencies, BinaryExistsLocally)
 
-	conn := target.NewConnection(sshTarget, ssh.ExecSSH)
+	sshExists, err := BinaryExistsLocally("ssh")
+	if err != nil {
+		return GenerateReport(dependencyStatuses, Status{}), err
+	}
+
+	authProbeEnabled := sshExists
+	opts := target.ConnectionOptions{
+		AuthProbeEnabled:  authProbeEnabled,
+		AcceptNewHostKeys: acceptNewHostKeys,
+		AuthProbeInput:    os.Stdin,
+		AuthProbeOutput:   os.Stdout,
+	}
+	conn := target.NewConnection(sshTarget, ssh.ExecSSH, opts)
 	targetStatus := ProbeHealthStatus(conn)
 	report := GenerateReport(dependencyStatuses, targetStatus)
+	if err := targetStatus.AuthError; err != nil {
+		if errors.Is(err, target.ErrPasswordAuthenticationRequired) {
+			return report, errors.New(passwordAuthErrorMessage)
+		}
+		return report, err
+	}
 	return report, nil
 }
--- a/internal/health/health.go
+++ b/internal/health/health.go
@@ -9,7 +9,9 @@ import (
 	"github.com/arm-debug/topo-cli/internal/target"
 )
 
-var passwordAuthErrorMessage = "SSH password-based authentication is not supported by topo. Please either:\n- create your own SSH keys for the target or\n- run `topo setup-keys --target <target>` to have topo generate keys for you and set up passwordless authentication"
+const passwordAuthErrorMessage = `Topo does not support SSH password-based authentication. To connect, either:
+- create your own SSH keys for the target, or
+- run 'topo setup-keys --target <target>' to let Topo generate keys and configure passwordless authentication`
 
 type HealthCheck struct {
 	Name    string
--- a/internal/health/health.go
+++ b/internal/health/health.go
@@ -109,7 +109,7 @@ func Check(sshTarget string, acceptNewHostKeys bool) (Report, error) {
 	targetStatus := ProbeHealthStatus(conn)
 	report := GenerateReport(dependencyStatuses, targetStatus)
 	if err := targetStatus.AuthError; err != nil {
-		if errors.Is(err, target.ErrPasswordAuthenticationRequired) {
+		if errors.Is(err, target.ErrPasswordAuthentication) {
 			return report, errors.New(passwordAuthErrorMessage)
 		}
 		return report, err
--- a/internal/health/status.go
+++ b/internal/health/status.go
@@ -20,6 +20,7 @@ func (h HardwareProfile) Capabilities() map[HardwareCapability]struct{} {
 type Status struct {
 	SSHTarget       ssh.Host
 	ConnectionError error
+	AuthError       error
 	Dependencies    []DependencyStatus
 	Hardware        HardwareProfile
 }
@@ -27,6 +28,12 @@ type Status struct {
 func ProbeHealthStatus(c target.Connection) Status {
 	var status Status
 	status.SSHTarget = c.SSHTarget
+
+	if err := c.ProbeAuthentication(); err != nil {
+		status.AuthError = err
+		status.ConnectionError = err
+		return status
+	}
 
 	if err := c.ProbeConnection(); err != nil {
 		status.ConnectionError = err
--- a/internal/health/status.go
+++ b/internal/health/status.go
@@ -31,7 +31,6 @@ func ProbeHealthStatus(c target.Connection) Status {
 
 	if err := c.ProbeAuthentication(); err != nil {
 		status.AuthError = err
-		status.ConnectionError = err
 		return status
 	}
 
--- a/internal/target/connection.go
+++ b/internal/target/connection.go
@@ -1,11 +1,9 @@
 package target
 
 import (
-	"bytes"
 	"errors"
 	"fmt"
 	"io"
-	"os/exec"
 	"slices"
 	"strings"
 
@@ -75,7 +73,7 @@ func (c *Connection) ProbeAuthentication() error {
 	}
 
 	if !c.opts.AcceptNewHostKeys {
-		if err := c.ensureKnownHost(c.opts.AuthProbeInput, c.opts.AuthProbeOutput); err != nil {
+		if err := c.ensureKnownHost(); err != nil {
 			return err
 		}
 	}
@@ -91,56 +89,40 @@ func (c *Connection) ProbeAuthentication() error {
 }
 
 var ErrHostKeyVerification = errors.New("ssh host key verification failed")
+var ErrAuthenticationFailure = errors.New("ssh authentication failed")
 
 func (c *Connection) isPasswordAuthenticated() (bool, error) {
+	// If public key auth succeeds, the target doesn't require password auth.
 	publicArgs := slices.Clone(publicKeyProbeArgs)
 	if c.opts.AcceptNewHostKeys {
 		publicArgs = slices.Concat(publicArgs, acceptNewHostKeyArgs)
 	}
-	publicOut, publicErr := c.runSSHProbe(publicArgs)
-	if publicErr == nil {
+	if err := c.runSSHProbe(publicArgs); err == nil {
 		return false, nil
-	}
-	if isHostKeyVerificationFailure(publicOut) {
-		return false, ErrHostKeyVerification
+	} else if !errors.Is(err, ErrAuthenticationFailure) {
+		return false, err
 	}
 
+	// Public key was rejected. Check if the target accepts password auth.
 	passwordArgs := slices.Clone(passwordProbeArgs)
 	if c.opts.AcceptNewHostKeys {
 		passwordArgs = slices.Concat(passwordArgs, acceptNewHostKeyArgs)
 	}
-	passwordOut, passwordErr := c.runSSHProbe(passwordArgs)
-	if passwordErr == nil {
+	if err := c.runSSHProbe(passwordArgs); err == nil {
 		return false, nil
-	}
-	if isHostKeyVerificationFailure(passwordOut) {
-		return false, ErrHostKeyVerification
-	}
-
-	if isAuthenticationFailure(passwordOut) {
+	} else if errors.Is(err, ErrAuthenticationFailure) {
 		return true, nil
+	} else {
+		return false, err
 	}
-
-	return false, fmt.Errorf("ssh probe failed: %w", passwordErr)
 }
 
-func (c *Connection) ensureKnownHost(in io.Reader, out io.Writer) error {
-	args := slices.Concat(knownHostProbeArgs, []string{string(c.SSHTarget), "true"})
-	cmd := exec.Command("ssh", args...)
-	cmd.Stdin = in
-	var buf bytes.Buffer
-	writer := io.Writer(&buf)
-	if out != nil {
-		writer = io.MultiWriter(out, &buf)
-	}
-	cmd.Stdout = writer
-	cmd.Stderr = writer
-	if err := cmd.Run(); err != nil {
-		output := buf.String()
+func (c *Connection) ensureKnownHost() error {
+	if err := c.runSSHProbe(knownHostProbeArgs); err != nil {
 		switch {
-		case isHostKeyVerificationFailure(output):
+		case errors.Is(err, ErrHostKeyVerification):
 			return ErrHostKeyVerification
-		case isAuthenticationFailure(output):
+		case errors.Is(err, ErrAuthenticationFailure):
 			return nil
 		default:
 			return err
@@ -149,9 +131,19 @@ func (c *Connection) ensureKnownHost(in io.Reader, out io.Writer) error {
 	return nil
 }
 
-func (c *Connection) runSSHProbe(sshArgs []string) (string, error) {
+func (c *Connection) runSSHProbe(sshArgs []string) error {
 	stdout, stderr, err := ssh.Exec(c.SSHTarget, "true", nil, sshArgs...)
-	return stdout + stderr, err
+	if err == nil {
+		return nil
+	}
+	output := stdout + stderr
+	if isHostKeyVerificationFailure(output) {
+		return ErrHostKeyVerification
+	}
+	if isAuthenticationFailure(output) {
+		return ErrAuthenticationFailure
+	}
+	return fmt.Errorf("ssh probe failed: %w", err)
 }
 
 func isHostKeyVerificationFailure(output string) bool {
--- a/internal/target/connection.go
+++ b/internal/target/connection.go
@@ -88,8 +88,10 @@ func (c *Connection) ProbeAuthentication() error {
 	return nil
 }
 
-var ErrHostKeyVerification = errors.New("ssh host key verification failed")
-var ErrAuthenticationFailure = errors.New("ssh authentication failed")
+var (
+	ErrHostKeyVerification   = errors.New("ssh host key verification failed")
+	ErrAuthenticationFailure = errors.New("ssh authentication failed")
+)
 
 func (c *Connection) isPasswordAuthenticated() (bool, error) {
 	// If public key auth succeeds, the target doesn't require password auth.
--- a/internal/target/connection.go
+++ b/internal/target/connection.go
@@ -125,21 +125,12 @@ func (c *Connection) runSSHProbe(sshArgs []string) error {
 		return nil
 	}
 	output := stdout + stderr
-	if isHostKeyVerificationFailure(output) {
+	output = strings.ToLower(output)
+	if strings.Contains(output, "host key verification failed") {
 		return ErrHostKeyVerification
 	}
-	if isAuthenticationFailure(output) {
+	if strings.Contains(output, "permission denied") || strings.Contains(output, "authentication failed") || strings.Contains(output, "password") {
 		return ErrAuthenticationFailure
 	}
 	return fmt.Errorf("ssh probe failed: %w", err)
-}
-
-func isHostKeyVerificationFailure(output string) bool {
-	lower := strings.ToLower(output)
-	return strings.Contains(lower, "host key verification failed")
-}
-
-func isAuthenticationFailure(output string) bool {
-	lower := strings.ToLower(output)
-	return strings.Contains(lower, "permission denied") || strings.Contains(lower, "authentication failed") || strings.Contains(lower, "password")
 }
--- a/internal/target/connection.go
+++ b/internal/target/connection.go
@@ -46,7 +46,7 @@ type ConnectionOptions struct {
 	AuthProbeOutput   io.Writer
 }
 
-var ErrPasswordAuthenticationRequired = errors.New("password authentication required")
+var ErrPasswordAuthentication = errors.New("password authentication required")
 
 func NewConnection(sshTarget string, exec execSSH, opts ConnectionOptions) Connection {
 	return Connection{
@@ -84,7 +84,7 @@ func (c *Connection) ProbeAuthentication() error {
 		return err
 	}
 	if needsSetup {
-		return ErrPasswordAuthenticationRequired
+		return ErrPasswordAuthentication
 	}
 	return nil
 }
--- a/internal/target/connection.go
+++ b/internal/target/connection.go
@@ -1,9 +1,34 @@
 package target
 
 import (
+	"bytes"
+	"errors"
 	"fmt"
+	"io"
+	"os/exec"
+	"strings"
 
 	"github.com/arm-debug/topo-cli/internal/ssh"
+)
+
+var (
+	publicKeyProbeArgs = []string{
+		"-o", "BatchMode=yes",
+		"-o", "PreferredAuthentications=publickey",
+	}
+	passwordProbeArgs = []string{
+		"-o", "BatchMode=yes",
+		"-o", "PreferredAuthentications=password",
+		"-o", "NumberOfPasswordPrompts=0",
+	}
+	knownHostProbeArgs = []string{
+		"-o", "PreferredAuthentications=publickey",
+		"-o", "PasswordAuthentication=no",
+		"-o", "NumberOfPasswordPrompts=0",
+	}
+	acceptNewHostKeyArgs = []string{
+		"-o", "StrictHostKeyChecking=accept-new",
+	}
 )
 
 type execSSH func(target ssh.Host, command string) (string, error)
@@ -11,12 +36,23 @@ type execSSH func(target ssh.Host, command string) (string, error)
 type Connection struct {
 	SSHTarget ssh.Host
 	exec      execSSH
+	opts      ConnectionOptions
 }
 
-func NewConnection(sshTarget string, exec execSSH) Connection {
+type ConnectionOptions struct {
+	AuthProbeEnabled  bool
+	AcceptNewHostKeys bool
+	AuthProbeInput    io.Reader
+	AuthProbeOutput   io.Writer
+}
+
+var ErrPasswordAuthenticationRequired = errors.New("password authentication required")
+
+func NewConnection(sshTarget string, exec execSSH, opts ConnectionOptions) Connection {
 	return Connection{
 		SSHTarget: ssh.Host(sshTarget),
 		exec:      exec,
+		opts:      opts,
 	}
 }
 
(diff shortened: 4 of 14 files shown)


tgonzalezorlandoarm:
@awphi I still owe you the abstraction for commands from the `setup-keys` command. The same applies here for the ssh commands as far as I can tell, so I'll group all of those changes (for both PRs) in a new PR that will come soon today.

internal/ssh/auth_detection.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	if isAuthenticationFailure(passwordOut) {
+		return true, nil
+	}
+
+	return false, fmt.Errorf("ssh probe failed: %w", passwordErr)
+}
+
+func detectWithKnownHostCheck(target Host, in io.Reader, out io.Writer) (bool, error) {
```
logan-scully-arm:
```suggestion
func detectWithKnownHostCheck(target Host, in io.Reader, out io.Writer) (bool, error) {
	for range 2 {
		needsSetup, err := detectPasswordAuthentication(target)
		if !errors.Is(err, ErrHostKeyVerification) {
		    return needsSetup, err
        }
		if err := ensureKnownHost(target, in, out); err != nil {
		    return false, err
		}
	}
	return false, ErrHostKeyVerification
}

```

internal/ssh/auth_detection.go
```diff
@@ -0,0 +1,115 @@
...
+		"-o", "PasswordAuthentication=no",
+		"-o", "NumberOfPasswordPrompts=0",
+	}
+)
+
+var ErrHostKeyVerification = errors.New("ssh host key verification failed")
+
+func IsPasswordAuthentication(target Host, in io.Reader, out io.Writer) (bool, error) {
+	if target.IsPlainLocalhost() {
+		return false, nil
```
logan-scully-arm: I don't think this is right, Although I might be misunderstanding whats happening.

If I try `ssh localhost` on my own machine rn, it'll ask for the password.

internal/ssh/auth_detection.go
```diff
@@ -0,0 +1,115 @@
...
+		case isAuthenticationFailure(output):
+			return nil
+		default:
+			return err
+		}
+	}
+	return nil
+}
+
+func runSSHProbe(target Host, sshArgs []string) (string, error) {
```
logan-scully-arm: I think connection may already cover part of this behavior, so it might be worth leaning more on Connection directly in a few of these functions (instead of using Host).

Although I have noticed that the stock `Exec()` function is very similar to `ExecSSH` that is used in connection. There might be reason to replace `ExecSSH` entirely with Exec(). 

You could then replace this with `conn.Run(...)` . Which can be used to test some of these functions.

internal/health/dependencies.go
```diff
@@ -29,6 +29,8 @@ type Dependency struct {
 }
 
 var HostRequiredDependencies = []Dependency{
+	// The first dependency is expected to be SSH, as the health check relies on it to determine the SSH authentication method.
+	// If this changes, the health check logic will need to be updated accordingly.
```
logan-scully-arm: You don't need these comments. SSH is a hard dependency of Topo, you don't need to document why here.

internal/health/health.go
```diff
@@ -102,6 +106,20 @@ func GenerateReport(hostDependencies []DependencyStatus, targetStatus Status) Re
 func Check(sshTarget string) (Report, error) {
 	dependencyStatuses := CheckInstalled(HostRequiredDependencies, BinaryExistsLocally)
 
+	// The first dependency is expected to be SSH
```
logan-scully-arm: Remove this comment too.

internal/health/health.go
```diff
@@ -102,6 +106,20 @@ func GenerateReport(hostDependencies []DependencyStatus, targetStatus Status) Re
...
+	if dependencyStatuses[0].Installed {
```
logan-scully-arm:
```suggestion
	if !dependencyStatuses[0].Installed {
	 return report := GenerateReport(dependencyStatuses, Status{}), nil
	}
```
You might need to init a targetStatus here instead.

IMO if you can ever early exit rather than nesting conditionals, you always should

awphi: Not sure about the magic number `0` here - would much rather see this as an explicit check. `BinaryExistsLocally` shouldn't be desperately expensive

internal/health/health.go
```diff
@@ -1,11 +1,15 @@
...
 
 import (
+	"errors"
+	"os"
 	"strings"
 
 	"github.com/arm-debug/topo-cli/internal/ssh"
 )
 
+const passwordAuthErrorMessage = "SSH password-based authentication is not supported by topo. Please either:\n- create your own SSH keys for the target or\n- run `topo setup-keys --target <target>` to have topo generate keys for you and set up passwordless authentication"
```
awphi: nit: would suggest a package-level error `var` instead of constructing a new one from this string in each time - makes comparison easy if we ever export this error var

logan-scully-arm (changes requested):
I think it would be nice to see auth_detection_test -- although it might be superficial?

It might need another look over, but I think you should use `Connection` and `Connection.Run` if possible

internal/ssh/auth_detection.go
awphi, line 1: Keen to see some tests for this

internal/health/health.go
logan-scully-arm: Not 100% sure, but  I think that nil should be an error? (I can't remember what happens on the report side)

internal/ssh/auth_detection.go
```diff
@@ -0,0 +1,115 @@
...
+)
+
+var ErrHostKeyVerification = errors.New("ssh host key verification failed")
+
+func IsPasswordAuthentication(target Host, in io.Reader, out io.Writer) (bool, error) {
+	if target.IsPlainLocalhost() {
+		return false, nil
+	}
+	return detectWithKnownHostCheck(target, in, out)
+}
```
awphi: Would this make more sense as a method on `ssh.Host`
```go
ssh.Host(sshTarget).IsPasswordAuthenticated(...)
// reads better to me than
ssh.IsPasswordAuthentication(ssh.Host(sshTarget), ...)
```

internal/health/dependencies.go
tgonzalezorlandoarm: I need to document that it should be the **first** dependency though, otherwise this will break the health check.

internal/health/health.go
tgonzalezorlandoarm: Does this makes sense to you? https://github.com/arm/topo/pull/17#discussion_r2799804940

internal/ssh/auth_detection.go
```diff
@@ -0,0 +1,115 @@
...
+				return false, err
+			}
+			continue
+		}
+		return needsSetup, err
+	}
+	return false, ErrHostKeyVerification
+}
+
+func ensureKnownHost(target Host, in io.Reader, out io.Writer) error {
```
awphi: Could we avoid the known host checking by adding `-o StrictHostKeyChecking=accept-new` to our probes?

internal/ssh/auth_detection.go
logan-scully-arm: This might actually be an issue with the rest of health and other parts of connection, as they still run with `ExecSSH`.  LMK what you think

internal/ssh/auth_detection.go
```diff
@@ -0,0 +1,115 @@
...
+		needsSetup, err := detectPasswordAuthentication(target)
+		if errors.Is(err, ErrHostKeyVerification) {
+			if err := ensureKnownHost(target, in, out); err != nil {
+				return false, err
+			}
+			continue
+		}
+		return needsSetup, err
+	}
+	return false, ErrHostKeyVerification
```
awphi: Found the flow of try detect -> check if error was known host-related -> add host key -> try detect again a bit tricky to follow - could we try https://github.com/arm/topo/pull/17/changes#r2799828712 or simplifying this function to always ensure host key exists first then try detection pw auth once.

tgonzalezorlandoarm:
> but I think you should use `Connection` and `Connection.Run` if possible

Very good point! Will get the logic inside Probe actually

internal/health/health.go
```diff
@@ -102,6 +106,20 @@ func GenerateReport(hostDependencies []DependencyStatus, targetStatus Status) Re
...
+		if err != nil {
+			return GenerateReport(dependencyStatuses, Status{}), err
+		}
+		if needsSetup {
+			return GenerateReport(dependencyStatuses, Status{}), errors.New(passwordAuthErrorMessage)
+		}
+	} else {
+		report := GenerateReport(dependencyStatuses, Status{})
+		return report, nil
+	}
```
awphi: Don't think this `else` block is necessary - if ssh isn't installed `conn.Probe` should fail and set its own `ConnectionError` prop properly.

awphi (changes requested):

internal/health/health.go
tgonzalezorlandoarm: done: https://github.com/arm/topo/pull/17/commits/b56b2fce65e2ed15b14a86b6a8669662a0a43d58

internal/health/health.go
```diff
@@ -107,6 +111,20 @@ func GenerateReport(hostDependencies []DependencyStatus, targetStatus Status) Re
 func Check(sshTarget string) (Report, error) {
 	dependencyStatuses := CheckInstalled(HostRequiredDependencies, BinaryExistsLocally)
 
+	sshExists, _ := BinaryExistsLocally("ssh")
+	if sshExists {
```
logan-scully-arm:
```suggestion
	if !sshExists {
	    report := GenerateReport(dependencyStatuses, Status{})
		return report, nil
	}
	needsSetup, err := ssh.IsPasswordAuthenticated(ssh.Host(sshTarget), os.Stdin, os.Stdout)
	if err != nil {
		return GenerateReport(dependencyStatuses, Status{}), err
	}
	if needsSetup {
		return GenerateReport(dependencyStatuses, Status{}), errors.New(passwordAuthErrorMessage)
	}
```

Although I think @awphi is right, conn.Probe() should handle that correctly itself, so you might be better off keeping it the same and removing the else.

internal/health/health.go
tgonzalezorlandoarm: done: https://github.com/arm/topo/pull/17/commits/144bf942906ab48d3a46a182dcac252924e6963c

internal/health/dependencies.go
tgonzalezorlandoarm: This is solved in this way instead https://github.com/arm/topo/pull/17#discussion_r2799693002.
Marking this thread as resolved as a result.

internal/health/health.go
tgonzalezorlandoarm: done: https://github.com/arm/topo/pull/17/commits/bee9d901cc85cbd25a18dfeb1bb89a7fff78804a

internal/ssh/auth_detection.go
tgonzalezorlandoarm: We could, but we are taking a security decision here on behalf of the user, so we'd need to inform them and ask for consent, which is what ssh does.
I think following ssh on this regard is the best option here.

tgonzalezorlandoarm: As you say, it would simplify the logic a lot though, which is why I'd like to apply this. But I think in this case we should side with the security decisions of ssh instead of our own.
WDYT?

tgonzalezorlandoarm: We could add this as an extra flag though? Otherwise this would become interactive

internal/ssh/auth_detection.go
tgonzalezorlandoarm: I chose this path for the reasons stated in the thread pointed at by you:

> simplifying this function to always ensure host key exists first then try detection pw auth once.

done here: https://github.com/arm/topo/pull/17/commits/2786331b4b897a73b210acddcc55e2d0f0161566

We can discuss further there, but I think this is resolved here?

internal/ssh/auth_detection.go
tgonzalezorlandoarm: Added the flag here:
https://github.com/arm/topo/pull/17/commits/ab8b285a588645a0c78ee66cf28510877dbf51be

internal/health/health.go
tgonzalezorlandoarm: why the `!sshExists` instead  of `sshExists` ?
I think this is resolved here? https://github.com/arm/topo/pull/17/commits/144bf942906ab48d3a46a182dcac252924e6963c

logan-scully-arm: Not necessary, but I generally prefer avoiding else when possible. You can often invert the condition and return early, which I find a bit more readable.

But since it was unneeded your change here is good

internal/ssh/auth_detection.go
tgonzalezorlandoarm: I think I prefer Logan's suggestion to use `Connection` if that's okay with you?
I'll open a new PR for this refactoring as this probably deserves some splitting

awphi: Definitely makes sense to me. I've got a PR open in a very related area atm #19 which pulls `Connection` out of health and into a new `target` package

tgonzalezorlandoarm: Thanks for the heads up! Will probably rebase on top of yours then!
I'm resolving here, let me know if I've misunderstood and there's something else to discuss in this thread

internal/ssh/auth_detection.go
tgonzalezorlandoarm: Refactored this as indicated by Adam

internal/ssh/auth_detection.go
tgonzalezorlandoarm: done: https://github.com/arm/topo/pull/17/commits/6d735b9a5cbc5841f2a83a7d3f9cd16f3b2f62b0

tgonzalezorlandoarm:
> I think it would be nice to see auth_detection_test

done: https://github.com/arm/topo/pull/17/commits/6d735b9a5cbc5841f2a83a7d3f9cd16f3b2f62b0

internal/ssh/auth_detection.go
tgonzalezorlandoarm: What you say makes sense, I'll add these changes in a new PR that uses #19

internal/ssh/auth_detection.go
tgonzalezorlandoarm: Could you tell me what happens if you do:
```
$ ./topo health --target localhost
Host
----
SSH: ✅ (ssh)
Container Engine: ✅ (docker)

Target
------
Subsystem Driver (remoteproc): ❌
```

internal/ssh/auth_detection_test.go
```diff
@@ -0,0 +1,277 @@
...
+		t.Fatalf("read args: %v", err)
+	}
+	lines := strings.Split(strings.TrimSpace(string(data)), "\n")
+	if len(lines) == 1 && lines[0] == "" {
+		return nil
+	}
+	return lines
+}
+
+func TestIsPasswordAuthenticated_Localhost(t *testing.T) {
```
logan-scully-arm: For tests, I think to line up with the codebases conventions this should be
```suggestion
TestIsPasswordAuthenticated(t *testing.T) {
 t.Run("correctly returns false when it is localhost", func(t *testing.T) { ...
}
```

And then all of the tests for that function go within

logan-scully-arm: Another thing to consider is if there Is a way to test this via mocking. The `Connection` struct that I was on about before should have some good established ways of handling this.

I'm really not a fan of `fake_ssh.sh`

internal/testutil/testdata/fake_ssh.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/sh
```
logan-scully-arm, line 1: I think you should refactor to remove this file. It might be good to write separate e2e tests to evaluate this, but you should default to testing raw functionality in the package tests.

If at any point you do need a file like this, I think it would be nice to try and use GO if possible to maintain a single language throughout.

tgonzalezorlandoarm: > but you should default to testing raw functionality in the package tests.

I think I'm doing this. In this case, the `.sh` file is just returning what is being indicated by the test that it should return.
I'm open to suggestions of how else I could test the password detection functionality

> it would be nice to try and use GO if possible to maintain a single language throughout.

The problem with this is we'd need to compile this extra Go file before testing

internal/ssh/auth_detection_test.go
```diff
@@ -0,0 +1,277 @@
...
+	got, err := IsPasswordAuthenticated(Host("localhost"), nil, nil, false)
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	if got {
+		t.Fatalf("expected localhost to not require setup")
```
logan-scully-arm: You shouldn't be using fatal on these tests. We normally use the assert package across all of our tests. This should really be something like 
```
require.NoError(t, err)
assert.False(t, got)
```

internal/ssh/auth_detection_test.go
```diff
@@ -0,0 +1,277 @@
+package ssh
```
logan-scully-arm: All of our tests are black box rather than white box testing. This means that we only test publicly visible functions -- Things like `detectPasswordAuthentication` should only be tested through various conditions passed through`IsPasswordAuthenticated`

tgonzalezorlandoarm: Why is that the case ?
And also, why shouldn't we have an extra level of testing?

internal/testutil/testdata/fake_ssh.sh
logan-scully-arm: I think if you use stuff brought in with `Connection` you can mock ExecSSH which can then mimic desired output. 

On the second point, you can use `go run` to execute code without building.

internal/ssh/auth_detection_test.go
logan-scully-arm: I believe because it validates the system from the outside, the same way real users or external systems interact with it. Although you'd have to talk to @muchzill4 for more of an explanation.

As it's the done-thing across the repo I'd recommend sticking to it here.

internal/ssh/auth_detection_test.go
tgonzalezorlandoarm: done https://github.com/arm/topo/pull/17/commits/8a265ecc6a505865061bfcf399a21c33ca33b85d

internal/ssh/auth_detection.go
logan-scully-arm: It asks for my password. 

```
 1 logscu01@e138103:~/Projects/TOPO/topo$ ./topo health --target localhost
logscu01@localhost's password: 
logscu01@localhost's password: 
logscu01@localhost's password: 
```
Because `Connection` uses ssh regardless of if it's `localhost`, it will still try and password authenticate. This really requires a separate refactor which replaces `ExecSSH` entirely with `Exec` which actually has localhost-safety.

internal/testutil/testdata/fake_ssh.sh
tgonzalezorlandoarm: `go run` does still build as far as I'm aware, it's just hiding it

tgonzalezorlandoarm: > Connection you can mock ExecSSH

yes, this makes sense, thanks

internal/ssh/auth_detection_test.go
tgonzalezorlandoarm: done: https://github.com/arm/topo/pull/17/commits/31e10c26dee11eaa3e95e21c3078a2e6aefcf008

logan-scully-arm: I think the main argument is that lots of the functionality is set in stone, whereas implementation in the codebase may change under-the-hood. This makes it easier for us to do those refactors, as we are only testing the top level package output.

internal/ssh/auth_detection_test.go
```diff
@@ -0,0 +1,180 @@
...
```
logan-scully-arm:
```suggestion
package ssh_test
```

tgonzalezorlandoarm: done: https://github.com/arm/topo/pull/17/commits/1db56f6ef7f048ae352cdbf1ee61e1bae1bbb885

internal/ssh/auth_detection_test.go
tgonzalezorlandoarm: Having dscussed with you offline, I'm completely onboard with teh mockExecSSH option.
I'll do this in a follow-up PR if it's ok with you

internal/ssh/auth_detection.go
tgonzalezorlandoarm: Removed the assumption here, will deal with teh ExecSSH change in a future PR as discussed

cmd/topo/health.go
```diff
@@ -35,5 +38,6 @@ var healthCmd = &cobra.Command{
 
 func init() {
 	addTargetFlag(healthCmd, &healthTarget)
+	healthCmd.Flags().BoolVar(&healthAcceptNewHostKey, "accept-new-host-keys", false, "Automatically trust and add new SSH host keys for the target")
```
federicobozzini: Would it make sense to make this var-less following #16 ?

internal/health/health.go
```diff
@@ -83,9 +87,20 @@ func GenerateReport(hostDependencies []DependencyStatus, targetStatus Status) Re
 	return report
 }
 
-func Check(sshTarget string) (Report, error) {
+func Check(sshTarget string, acceptNewHostKeys bool) (Report, error) {
```
federicobozzini, line 92: While the `sshTarget` is quite a foundamental piece of this function, `acceptNewHostKeys` seems more like an option. How about changing the second argument to be a struct of options instead of a boolean value?

internal/health/health.go
```diff
@@ -83,9 +87,20 @@ func GenerateReport(hostDependencies []DependencyStatus, targetStatus Status) Re
...
 	dependencyStatuses := CheckInstalled(HostRequiredDependencies, BinaryExistsLocally)
 
+	sshExists, _ := BinaryExistsLocally("ssh")
```
federicobozzini: Any reason to ignore the error instead of checking it?

internal/health/health.go
```diff
@@ -83,9 +87,20 @@ func GenerateReport(hostDependencies []DependencyStatus, targetStatus Status) Re
...
+	if sshExists {
+		needsSetup, err := ssh.IsPasswordAuthenticated(ssh.Host(sshTarget), os.Stdin, os.Stdout, acceptNewHostKeys)
+		if err != nil {
+			return GenerateReport(dependencyStatuses, Status{}), err
+		}
+		if needsSetup {
+			return GenerateReport(dependencyStatuses, Status{}), errors.New(passwordAuthErrorMessage)
+		}
```
federicobozzini: Just to understand the logic, why are we OK here returning a partial report with valid `hostDependencies` but not `targetStatus`?

As in, are we sure we can't get the `targetStatus` or we just don't want to include it?

internal/health/health.go
tgonzalezorlandoarm: done

internal/health/health.go
tgonzalezorlandoarm: Only place we are returning a partial report now in is when binary doesn't exist locally

internal/health/health.go
tgonzalezorlandoarm: done. I've forced push now so probably better to look at the Files changed tab

internal/health/health.go
```diff
@@ -1,12 +1,16 @@
...
 import (
+	"errors"
+	"os"
 	"strings"
 
 	"github.com/arm-debug/topo-cli/internal/ssh"
 	"github.com/arm-debug/topo-cli/internal/target"
 )
 
+var passwordAuthErrorMessage = "SSH password-based authentication is not supported by topo. Please either:\n- create your own SSH keys for the target or\n- run `topo setup-keys --target <target>` to have topo generate keys for you and set up passwordless authentication"
```
th3james: - const > var
- multi-line strings are more readable
- more active voice
- I think Topo should be in title case
```suggestion
const passwordAuthErrorMessage = `Topo does not support SSH password-based authentication. To connect, either:
- create your own SSH keys for the target, or
- run 'topo setup-keys --target <target>' to let Topo generate keys and configure passwordless authentication`
```

internal/health/health.go
```diff
@@ -83,11 +87,29 @@ func GenerateReport(hostDependencies []DependencyStatus, targetStatus Status) Re
...
+func Check(sshTarget string, acceptNewHostKeys bool) (Report, error) {
 	dependencyStatuses := CheckInstalled(HostRequiredDependencies, BinaryExistsLocally)
 
-	conn := target.NewConnection(sshTarget, ssh.ExecSSH)
+	sshExists, err := BinaryExistsLocally("ssh")
+	if err != nil {
+		return GenerateReport(dependencyStatuses, Status{}), err
+	}
+
+	authProbeEnabled := sshExists
```
th3james: You already have this information in memory. You could consider extracting an `isInstalled` helper
```suggestion
	authProbeEnabled := false
	for _, s := range dependencyStatuses {
		if s.Dependency.Name == "ssh" {
			authProbeEnabled = s.Installed
			break
		}
	}
```

internal/target/connection.go
```diff
@@ -31,3 +67,102 @@ func (c *Connection) BinaryExists(bin string) (bool, error) {
...
+	}
+	return nil
+}
+
+var ErrHostKeyVerification = errors.New("ssh host key verification failed")
+
+func (c *Connection) IsPasswordAuthenticated() (bool, error) {
+	args := []string{}
+	if !c.opts.AcceptNewHostKeys {
+		if err := c.ensureKnownHost(c.opts.AuthProbeInput, c.opts.AuthProbeOutput); err != nil {
```
th3james: IMHO this is the wrong place for this check - I'd move this one layer up the stack to `ProbeAuthentication`, then merge `detectPasswordAuthentication` into this function.

internal/health/status.go
```diff
@@ -28,6 +29,12 @@ func ProbeHealthStatus(c target.Connection) Status {
 	var status Status
 	status.SSHTarget = c.SSHTarget
 
+	if err := c.ProbeAuthentication(); err != nil {
+		status.AuthError = err
+		status.ConnectionError = err
```
th3james: This doesn't seem semantically correct?
```suggestion
```

internal/target/connection.go
```diff
@@ -1,22 +1,58 @@
...
+
+type ConnectionOptions struct {
+	AuthProbeEnabled  bool
+	AcceptNewHostKeys bool
+	AuthProbeInput    io.Reader
+	AuthProbeOutput   io.Writer
 }
 
-func NewConnection(sshTarget string, exec execSSH) Connection {
+var ErrPasswordAuthenticationRequired = errors.New("password authentication required")
```
th3james: This variable name doesn't read right to me - I would interpret it as telling me I need to use password authentication

```suggestion
var ErrPasswordAuthentication = errors.New("password authentication required")
```

internal/target/connection.go
```diff
@@ -31,3 +67,102 @@ func (c *Connection) BinaryExists(bin string) (bool, error) {
...
+	} else {
+		args = append(args, acceptNewHostKeyArgs...)
+	}
+
+	return c.detectPasswordAuthentication(args)
+}
+
+func (c *Connection) detectPasswordAuthentication(extraArgs []string) (bool, error) {
+	publicArgs := append([]string{}, publicKeyProbeArgs...)
+	publicArgs = append(publicArgs, extraArgs...)
```
th3james:
```suggestion
	publicArgs := slices.Concat(publicKeyProbeArgs, extraArgs)
```

internal/target/connection.go
```diff
@@ -31,3 +67,102 @@ func (c *Connection) BinaryExists(bin string) (bool, error) {
...
+	publicOut, publicErr := c.runSSHProbe(publicArgs)
+	if publicErr == nil {
+		return false, nil
+	}
+	if isHostKeyVerificationFailure(publicOut) {
+		return false, ErrHostKeyVerification
+	}
+
+	passwordArgs := append([]string{}, passwordProbeArgs...)
+	passwordArgs = append(passwordArgs, extraArgs...)
```
th3james:
```suggestion
	passwordArgs := slices.Concat(passwordProbeArgs, extraArgs)
```

internal/target/connection.go
```diff
@@ -31,3 +67,102 @@ func (c *Connection) BinaryExists(bin string) (bool, error) {
...
+func (c *Connection) detectPasswordAuthentication(extraArgs []string) (bool, error) {
+	publicArgs := append([]string{}, publicKeyProbeArgs...)
+	publicArgs = append(publicArgs, extraArgs...)
+	publicOut, publicErr := c.runSSHProbe(publicArgs)
+	if publicErr == nil {
+		return false, nil
+	}
+	if isHostKeyVerificationFailure(publicOut) {
+		return false, ErrHostKeyVerification
+	}
```
th3james: This confused me for a while because publicErr doesn't actually contain the error, it tells you that you have to look inside publicOut, which contains both stdout and stderr. I think concatenating both stdout and stderr into just `stdout` is confusing.

Here's a more readable proposal
```go
  func (c *Connection) runSSHProbe(sshArgs []string) error {
        stdout, stderr, err := ssh.Exec(c.SSHTarget, "true", nil, sshArgs...)
        if err == nil {
                return nil
        }
        output := stdout + stderr
        if isHostKeyVerificationFailure(output) {
                return ErrHostKeyVerification
        }
        if isAuthenticationFailure(output) {
                return ErrAuthenticationFailure
        }
        return fmt.Errorf("ssh probe failed: %w", err)
  }

  func (c *Connection) detectPasswordAuthentication(extraArgs []string) (bool, error) {
        // If public key auth succeeds, the target doesn't require password auth.
        publicArgs := slices.Concat(publicKeyProbeArgs, extraArgs)
        if err := c.runSSHProbe(publicArgs); err == nil {
                return false, nil
        } else if !errors.Is(err, ErrAuthenticationFailure) {
                return false, err
        }

        // Public key was rejected. Check if the target accepts password auth.
        passwordArgs := slices.Concat(passwordProbeArgs, extraArgs)
        if err := c.runSSHProbe(passwordArgs); err == nil {
                return false, nil
        } else if errors.Is(err, ErrAuthenticationFailure) {
                return true, nil
        } else {
                return false, err
        }
  }
```

internal/health/health.go
tgonzalezorlandoarm: We had (partially) this discussion here with @awphi :
https://github.com/arm/topo/pull/17#discussion_r2799693002

I wanted to avoid looping over (even if it's just a few times) and I agreed with Adam that the hardcoded `0` was probably not the best idea so might we worth just checking the binary.
That being said, I don't feel too strongly about avoiding looping over. Done: https://github.com/arm/topo/pull/17/commits/4b155e26a69c23c2b997b90ebd6f10d5b25db7cf

internal/health/health.go
tgonzalezorlandoarm: Regarding the const, @awphi preferred the var:
https://github.com/arm/topo/pull/17#discussion_r2799700400

I'm applying democratic principles here and changing back to const!
Done: https://github.com/arm/topo/pull/17/commits/00bac9047c0fb2d1c4bbaff94ebf92b974d7d05d

tgonzalezorlandoarm: The linter doesn't like the capitalized Topo at the beginning og the phrase. Added the "note:" at the beginning to comply with it. https://github.com/arm/topo/pull/17/commits/cedc1f355e5a6f7a326723944c5d41a8ec66cb89

awphi: Ah sorry my original suggestion may have been unclear - I was suggesting an error var i.e. 
```go
var passwordAuthError = errors.New("....")
```
but it's kind of redundant since we never actually need to compare against this error which is the main advantage of the above pattern

internal/target/connection.go
logan-scully-arm: `ErrTargetRequiresPassword`

internal/health/health.go
th3james: Calling `BinaryExistsLocally` is likely to be at least a couple of orders of magnitude more costly - spawning a sub shell is  much more work than reading a value that's already in memory



- Edit: I was looking at `Connection.BinaryExists`, there's no sub shell. It still hits the filesystem though, so it is going to be much slower

internal/target/connection.go
tgonzalezorlandoarm: IMHO, you are right xD
done: https://github.com/arm/topo/pull/17/commits/5671f467056cdb52095e7aa7d0497979f110a1e2

internal/target/connection.go
tgonzalezorlandoarm: done: https://github.com/arm/topo/pull/17/commits/7aed9558cc99eec93c5000d572d50b07e6f09554

internal/target/connection.go
tgonzalezorlandoarm: https://github.com/arm/topo/pull/17/commits/73f46be48912b63ca54e0eb75003b63feb26d9d0

internal/target/connection.go
tgonzalezorlandoarm: https://github.com/arm/topo/pull/17/commits/73f46be48912b63ca54e0eb75003b63feb26d9d0

internal/target/connection.go
tgonzalezorlandoarm: That indeed improved readability, thanks. https://github.com/arm/topo/commit/373f033628aacdd1975fdf45a6f0549e8e5502cc
I was going to bin `runSSHProbe` on the next PR that does an `Exec` + `ExecSSH` refactoring. But let's see how that takes shape in that PR

internal/health/status.go
tgonzalezorlandoarm: done: https://github.com/arm/topo/pull/17/commits/a6b446ff623aa2966a0180f3aa46dde9aa0b6ff2

cmd/topo/health.go
tgonzalezorlandoarm: Do I understand correctly that that PR is for making Varless the target flag in particular?
If that's the case, that would get dealt with in Conflict resolution if it gets merged before this PR
If I'm getting the request wrong and you are talking about the healthCmd flag, how should I modify this BoolVar?

tgonzalezorlandoarm:
I think I've now addressed everyone's comments, please shout otherwise!

internal/target/connection.go
```diff
@@ -31,3 +66,94 @@ func (c *Connection) BinaryExists(bin string) (bool, error) {
...
+	} else if !errors.Is(err, ErrAuthenticationFailure) {
+		return false, err
+	}
+
+	// Public key was rejected. Check if the target accepts password auth.
+	passwordArgs := slices.Clone(passwordProbeArgs)
+	if c.opts.AcceptNewHostKeys {
+		passwordArgs = slices.Concat(passwordArgs, acceptNewHostKeyArgs)
+	}
+	if err := c.runSSHProbe(passwordArgs); err == nil {
```
th3james: if you compute `extraArgs` at the start of the function, you reduce the conditionals later
```suggestion
	var extraArgs []string
	if c.opts.AcceptNewHostKeys {
		extraArgs = acceptNewHostKeyArgs
	}

	// If public key auth succeeds, the target doesn't require password auth.
	if err := c.runSSHProbe(slices.Concat(publicKeyProbeArgs, extraArgs)); err == nil {
		return false, nil
	} else if !errors.Is(err, ErrAuthenticationFailure) {
		return false, err
	}

	// Public key was rejected. Check if the target accepts password auth.
	if err := c.runSSHProbe(slices.Concat(passwordProbeArgs, extraArgs)); err == nil {
```

internal/target/connection.go
```diff
@@ -31,3 +66,94 @@ func (c *Connection) BinaryExists(bin string) (bool, error) {
...
+func (c *Connection) ensureKnownHost() error {
+	if err := c.runSSHProbe(knownHostProbeArgs); err != nil {
+		switch {
+		case errors.Is(err, ErrHostKeyVerification):
+			return ErrHostKeyVerification
+		case errors.Is(err, ErrAuthenticationFailure):
+			return nil
+		default:
+			return err
+		}
```
th3james: Can be simplified
```suggestion
	err := c.runSSHProbe(knownHostProbeArgs)
	if errors.Is(err, ErrAuthenticationFailure) {
		return nil
```

internal/target/connection.go
tgonzalezorlandoarm: done: https://github.com/arm/topo/pull/17/commits/3f2f8093c72faaf31d23aabdce7bd959c4402ce9

internal/target/connection.go
tgonzalezorlandoarm: done: https://github.com/arm/topo/pull/17/commits/84a45ac5fca0276ba0610828afad3424f08d6122

cmd/topo/health.go
federicobozzini: What I'm saying is to use

```go
healthCmd.Flags().Bool("accept-new-host-keys", false, "Automatically trust and add new SSH host keys for the target")
```

this would return a new value, removing th need to pass `healthAcceptNewHostKey` as an argument.

tgonzalezorlandoarm: Thanks a lot, I understand now.
I have to admit this goes against my taste because we have a lot of extra logic (even an almost unreachable panic!) just to get rid of a variable we end up using once, but I understand yours and James' POV.
Done: https://github.com/arm/topo/pull/17/commits/e84359dc60e8a0bb3077decdadfdc0387adef6ad

federicobozzini: I would add that maybe we can avoid checking the error entirely :fearful:.

It seems to be a common pattern in other codebases:

- https://github.com/docker/compose/blob/f9828dfab909e9dd0dd489a49088c8619ec2ca7e/cmd/compose/compose.go#L497
- https://github.com/argoproj/argo-workflows/blob/465af6839a5b6cee266981016bb3100142e1bebf/hack/featuregen/main.go#L48
- https://github.com/zalando/postgres-operator/blob/master/kubectl-pg/cmd/logs.go?utm_source=chatgpt.com#L42
- https://github.com/cosmos/cosmos-sdk/blob/main/client/keys/mnemonic.go?utm_source=chatgpt.com#L29


ChatGPT seems to agree that some (most?) large pragmatic codebases don't really care.

Unless I'm not mistaken the line of code that `panic`s cannot literally be executed, so it's just useless.

tgonzalezorlandoarm: Let's leave it that way, for now as discussed offline. Thanks for the input Federico!

yejseo01 (approved):
Looks Good to me!

logan-scully-arm (approved):
Looks good now, rlly good work w/ this.

Approving w/ understanding that shell script will be removed in future PR.

awphi (approved):

PR #19, opened 2026-02-16
refactor: pull target connection + probe out of health

awphi:
## Changes

- Moves common target-connection utilities (establishing a connection, probing remoteproc, probing host hardware + extracting arm features) to a new `target` package, out of `health`.
  - This package is consumed widely by tests and as needed `templates`, `describe` and `health` so they each do the minimum amount of work required to function. This also means topo health will correctly detect remoteproc even when lscpu is missing!
  - This should make it easier to add more probes that can be picked-and-chosen by consuming commands in the future without tightly coupling them to any specific command.

Apologies for the big PR but it's 95% just moving existing stuff about and a bit of plumbing around the edges!


internal/target/connection.go
```diff
@@ -0,0 +1,33 @@
+package target
+
+import (
+	"fmt"
+
+	"github.com/arm-debug/topo-cli/internal/ssh"
+)
+
+type execSSH func(target ssh.Host, command string) (string, error)
```
logan-scully-arm, line 9: Possibly a concern of this PR, but lmk if you disagree. 

In the ssh package there's a lot of forms of the 'Exec' function which don't need to be there IMO. For example theres `ExecSSH` and `Exec` which appear to be almost identical.

I can do the PR for this after if you don't think it should be a part of this one.

awphi: I think that sounds like a good improvement but keen to minimize the blast radius of this PR as much as possible as it's already quite beefy!

logan-scully-arm: Yeah fair dos :+1: IMO this looks good then.

logan-scully-arm (approved):
LGTM 👍 Possibly check w/ @tgonzalezorlandoarm before merging.

tgonzalezorlandoarm:
> LGTM 👍 Possibly check w/ @tgonzalezorlandoarm before merging.

Taking a look now!

tgonzalezorlandoarm (approved):
LGTM, this is very useful for the password-based authentication as well, thanks Adam!

PR #20, opened 2026-02-16
Remove individual dry-run & package level vars

logan-scully-arm:
## Changes

- Removes individual dry-run initialization, in favor of an `addDryRun` function
- Replaces Package level  dry-run vars with GetBool during command execution.


cmd/topo/deploy.go
```diff
@@ -44,6 +43,12 @@ Use --dry-run to see what commands would be executed without actually running th
 	Args: cobra.ExactArgs(0),
 	RunE: func(cmd *cobra.Command, args []string) error {
 		cmd.SilenceUsage = true
+
+		dryRun, err := cmd.Flags().GetBool("dry-run")
+		if err != nil {
+			panic(fmt.Sprintf("bug: dry-run flag not registered: %v", err))
```
federicobozzini: Is there a need to `panic` here? Can we not just return `err`?

logan-scully-arm: This was to line up with what @th3james suggested in #16. 

It might be better to do this as a third return value...? LMK what you think in that PR :pray:

th3james: I remain of the opinion that we should distinguish between errors that we believe could occur in a real system, which we should propagate for handling, vs errors which only occur if we've made a programming mistake, which the program should not try to handle because they should never happen. By propagating an error that should never happen, we risk callers wasting time and code handling those cases.

federicobozzini: I see the logic now. It makes sense to me to `panic`, but maybe then it should be an utility function `requireDryRun` that returns the value and panic if fails.

logan-scully-arm: what would be the use in having another function for it? 

Also I think `requireDryRun` implies that you have to use the flag.

federicobozzini: > what would be the use in having another function for it?

If the logic is exactly the same, to me it makes sense not to repeat the code multiple times.

> Also I think requireDryRun implies that you have to use the flag.

We can use a different name maybe?

BTW I've also notice `panic(fmt.Sprintf("bug: dry-run flag not registered: %v", err))`. I would argue that `internal error` is more appropriate than `bug`.

yejseo01 (approved):
LGTM

PR #22, opened 2026-02-16
Updated template urls

federicobozzini:
## Changes

- Updated template URLs to the new names under the `Arm-Examples` org.

### DIFF
--- a/scripts/update_templates/main.go
+++ b/scripts/update_templates/main.go
@@ -10,7 +10,6 @@ const outputJSONPath = "internal/catalog/data/templates.json"
 
 var repoList = []string{
 	"Arm-Examples/topo-template-welcome#main",
-	"Arm-Debug/topo-kleidi-service#main",
 	"Arm-Examples/topo-template-lightbulb-moment#main",
 	"Arm-Examples/topo-template-v9-cpu-chat#main",
 	"Arm-Examples/topo-template-simd-visual-benchmark#main",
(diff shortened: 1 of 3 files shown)


logan-scully-arm (changes requested):
This should be updated in `update_templates` not the JSON. 

Running update templates will then automatically read in the x-topo data.

federicobozzini:
Sorry I was not aware of this change. Done, but I would argue `template.json` should not be committed then.

logan-scully-arm:
I think this was a solution to them not being public. When the templates go public, I don't think you'll even need to create a JSON (not 100% on this though). The generation was just so that everyone didn't have to set up the specific token etc.

scripts/update_templates/main.go
```diff
@@ -9,11 +9,11 @@ import (
 const outputJSONPath = "internal/catalog/data/templates.json"
 
 var repoList = []string{
-	"Arm-Debug/topo-cortexa-welcome#main",
+	"Arm-Examples/topo-template-welcome#main",
 	"Arm-Debug/topo-kleidi-service#main",
```
awphi: Can be binned I think?

federicobozzini: Aaaand it's gone!

awphi (approved):
🚀

logan-scully-arm:
~ ~I'll approve, but I think you've probably got a Kleidi test still to remove~ ~

Ignore me, It just hadn't started the runs yet :laughing:

logan-scully-arm (approved):
Hazzah 🚀

PR #23, opened 2026-02-17
Varless flags in deploy command

logan-scully-arm:
## Changes

- Removes the package level args to replace them with GetString and getBool.
- These result in panics whenever there is an issue -- this is highly unlikely so it's clarified to be an internal error here.


tgonzalezorlandoarm (changes requested):
I realize that I haven't put the request changes in the comment before

yejseo01 (commented):
Why is being varless better?

logan-scully-arm:
@yejseo01  When we bind flags to vars, they end up being shared across all commands, which usually leads to a bunch of variables at the top like xyzDryRun. James and I found that using GetXxx() is much cleaner—it avoids that shared mutable state and keeps the flag values scoped to the command that actually uses them.

tgonzalezorlandoarm:
I agree with @yejseo01 :

I don't think this is improving the codebase, just making it (slightly) more complicated.
Theoretically we are tracking less state, but in practice the vars approach was a lot more direct and was very simple to get right, which is what really matters (hence why we were correctly using them all over the codebase).

Also, it means that we don't have to manage errors where supposedly the variable is not being set.

logan-scully-arm:
@tgonzalezorlandoarm you know what, you're probably right. 

I think the flags that are carried between functions are all handled now: `dry-run`, `target`, etc. That's where this approach probably saw the most benefit. 

I'll close the remainders and carry on w/ the more important parts.

PR #24, opened 2026-02-18
chore: refactor ssh executions and unify them under Connections

tgonzalezorlandoarm:
Currently, there are a lot of different APIs to execute command through SSH both on localhost and on target.
This PR aims to unify all of these under a single API on the `Connection` structure (`Run` as a public method, `Exec` as the private method that implements `Run`).

## Changes

- Unify all SSH executions into the Connection framework


internal/target/connection.go
```diff
@@ -56,14 +58,26 @@ func NewConnection(sshTarget string, exec execSSH, opts ConnectionOptions) Conne
 }
 
 func (c *Connection) Run(command string) (string, error) {
-	return c.exec(c.SSHTarget, command)
+	if c.opts.WithLoginShell {
+		command = ssh.ShellCommand(command)
+	}
+
+	stdout, stderr, err := c.exec(c.SSHTarget, command, c.opts.WithStdin)
+	if err != nil {
```
tgonzalezorlandoarm, line 66: I'm unsure about this part.
1. Should I just merge stdout + stderr?
2. Should I remove this weird if condition?

internal/install/install.go
```diff
@@ -263,7 +267,8 @@ func install(installPath string, targetHost ssh.Host, binaries map[string][]byte
 		}
 
 		installCmd := fmt.Sprintf("install -D -m %s /dev/stdin %s/%s", mode, installPath, binaryName)
-		_, stderr, err := ssh.Exec(targetHost, installCmd, binaryData)
+		conn := target.NewConnection(string(targetHost), ssh.Exec, target.ConnectionOptions{WithLoginShell: true, WithStdin: binaryData})
```
tgonzalezorlandoarm, line 270: From what I understand of Go, this is not copying that data but rather just copying the pointer

internal/target/connection.go
tgonzalezorlandoarm: Decided to do all of hte above

internal/target/connection.go
```diff
@@ -56,14 +58,26 @@ func NewConnection(sshTarget string, exec execSSH, opts ConnectionOptions) Conne
...
 func (c *Connection) BinaryExists(bin string) (bool, error) {
 	if err := ssh.ValidateBinaryName(bin); err != nil {
 		return false, err
 	}
-	_, err := c.exec(c.SSHTarget, ssh.ShellCommand(fmt.Sprintf("command -v %s", bin)))
+
+	prevLoginShell := c.opts.WithLoginShell
+	c.opts.WithLoginShell = true
+	_, err := c.Run(fmt.Sprintf("command -v %s", bin))
+	c.opts.WithLoginShell = prevLoginShell
```
logan-scully-arm: Is there a reason this needs to be a login shell? 

Secondary to that, is the `WithLoginShell` option necessary for anything? I'm not sure that run needs it, and this function overwrites it?

internal/target/connection_auth_test.go
```diff
@@ -1,151 +1,174 @@
...
+	command string
+	args    []string
+}
+
+type mockResponse struct {
+	stdout string
+	err    error
+}
+
+func newMockExec(responses map[string]mockResponse, calls *[]sshTestCall) func(ssh.Host, string, []byte, ...string) (string, error) {
```
logan-scully-arm, line 25: I'm fine w/ this, but I think others might prefer you to write simpler versions of this that are specific to each test, rather than a universal one. 

Wait for another opinion before refactoring though, bc I'd say this is probably alright 👍

logan-scully-arm:
Great work, just two nitpicks/questions. Happy to approve after those have replies :+1:

internal/target/connection_auth_test.go
awphi: Agree this is okay - neither option is perfect imo. This new mocking approach means each test has the flexibility to use this god mock or a test-specific one!

internal/target/connection_auth_test.go
```diff
@@ -1,151 +1,174 @@
...
+		if !ok {
+			return "", fmt.Errorf("unexpected ssh mode: %s", mode)
+		}
+		return resp.stdout, resp.err
 	}
-	return target.NewConnection("user@host", mockExec, opts)
 }
 
 func TestProbeAuthentication(t *testing.T) {
-	testutil.RequireOS(t, "linux")
```
awphi, line 21: Good to see things just work on windows!

awphi (approved):
Nice! I wasn't involved much in the original discussion about the ssh mock so would want to give the final vote to @logan-scully-arm but this looks like a nice refactor to me 🚀

internal/ssh/exec.go
```diff
@@ -70,15 +53,9 @@ func Exec(target Host, command string, stdin []byte, sshArgs ...string) (stdout,
...
-	return stdoutBuf.String(), stderrBuf.String(), err
-}
-
-// ExecWithShell runs a command in a login shell on the target host. If the target is localhost, it runs locally.
-func ExecWithShell(target Host, command string) (string, error) {
-	stdout, stderr, err := Exec(target, ShellCommand(command), nil)
+	err := cmd.Run()
 	if err != nil {
-		return "", fmt.Errorf("command failed: %w | stderr: %s", err, stderr)
+		return "", fmt.Errorf("ssh command to %s failed: %w | stderr: %s", string(target), err, stderrBuf.String())
```
federicobozzini: Here the target might come from a env variable and we migthe be exposing the connection password.

logan-scully-arm (approved):
My queries are non-blocking so approving now.

Good work 👍

internal/ssh/exec.go
federicobozzini: It's not the password, just the connection. Ignore this :facepalm:

tgonzalezorlandoarm: Having discussed offline, we are exposing the user here, which is not considered sensitive to the person that is calling `topo`.
Resolving

internal/target/connection.go
tgonzalezorlandoarm: > Is there a reason this needs to be a login shell?

I think you are right and this does not need to override `WithLoginShell` in this case.

> is the WithLoginShell option necessary for anything?

`WithLoginShell` is necessary to access variables like `$HOME` in functions like `getHomeDir` (mostly the `install` functions)

tgonzalezorlandoarm: https://github.com/arm/topo/compare/7e1146dd67a2060000d629c6e7dd8e061a2aad22..8935bb84b666b8c6ea297f063ebdcce54de06210

PR #25, opened 2026-02-18
Use same variable for the github token

federicobozzini:
## Changes

- I noticed we use `GH_PATH` for the update script but in our codebases (including this one) we normally use `GITHUB_TOKEN`. This PR makes things more consistent.

### DIFF
--- a/scripts/update_templates/main.go
+++ b/scripts/update_templates/main.go
@@ -24,9 +24,9 @@ type Template struct {
 }
 
 func main() {
-	token := os.Getenv("GH_PAT")
+	token := os.Getenv("GITHUB_TOKEN")
 	if token == "" {
-		fmt.Fprintln(os.Stderr, "GH_PAT is not set: create a personal access token and set the envvar")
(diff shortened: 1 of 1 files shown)


tgonzalezorlandoarm (commented):
How can we verify that this works with the current setup?

tgonzalezorlandoarm:
> How can we verify that this works with the current setup?

Can we trigger a CI run to show that this works and link it here?

yejseo01 (commented):
Seems like a simple change, but I think @tgonzalezorlandoarm should approve it

federicobozzini:
> How can we verify that this works with the current setup?

You can just run `unset GH_PAT; go run ./scripts/update_templates` and see that nothing breaks


> Can we trigger a CI run to show that this works and link it here?

I'm afraid not. This functionality is not (yet) used by the CI. We should but it might be a bit out of scope here.

tgonzalezorlandoarm (approved):
I understand this change better now. Sorry for the noise!

PR #26, opened 2026-02-18
Protect against ssh flag-injection

logan-scully-arm:
## Changes

- SSH already contains the right functionality to protect against command injection.
- With the use of `--` injected options are rejected.

### DIFF
--- a/internal/ssh/exec.go
+++ b/internal/ssh/exec.go
@@ -42,7 +42,7 @@ func Exec(target Host, command string, stdin []byte, sshArgs ...string) (string,
 	if target.IsPlainLocalhost() {
 		cmd = exec.Command("/bin/sh", "-c", command)
 	} else {
-		args := slices.Concat(sshArgs, []string{string(target), command})
+		args := slices.Concat(sshArgs, []string{"--", string(target), command})
 		cmd = exec.Command("ssh", args...)
 	}
 


cmd/topo/root.go
```diff
@@ -80,3 +89,47 @@ func resolveOutput(cmd *cobra.Command) (term.Format, error) {
...
+	if s == "" {
+		return false
+	}
+	for _, r := range s {
+		if !unicode.IsLetter(r) && !unicode.IsDigit(r) && r != '-' && r != '_' {
+			return false
+		}
+	}
+	return true
+}
```
awphi: I think this breaks for simple raw ip hosts because of the dots `x.y.z.w`

awphi: I think this will throw up a bunch of false positives actually - most punctuation is valid in `Host` aliases

awphi (changes requested):
I'm a bit concerned about our validation being overzealous here and bricking topo for a lot of perfectly valid host names.

I'm not even sure we're vulnerable to command injection atm given we're using `exec.Command` which invokes ssh/docker/whatever directly instead of invoking the shell - with the exception of where we use login shells in `topo install` which may not be necessary.

If I'm wrong and we do need it could we explore escaping as opposed to validation?

awphi:
Actually could we perform the validation by running `exec.Command("ssh", "-G", target)` and checking for a non-zero exit code since this seems to perform ssh's built-in host validation

```
❯ ssh -G "foo bar"
hostname contains invalid characters
```

We'd still need to split port + validate that separately but that's a more trivial task

cmd/topo/root.go
logan-scully-arm: Yeah I just realized this :skull:

logan-scully-arm:
> Actually could we perform the validation by running `exec.Command("ssh", "-G", target)` and checking for a non-zero exit code since this seems to perform ssh's built-in host validation
> 
> ```
> ❯ ssh -G "foo bar"
> hostname contains invalid characters
> ```
> 
> We'd still need to split port + validate that separately but that's a more trivial task

Good shout

logan-scully-arm:
@awphi Thank you for making me fix this :pray: 

SSH will check all of this by itself, we can actually use the already existing function `ProbeConnection` which will return an error if there are invalid characters.

tgonzalezorlandoarm (changes requested):
I don't think this is the approach we want to take:
1. If we'd have to run `ProbeConnection` every time we create a new connection, then we should just add it inside the `NewConnection` function
2. This checks that we can `ssh` into the target, that it's something that `health` already does when it reports on `connected`.
3. This has performance implications, it's running an extra `ssh` connection (that IMO is unneded).

The more general question is:
What are we trying to validate and why?

IMO, what we should be trying to validate is not whether we can connect to the target (as mentioned, `health` already does this) but rather that the target is that, a valid target, and just that:
No prompt injection, no malicious options, etc.

A little bit of this is covered by how we trigger the `Command("ssh",...)`

I think one approach for this may be adding the `--` flag before the target when we invoke the ssh command, but it probably doesn't cover everything we should be looking for.

logan-scully-arm:
@tgonzalezorlandoarm I completely forgot about adding `--`, I can def do that

The reason I used probe is that ssh already performs the necessary validation internally. By probing it, we effectively verify that the input is valid. My thinking was that since a successful connection is ultimately expected behavior, validating the actual connection felt reasonable as part of the check.

That said, I could have used -G to validate the configuration without connecting; but i felt that You'd want to check that you can connect anyway

tgonzalezorlandoarm:
> @tgonzalezorlandoarm I completely forgot about adding `--`, I can def do that
> 
> The reason I used probe is that ssh already performs the necessary validation internally. By probing it, we effectively verify that the input is valid. My thinking was that since a successful connection is ultimately expected behavior, validating the actual connection felt reasonable as part of the check.
> 
> That said, I could have used -G to validate the configuration without connecting; but i felt that You'd want to check that you can connect anyway

Perhaps I'm missing something, but if we are already running `ssh` commands in our codebase's logic and `ssh` is the way to verify, why do we inject this extra command to probe? What's the added benefit against just erroring out on the codebase's logic ssh commands?

logan-scully-arm:
@tgonzalezorlandoarm Yeah I can actually just remove this check :laughing: 

I'll keep `--`, but I think we should be okay for everything else then.

Thank you :smiling_face_with_tear:

awphi (approved):
I've been playing around with this more this morning trying and so far been unable to perform any command injection via `--target` apart from the flag injection approach @tgonzalezorlandoarm pointed out (e.g. `./topo health --target '-oProxyCommand=sh -c "id > /tmp/pwn" #'`) which you've now mitigated against so LGTM 🚀

tgonzalezorlandoarm (approved):
LGTM, thanks Logan! 🚀

PR #27, opened 2026-02-18
Filter templates now handles errors

logan-scully-arm:
## Changes

- I noticed that the template filtering on target failed silently if you could not connect to the target
- Now fails when a target is unavailable.

### DIFF
--- a/cmd/topo/templates.go
+++ b/cmd/topo/templates.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"fmt"
 	"os"
 
 	"github.com/arm-debug/topo-cli/internal/catalog"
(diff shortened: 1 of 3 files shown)


cmd/topo/templates.go
```diff
@@ -31,7 +32,10 @@ var templatesCmd = &cobra.Command{
 			return err
 		}
 
-		repos = catalog.FilterTemplateRepos(templateFilters, repos)
+		repos, err = catalog.FilterTemplateRepos(templateFilters, repos)
+		if err != nil {
+			return fmt.Errorf("could not filter templates: %w", err)
```
federicobozzini, line 37:
```suggestion
			return fmt.Errorf("an error happened while filtering templates: %w", err)
```

Maybe?

federicobozzini (approved):

cmd/topo/templates.go
logan-scully-arm: The logger at the end will output `ERROR: could not filter templates: ...`.
I don't agree with your suggestion, but happy if there's a better option.

federicobozzini: Sounds reasonable then. I'd keep your message :+1:

PR #28, opened 2026-02-18
fix: ubuntu CI failures

awphi:
## Changes

- Removes unneeded CI testutil code that was causing failures due to some newly discovered race condition.


tgonzalezorlandoarm (approved):
nit: newly introduced -> newly discovered

PR #30, opened 2026-02-19
Removed references to topo-cli

federicobozzini:
## Changes

- We don't want to reference `topo-cli`. Project is just `topo`.


logan-scully-arm (approved):
Good spot 👍

tgonzalezorlandoarm (approved):

PR #31, opened 2026-02-19
Templates update

federicobozzini:
## Changes

- Periodid templates update


logan-scully-arm (approved):
I wonder if we could set this up to run itself periodically so that we don't have to?

federicobozzini:
> I wonder if we could set this up to run itself periodically so that we don't have to?

We definitely should. I also don't like that our tests are tied to the values in the templates file. One step at a time though..

PR #34, opened 2026-02-19
feat: add Windows Public Key transfer support

tgonzalezorlandoarm:
On Windows, the ssh-copy-id tool is not available.

 * use ssh commands (through the Connection interface) to perform the transfer to the target instead.
 * remove Linux requirements for the related tests.
 * add a new way to mock tests.

### DIFF
--- a/cmd/topo/setup_keys.go
+++ b/cmd/topo/setup_keys.go
@@ -4,7 +4,7 @@ import (
 	"fmt"
 	"os"
 
-	"github.com/arm/topo/internal/setupkeys"
+	"github.com/arm/topo/internal/setup_keys"
 	"github.com/spf13/cobra"
 )
 
@@ -29,12 +29,13 @@ Use --dry-run to see what commands would be executed without actually running th
 		if err != nil {
 			return err
 		}
-		privKeyPath, err := setupkeys.CreateKeyPair(resolvedTarget, setupKeysKeyPath, os.Stdout, dryRun)
+		targetFileName := setup_keys.SanitizeTarget(resolvedTarget)
+		privKeyPath, err := setup_keys.CreateKeyPair(resolvedTarget, targetFileName, setupKeysKeyPath, os.Stdout, dryRun)
 		if err != nil {
 			return err
 		}
 
-		return setupkeys.TransferPubKey(resolvedTarget, privKeyPath, os.Stdout, dryRun)
+		return setup_keys.TransferPubKey(resolvedTarget, privKeyPath, os.Stdout, dryRun)
 	},
 }
 
--- a/cmd/topo/setup_keys.go
+++ b/cmd/topo/setup_keys.go
@@ -29,20 +29,12 @@ Use --dry-run to see what commands would be executed without actually running th
 		if err != nil {
 			return err
 		}
-		keyPairGenCmd, privKeyPath, err := setupkeys.NewKeyPairCreation(resolvedTarget, setupKeysKeyPath)
+		privKeyPath, err := setupkeys.CreateKeyPair(resolvedTarget, setupKeysKeyPath, os.Stdout, dryRun)
 		if err != nil {
-			return err
-		}
-		if err := setupkeys.RunKeyPairCreation(keyPairGenCmd, os.Stdout, dryRun); err != nil {
 			return err
 		}
 
-		pubKeyTransfer, err := setupkeys.NewPubKeyTransfer(resolvedTarget, privKeyPath, dryRun)
-		if err != nil {
-			return err
-		}
-		err = setupkeys.RunPubKeyTransfer(pubKeyTransfer, privKeyPath, os.Stdout, dryRun)
-		return err
+		return setupkeys.TransferPubKey(resolvedTarget, privKeyPath, os.Stdout, dryRun)
 	},
 }
 
--- a/cmd/topo/setup_keys.go
+++ b/cmd/topo/setup_keys.go
@@ -3,7 +3,6 @@ package main
 import (
 	"fmt"
 	"os"
-	"runtime"
 
 	"github.com/arm/topo/internal/setupkeys"
 	"github.com/spf13/cobra"
@@ -24,10 +23,6 @@ Use --dry-run to see what commands would be executed without actually running th
 		dryRun, err := cmd.Flags().GetBool("dry-run")
 		if err != nil {
 			panic(fmt.Sprintf("internal error: dry-run flag not registered: %v", err))
-		}
-
-		if runtime.GOOS != "linux" {
-			return fmt.Errorf("topo setup-keys currently supports Linux hosts only")
 		}
 
 		resolvedTarget, err := requireTarget(cmd)
--- a/cmd/topo/setup_keys.go
+++ b/cmd/topo/setup_keys.go
@@ -34,16 +34,20 @@ Use --dry-run to see what commands would be executed without actually running th
 		if err != nil {
 			return err
 		}
-
-		seq, err := setupkeys.NewKeyCreationAndPlacementOnTarget(resolvedTarget, setupKeysKeyPath)
+		keyPairGenCmd, privKeyPath, err := setupkeys.NewKeyPairCreation(resolvedTarget, setupKeysKeyPath)
 		if err != nil {
 			return err
 		}
-
-		if dryRun {
-			return seq.DryRun(os.Stdout)
+		if err := setupkeys.RunKeyPairCreation(keyPairGenCmd, os.Stdout, dryRun); err != nil {
+			return err
 		}
-		return seq.Run(os.Stdout)
+
+		pubKeyTransfer, err := setupkeys.NewPubKeyTransfer(resolvedTarget, privKeyPath, dryRun)
+		if err != nil {
+			return err
+		}
+		err = setupkeys.RunPubKeyTransfer(pubKeyTransfer, privKeyPath, os.Stdout, dryRun)
+		return err
 	},
 }
 
--- a/internal/setup_keys/setup_keys_test.go
+++ b/internal/setup_keys/setup_keys_test.go
@@ -1,4 +1,4 @@
-package setupkeys
+package setup_keys_test
 
 import (
 	"bytes"
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/arm/topo/internal/setup_keys"
 	"github.com/arm/topo/internal/ssh"
 	"github.com/stretchr/testify/require"
 )
@@ -51,7 +52,8 @@ func TestNewKeyCreationAndPlacementOnTargetDryRun(t *testing.T) {
 			}
 
 			var buf bytes.Buffer
-			keyPath, err := CreateKeyPair("user@example.com", inputKeyPath, &buf, true)
+			targetFileName := setup_keys.SanitizeTarget("user@example.com")
+			keyPath, err := setup_keys.CreateKeyPair("user@example.com", targetFileName, inputKeyPath, &buf, true)
 			require.NoError(t, err)
 
 			wantKeyPath := filepath.Join(tmp, tt.wantKeyPath)
@@ -62,7 +64,7 @@ func TestNewKeyCreationAndPlacementOnTargetDryRun(t *testing.T) {
 			require.Contains(t, got, "ssh-keygen", "DryRun output should include keygen command")
 			require.Contains(t, got, wantKeygen, "DryRun output should include keygen arguments")
 
-			transferErr := TransferPubKey("user@example.com", keyPath, &buf, true)
+			transferErr := setup_keys.TransferPubKey("user@example.com", keyPath, &buf, true)
 			require.NoError(t, transferErr)
 			got = buf.String()
 			require.Contains(t, got, "ssh user@example.com", "DryRun output should include ssh command")
@@ -76,14 +78,15 @@ func TestKeyCreationAndPlacementOnTarget(t *testing.T) {
 	logFile := filepath.Join(t.TempDir(), "commands.log")
 
 	testLogFile := logFile
-	origExec := execCommand
-	execCommand = func(command string, args ...string) *exec.Cmd {
+	origExec := setup_keys.ExecCommand
+	setup_keys.ExecCommand = func(command string, args ...string) *exec.Cmd {
 		return mockExecCommand(testLogFile, command, args...)
 	}
-	t.Cleanup(func() { execCommand = origExec })
+	t.Cleanup(func() { setup_keys.ExecCommand = origExec })
 
 	var buf bytes.Buffer
-	keyPath, keyPairErr := CreateKeyPair("user@example.com", keyPath, &buf, false)
+	targetFileName := setup_keys.SanitizeTarget("user@example.com")
+	keyPath, keyPairErr := setup_keys.CreateKeyPair("user@example.com", targetFileName, keyPath, &buf, false)
 	require.NoError(t, keyPairErr)
 	require.Contains(t, buf.String(), "ssh-keygen invoked", "Run output should include fake ssh-keygen output")
 
@@ -92,21 +95,22 @@ func TestKeyCreationAndPlacementOnTarget(t *testing.T) {
 	log := string(logData)
 	require.Contains(t, log, fmt.Sprintf("ssh-keygen -t ed25519 -f %s -C user@example.com", keyPath))
 
-	origSSHExec := sshExec
+	origSSHExec := setup_keys.SSHExec
 	var gotCommand string
 	var gotStdin []byte
-	sshExec = func(_ ssh.Host, command string, stdin []byte, _ ...string) (string, error) {
+	setup_keys.SSHExec = func(_ ssh.Host, command string, stdin []byte, _ ...string) (string, error) {
 		gotCommand = command
 		gotStdin = stdin
 		return "", nil
 	}
-	t.Cleanup(func() { sshExec = origSSHExec })
+	t.Cleanup(func() { setup_keys.SSHExec = origSSHExec })
 
-	transferErr := TransferPubKey("user@example.com", keyPath, &buf, false)
+	transferErr := setup_keys.TransferPubKey("user@example.com", keyPath, &buf, false)
 	require.NoError(t, transferErr)
 	pubKey, err := os.ReadFile(keyPath + ".pub")
 	require.NoError(t, err)
-	require.Equal(t, ssh.ShellCommand(remoteAuthorizedKeysCommand), gotCommand)
+	wantCommand := ssh.ShellCommand("mkdir -p ~/.ssh && chmod 700 ~/.ssh && cat >> ~/.ssh/authorized_keys && chmod 600 ~/.ssh/authorized_keys")
+	require.Equal(t, wantCommand, gotCommand)
 	require.Equal(t, pubKey, gotStdin)
 }
 
@@ -122,7 +126,7 @@ func TestSanitizeTarget(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.input, func(t *testing.T) {
-			if got := sanitizeTarget(tt.input); got != tt.want {
+			if got := setup_keys.SanitizeTarget(tt.input); got != tt.want {
 				t.Fatalf("sanitizeTarget(%q) = %q, want %q", tt.input, got, tt.want)
 			}
 		})
--- a/internal/setup_keys/setup_keys_test.go
+++ b/internal/setup_keys/setup_keys_test.go
@@ -2,21 +2,16 @@ package setup_keys_test
 
 import (
 	"bytes"
-	"fmt"
-	"os"
-	"os/exec"
 	"path/filepath"
 	"runtime"
-	"slices"
 	"strings"
 	"testing"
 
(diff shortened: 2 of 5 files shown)


cmd/topo/setup_keys.go
```diff
@@ -34,16 +34,20 @@ Use --dry-run to see what commands would be executed without actually running th
 		if err != nil {
 			return err
 		}
-
-		seq, err := setupkeys.NewKeyCreationAndPlacementOnTarget(resolvedTarget, setupKeysKeyPath)
+		keyPairGenCmd, privKeyPath, err := setupkeys.NewKeyPairCreation(resolvedTarget, setupKeysKeyPath)
 		if err != nil {
 			return err
 		}
+		if err := setupkeys.RunKeyPairCreation(keyPairGenCmd, os.Stdout, dryRun); err != nil {
```
logan-scully-arm: I'm not too sure about this pattern, IMO it might be nicer to call the command creation in `RunKeyPairCreation`

`CreateKeyPair` might be a clearer name?

cmd/topo/setup_keys.go
logan-scully-arm: This isn't true anymore right?

cmd/topo/setup_keys.go
tgonzalezorlandoarm: That makes sense to me, instead of having a `New` + `Run` cycle I'll just have a `Run` command, for both key creating and for transferring to target

cmd/topo/setup_keys.go
tgonzalezorlandoarm: Good spot!
Done: https://github.com/arm/topo/pull/34/commits/98b8e85bcc26241aec14b15bafd42bc1a3cd2575
Which brings the point that I haven't tested this on an actual Windows box, just ran the tests on CI for Windows.
Will do that before merging, thanks for spotting!

cmd/topo/setup_keys.go
logan-scully-arm: I think there's lacking a bit of clarity because of the two `New` functions. One returns a command while the other returns a connection? Maybe I'm overthinking it, but I wouldn't expect it

logan-scully-arm: Sorry I wrote the before refreshing ^

tgonzalezorlandoarm: I guess the main things about this pattern though is that I can easily hook a test this way, will see how to adapt

internal/setup_keys/setup_keys_test.go
```diff
@@ -4,77 +4,119 @@ import (
...
+		{
+			name:         "custom key path",
+			inputKeyPath: filepath.Join("custom_keys", "id_ed25519_custom"),
+			wantKeyPath:  filepath.Join("custom_keys", "id_ed25519_custom"),
+		},
+	}
 
-		seq, err := NewKeyCreationAndPlacementOnTarget("user@example.com", "")
-		require.NoError(t, err)
+	for _, tt := range tests {
```
logan-scully-arm, line 32: I'm not entirely sold on the loop here, I'd probably prefer these as two separate tests, even if there is repeated code -- albeit this isn't a hill I'd die on 😆

internal/setup_keys/setup_keys_test.go
```diff
@@ -4,77 +4,119 @@ import (
...
 	"github.com/stretchr/testify/require"
 )
 
-func TestNewKeyCreationAndPlacementOnTarget(t *testing.T) {
-	testutil.RequireOS(t, "linux")
-
-	t.Run("default key path", func(t *testing.T) {
-		tmp := t.TempDir()
-		t.Setenv("HOME", tmp)
+func TestNewKeyCreationAndPlacementOnTargetDryRun(t *testing.T) {
```
logan-scully-arm: Because you've split the functions apart, these tests should be changed to reflect that. 
`TestCreateKeyPair`
`TestTransferPublicKey`

internal/setup_keys/setup_keys_test.go
tgonzalezorlandoarm: Originally I had them separated, but while editting the tests I had to duplicate efforts for editting both and for keeping them consistent. This is basically a result of going through the pain of the other approach while developing, so I prefer to keep it as is if it's ok with you

internal/setup_keys/setup_keys_test.go
logan-scully-arm: I'd also veer away from dry-run functions in this. 

The `TestCreateKeyPair` could do `does x when default key path in dry run`?

internal/setup_keys/setup_keys_test.go
logan-scully-arm: Yeah all good with me 👍

cmd/topo/setup_keys.go
logan-scully-arm: Just an aside that isn't really a point of this PR, but I just noticed it. Your file `setupkeys.go` should be `setup_keys.go`. And is there any reason that this isn't a part of the SSH package?

cmd/topo/setup_keys.go
logan-scully-arm: If the tests were passing it might be a good point to add some e2e tests for this

internal/setup_keys/setup_keys_test.go
```diff
@@ -4,77 +4,119 @@ import (
...
+			require.NoError(t, err)
+			require.NoError(t, RunPubKeyTransfer(conn, keyPath, &buf, true))
+			got = buf.String()
+			require.Contains(t, got, "ssh user@example.com", "DryRun output should include ssh command")
+			require.Contains(t, got, keyPath+".pub", "DryRun output should include public key path")
+		})
+	}
 }
 
 func TestKeyCreationAndPlacementOnTarget(t *testing.T) {
```
logan-scully-arm: I think these should be updated to t.Run() formatting

internal/setupkeys/setupkeys.go
```diff
@@ -9,29 +9,77 @@ import (
...
-		keyName := fmt.Sprintf("id_ed25519_topo_%s", sanitizeTarget(target))
-		keyPath = filepath.Join(home, ".ssh", keyName)
+		keyName := fmt.Sprintf("id_ed25519_topo_%s", sanitizeTarget(targetHost))
+		privKeyPath = filepath.Join(home, ".ssh", keyName)
 	}
 
-	if err := ensureDir(keyPath); err != nil {
-		return nil, err
+	if err := ensureDir(privKeyPath); err != nil {
+		return nil, "", err
```
awphi: This gets created even on the dry-run path which while not directly harmful isn't really the spirit of a dry run

internal/setupkeys/setupkeys.go
```diff
@@ -9,29 +9,77 @@ import (
...
 	}
 
-	ops := []goperation.Operation{
-		newSetupKeysOperation("Generate SSH key pair for target", []string{"ssh-keygen", "-t", "ed25519", "-f", keyPath, "-C", target}),
-		newSetupKeysOperation("Copy SSH public key to target", []string{"ssh-copy-id", "-i", keyPath + ".pub", target}),
+	return execCommand("ssh-keygen", "-t", "ed25519", "-f", privKeyPath, "-C", targetHost), privKeyPath, nil
```
awphi: I'm not a fan of moving away from the sequence/operation abstraction we use elsewhere in the codebase + here previously if for nothing else but consistency. What was the reason for moving away from it?

internal/setup_keys/setup_keys_test.go
```diff
@@ -96,12 +138,64 @@ func TestSanitizeTarget(t *testing.T) {
...
+	cs := slices.Concat([]string{"-test.run=TestHelperProcess", "--", command}, args)
+	cmd := exec.Command(os.Args[0], cs...)
+	cmd.Env = append(os.Environ(),
+		"GO_WANT_HELPER_PROCESS=1",
+		"TOPO_TEST_LOG="+logFile,
+	)
+	return cmd
+}
+
+func TestHelperProcess(t *testing.T) {
```
awphi: Why can't we just use the real commands i.e. `ssh-keygen` in this case in the tests?

internal/setupkeys/setupkeys.go
```diff
@@ -9,29 +9,77 @@ import (
...
-	goperation "github.com/arm/topo/internal/deploy/operation"
+	"github.com/arm/topo/internal/ssh"
+	"github.com/arm/topo/internal/target"
 )
 
-func NewKeyCreationAndPlacementOnTarget(target string, keyPath string) (goperation.Sequence, error) {
-	if keyPath == "" {
+const remoteAuthorizedKeysCommand = "mkdir -p ~/.ssh && chmod 700 ~/.ssh && cat >> ~/.ssh/authorized_keys && chmod 600 ~/.ssh/authorized_keys"
+
+var execCommand = exec.Command
```
awphi: Similar to the other comment - not a fan of global state that is manipulated like this if it can be avoided. Imo we should pass it in the function consuming it on call site or skip it entirely if we don't need to use different exec.Command impls

awphi (commented):
Few bits I'd like to understand the reasoning behind a bit more

internal/setupkeys/setupkeys_test.go
awphi: Also noticed this should be in its own `setupkeys_test` package for proper blackbox testing in line with our other `_test` files we should only test what is exported which is mostly true of this file atm anyway.

cmd/topo/setup_keys.go
tgonzalezorlandoarm: I've now done the unification here:
https://github.com/arm/topo/pull/34/commits/78b384f1eaf2987f4c033af5f7d6d07fafcd3a5a

internal/setupkeys/setupkeys.go
logan-scully-arm: I agree with this ^

cmd/topo/setup_keys.go
tgonzalezorlandoarm: > is there any reason that this isn't a part of the SSH package?

This uses `ssh` commands to provide other functionality. It uses the ssh functionality to create keys. I see what you mean though, but it's a separate command on its own.... I'm undecided. Let's see what @awphi thinks?

tgonzalezorlandoarm: > Your file setupkeys.go should be setup_keys.go

Why should it be like this?

internal/setupkeys/setupkeys.go
tgonzalezorlandoarm: Good point, this should be fixed as of: https://github.com/arm/topo/pull/34/commits/78b384f1eaf2987f4c033af5f7d6d07fafcd3a5a

internal/setupkeys/setupkeys.go
tgonzalezorlandoarm: mm, looking back at https://github.com/arm/topo/pull/14#discussion_r2792841866
you were saying that we should have one abstraction for the ssh key gen operation and one for the ssh-copy-id/equivalent ,  right?

My thinking was that this was an overkill of abstractions for something that is not repeatable.
Happy to update, but what do you think about this?

internal/setup_keys/setup_keys_test.go
tgonzalezorlandoarm: I don't feel comfortable creating actual keys for the user even if they are temporary ones

internal/setupkeys/setupkeys.go
tgonzalezorlandoarm: I prefer having this type of global state because, IMO, it keeps testing and actual implementation logic separate.

Part of the reason that we were deviating so much with the ExecSSH functions to the point it needed some refactoring in #24  is because we are currently injecting that `ExecSSH` into the Connection, when in fact it could just be an internal method.
We are only injecting the function so as to mock test, so testing logic is affecting production code, which in turn ir made us produce fragmentation in our code.
In fact, the testing logic is affecting so much the production code on that regard, that we cannot easily implement a `Connection.DryRun()` because it's very hard to do so with `ExecSSH` function injection

tgonzalezorlandoarm: Conditionally compiling testing logic only on tests sounds like a good separation to me, it keeps code clean (avoiding the above issues) while also enforcing a barrier between what goes into production and what is development-related logic

cmd/topo/setup_keys.go
logan-scully-arm: That is the naming convention of the files in the project. `static_provider.go`, `cli_provider.go`

internal/setupkeys/setupkeys.go
awphi: > you were saying that we should have one abstraction for the ssh key gen operation and one for the ssh-copy-id/equivalent , right?

Yes exactly, sorry if I wasn't clear. I think this is good enough but it's a shame the output formatting of deploy and setup-keys are different - i.e. with deploy you get the nice named headers from `Sequence.Run/DryRun`. 

Using that abstraction also means we could use conditional operations to fail more gracefully when running `setup-keys` on the same target multiple times in a row:

Screenshot 2026-02-20 at 15 06 19

Will leave the final decision to you!

internal/setup_keys/setup_keys_test.go
awphi: Ahh gotcha - I was only thinking about CI and not us running the tests locally

internal/setupkeys/setupkeys.go
logan-scully-arm: IMO all code should be written with how testable it is in mind. I think the mocking is a worthwhile sacrifice for that.

I can understand your point tho.

internal/setupkeys/setupkeys_test.go
tgonzalezorlandoarm: done: https://github.com/arm/topo/pull/34/commits/1a4bee36a476defd0fa6a192688d2d3790ecea71

internal/setupkeys/setupkeys.go
awphi: > Part of the reason that we were deviating so much with the ExecSSH functions to the point it needed some refactoring in https://github.com/arm/topo/pull/24 is because we are currently injecting that ExecSSH into the Connection, when in fact it could just be an internal method.

I agree this is an issue but I don't think the global state approach is the answer - it introduces extra boilerplate to the tests and prevents us from safely running test packages in parallel that manipulate this global state. I can foresee confusing error messages and inconsistent CI failures in the feature with this approach.

However, the main problem with it is it makes these functions impure and adds an implicit dependency the caller of the function must be aware about but isn't exposed to them via an argument which makes it harder to reason about imo. State is evil and global state even moreso so unless we absolutely have to use it it's best avoided imo.

> In fact, the testing logic is affecting so much that the production code on that regard, that we cannot easily implement a Connection.DryRun() because it's very hard to do so with ExecSSH function injection

I think this has other solutions - i.e. passing a function that returns `*exec.Cmd` into `NewConnection` instead of one that creates and then runs them - then it's trivial to implement a Connection.DryRun. The problem is with the dependency being injected into the connection, not the method of dependency injection itself.

logan-scully-arm: > passing a function that returns *exec.Cmd into NewConnection

I like the sound of this approach ngl

internal/setup_keys/setup_keys_test.go
awphi: I'm still not a great fan of this regardless tbh - if we're not actually testing `ssh-keygen` I wonder if it should ever be invoked by code that we're testing at all. It might be better to just check that our code produces the expected `ssh-keygen` command string like we do for most of our deploy tests with docker commands

internal/setupkeys/setupkeys.go
tgonzalezorlandoarm: Sure, I've created #35 to start with this process

internal/setupkeys/setupkeys.go
tgonzalezorlandoarm: Sure, will change this patch to inject a mock function as well, with the `*exec.Cmd` instead of the exec function

I had thought of the `*exec.Cmd` tbh and I can't remember why I didn't go through with it, but given that it has support from both of you and the argument about state is a good one, I'll just change this PR accordingly :cry:

internal/setupkeys/setupkeys.go
awphi: Ah didn't even realise it was in `deploy/docker` - ty!

tgonzalezorlandoarm: The problem with this here is that the operation for `SSHKeyGen` is doable, but for `PubKeyTransfer` we need to go through a `Connection` as all it is is an ssh command.
If we want to make it dry-runnable, we need to make Connection dry-runnable, which deserves its own work as well

tgonzalezorlandoarm: Is it ok with you if I make the `SSHKeyGen` command available but not the `PubKeyTransfer` (and we go throuigh the function here) until the refactoring of Connection is done?

tgonzalezorlandoarm: As discussed offline, I'll do some refactoring later for both of the commands together, after making Connection Dry-Runnable

cmd/topo/setup_keys.go
tgonzalezorlandoarm: done!

internal/setup_keys/setup_keys_test.go
```diff
@@ -0,0 +1,196 @@
...
+func TestNewKeyCreationAndPlacementOnTargetDryRun(t *testing.T) {
+	tests := []struct {
+		name         string
+		inputKeyPath string
+		wantKeyPath  string
+	}{
+		{
+			name:         "default key path",
+			inputKeyPath: "",
+			wantKeyPath:  filepath.Join(".ssh", "id_ed25519_topo_user_example.com"),
```
yejseo01, line 23: nit: It's good practice to not call example "example". It's much better for the readability of the tests if you name them properly like "id_ed25519_topo_beefy_user.com" and "id_ed25519_topo_porky_user.com"

internal/setup_keys/setup_keys_test.go
```diff
@@ -0,0 +1,196 @@
...
+				}
+			}
+
+			inputKeyPath := tt.inputKeyPath
+			if inputKeyPath != "" {
+				inputKeyPath = filepath.Join(tmp, inputKeyPath)
+			}
+
+			var buf bytes.Buffer
+			targetFileName := setup_keys.SanitizeTarget("user@example.com")
```
yejseo01, line 50: nit: same here. if you call all example examples, it gets quite difficult to locate which test is failing when it does fail. Try using something like `happyUser@pancakeLand.com` `unhappyUser@securityHell.com`

internal/setup_keys/setup_keys_test.go
```diff
@@ -0,0 +1,196 @@
...
+			transferErr := setup_keys.TransferPubKey("user@example.com", keyPath, &buf, true)
+			require.NoError(t, transferErr)
+			got = buf.String()
+			require.Contains(t, got, "ssh user@example.com", "DryRun output should include ssh command")
+			require.Contains(t, got, keyPath+".pub", "DryRun output should include public key path")
+		})
+	}
+}
+
+func TestKeyCreationAndPlacementOnTarget(t *testing.T) {
```
yejseo01: It seems like this test is trying to test two functions again. I think how we have been doing in the past is that we have a unit test file and an e2e test file. Unit test files would include all the unit test, and the tests in there would only examine the features of a single function at a time. e2e test file would examine how these functions are working together. TestKeyCreationAndPlacementOnTarget and TestNewKeyCreationAndPlacementOnTargetDryRun sounds like it belongs in e2e test files.

I suggest the following:
1. setup_keys.go: this file would include all the function's unit test, testing the creation of new keys and placements separately.
2. make another file called setup_keys_e2e.go: this file would include e2e tests such as TestKeyCreationAndPlacementOnTarget and TestNewKeyCreationAndPlacementOnTargetDryRun

If you think this is a different refactoring task, feel free to not include this change in this PR

yejseo01 (commented):
It seems like renaming of the file made the actual diffs much bigger than it is. Is it possible to have the renaming of the file in a different PR?

internal/setup_keys/setup_keys_test.go
tgonzalezorlandoarm: no mocking in this PR now, we can discuss this later on a future PR

tgonzalezorlandoarm: But I'd say that we do end up running the docker deployments

internal/setupkeys/setupkeys.go
tgonzalezorlandoarm: I think this needs some proper refactoring for Connection, so this will have to be dealt with later in a future PR

internal/setup_keys/setup_keys_test.go
tgonzalezorlandoarm: I think this is the way to go. Unfortunately, there's several refactorings that need to happen, mainly because `Connection` and how it handles `ExecSSH`

tgonzalezorlandoarm: So I'll apply this in a new PR

tgonzalezorlandoarm:
Thanks for all the useful feedback. Will Close this PR and create smaller ones following up on the feedback

tgonzalezorlandoarm:
As there are several refactors involved

PR #35, opened 2026-02-20
chore: make Connection dry-runnable

tgonzalezorlandoarm:
Currently:
1. The `command` package is being used by `Connection` but it's currently located under docker operations. As this is being outside of docker , we should move this to a more generic place.
2. Because we are injecting ExecSSH as a function to Connection, the command is not dry-runnable, as functions are not  dry-runnable (only *exec.Cmd are dry-runnable). Note: we are injecting ExecSSH so that we can afterwards mock test it.

## Changes

- Move the `command` package to an internal package instead of a `docker` one.
- Instead of injecting a `function to be executed` to Connection, inject a `function that returns a command to be executed`. This allows us `Connection.DryRun()`

Connection being dry-runnable will come in handy for other operations that:
1. Use Connection in their underlying implementation (like install or health)
2. need to be dry-runnable

This will be the case of `setup-keys`, whose implementation needs to change to use Connection and invoke ssh commands in order for it to be Windows compatible, as `ssh-copy-id` is not available on Windows.


internal/target/connection.go
```diff
@@ -62,11 +65,27 @@ func (c *Connection) Run(command string) (string, error) {
...
-	stdout, err := c.exec(c.SSHTarget, command, c.opts.WithStdin)
+	cmd := c.exec(c.SSHTarget, command, c.opts.WithStdin)
+	var stdoutBuf, stderrBuf bytes.Buffer
+	cmd.Stdout = &stdoutBuf
+	cmd.Stderr = &stderrBuf
+
+	err := cmd.Run()
 	if err != nil {
-		return "", err
+		combined := stdoutBuf.String() + stderrBuf.String()
```
logan-scully-arm, line 75: I have no clue about this so I could be wrong, but I'm not sure if we should me mixing out & err in  this way. 

I'd probably just return out and then keep the err wrapped in the `Run()`'s `errorf`.

internal/command/command.go
logan-scully-arm, line 1: Could you explain the rationale for moving this out of Docker? I see that it might enable additional use cases, but which ones are you targeting?

tgonzalezorlandoarm: This is going to be used by setup-keys ' s SSHKeyGen operation. Perhaps this commit in particular should be moved to that PR instead.

internal/target/connection.go
tgonzalezorlandoarm: This is what we were already doing in
`func Exec(target Host, command string, stdin []byte, sshArgs ...string) (string, error) {`
(see the code removal in `internal/ssh/exec.go`
I'm replicating the same logic here.

This was fine as far as I can tell, we have the error code as well to inform us of whether what we are getting is an error or not

internal/command/command.go
tgonzalezorlandoarm: done!

internal/target/connection.go
logan-scully-arm: I must've missed that there :laughing: Not a subject of this PR in that case, but I'd still say it's be better to just return out and err separately. 

Going to resolve this :+1:

internal/command/command.go
tgonzalezorlandoarm: Ah, there we go, when I removed the commit, the CI fails as the `Connection` uses command already but it's not docker related!

tgonzalezorlandoarm: @logan-scully-arm I removed and then re-added the commit so going to leave this thread to be resolved by you as I added some accidental noise!

logan-scully-arm (approved):
LGTM

PR #36, opened 2026-02-20
feat: support multiplexing target ssh connections

awphi:
## Changes

- Adds a new `Multiplex` option for `target.Connection` to support appending SSH multiplexing args to our invocation of ssh  - this means (on non-Windows systems) ssh will re-use established connections within a given timeout, cutting out a lot time spent re-handshaking
- Utilizes the multiplexing in `topo health`, `topo describe` and `topo templates`. Left out `topo install` for now as a) it's run far less frequently than these others so the extra ms don't matter too much and b) it would benefit from a bit of refactoring to use a single `target.Connection` before making this change

Perf improvement for health seems to be ~2-3x - down from 4.5-5s on the imx93 to 1.5-2s on my machine (outside the office).


tgonzalezorlandoarm:
Could we have this merged after #37 if possible?

logan-scully-arm (approved):
Fantastic Addition!!

tgonzalezorlandoarm (approved):

tgonzalezorlandoarm:
> Could we have this merged after #37 if possible?

This can be merged now actually, instead of 37 I should have said 35. Thanks for waiting!

PR #37, opened 2026-02-21
chore: allow for an injectable ssh hook without modifying API

tgonzalezorlandoarm:
Keep production behavior untouched while enabling SSH mocking in tests.
This avoids runtime risk (e.g., accidental test code paths in prod, API drift, or hidden side effects).

## Changes

- add an `WithMockExec` option to `ConnectionOptions` to explicitly mock the SSH function in Connection.
- Remove the execSSH function parameter in `NewConnection`, as this should default to the normal execSSH function.

### DIFF
--- a/internal/target/connection.go
+++ b/internal/target/connection.go
@@ -50,14 +50,19 @@ type ConnectionOptions struct {
 	WithLoginShell    bool
 	WithStdin         []byte
 	Multiplex         bool
+	WithMockExec      execSSH
 }
 
 var ErrPasswordAuthentication = errors.New("only password authentication is configured; key-based ssh is required")
 
-func NewConnection(sshTarget string, exec execSSH, opts ConnectionOptions) Connection {
+func NewConnection(sshTarget string, opts ConnectionOptions) Connection {
+	execFn := ssh.ExecCmd
+	if opts.WithMockExec != nil {
+		execFn = opts.WithMockExec
+	}
 	return Connection{
 		SSHTarget: ssh.Host(sshTarget),
-		exec:      exec,
+		exec:      execFn,
 		opts:      opts,
 	}
 }
--- a/internal/target/connection_auth_test.go
+++ b/internal/target/connection_auth_test.go
@@ -63,8 +63,8 @@ func TestProbeAuthentication(t *testing.T) {
 		mockExec := newMockExec(map[string]mockResponse{
 			"public": {stdout: "", err: nil},
 		}, &calls)
-		opts := target.ConnectionOptions{AuthProbeEnabled: true, AcceptNewHostKeys: true}
-		conn := target.NewConnection("user@host", mockExec, opts)
+		opts := target.ConnectionOptions{AuthProbeEnabled: true, AcceptNewHostKeys: true, WithMockExec: mockExec}
+		conn := target.NewConnection("user@host", opts)
 		err := conn.ProbeAuthentication()
 		require.NoError(t, err)
 
@@ -78,8 +78,8 @@ func TestProbeAuthentication(t *testing.T) {
 		mockExec := newMockExec(map[string]mockResponse{
 			"public": {stdout: "Host key verification failed", err: errSSH},
 		}, &calls)
-		opts := target.ConnectionOptions{AuthProbeEnabled: true, AcceptNewHostKeys: true}
-		conn := target.NewConnection("user@host", mockExec, opts)
+		opts := target.ConnectionOptions{AuthProbeEnabled: true, AcceptNewHostKeys: true, WithMockExec: mockExec}
+		conn := target.NewConnection("user@host", opts)
 		err := conn.ProbeAuthentication()
 		require.ErrorIs(t, err, target.ErrHostKeyVerification)
 		require.Len(t, calls, 1)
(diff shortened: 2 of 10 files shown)


tgonzalezorlandoarm:
Please review #35 first

logan-scully-arm:
@tgonzalezorlandoarm I've had a quick look at #35 and am happy to approve after my two comments are replied to.

Would you be able to set these to merge into that branch rather than main so that I can have a look over these :pray: It'll change back to main once that is merged

tgonzalezorlandoarm:
> @tgonzalezorlandoarm I've had a quick look at #35 and am happy to approve after my two comments are replied to.
> 
> Would you be able to set these to merge into that branch rather than main so that I can have a look over these 🙏 It'll change back to main once that is merged

Done!

internal/target/connection_auth_test.go
```diff
@@ -64,7 +64,8 @@ func TestProbeAuthentication(t *testing.T) {
 			"public": {stdout: "", err: nil},
 		}, &calls)
 		opts := target.ConnectionOptions{AuthProbeEnabled: true, AcceptNewHostKeys: true}
-		conn := target.NewConnection("user@host", mockExec, opts)
+		conn := target.NewConnection("user@host", opts)
+		conn.AddMock(mockExec)
```
logan-scully-arm: I'm personally not a fan of this pattern. IMO constructing w/ the mock is much cleaner.

But maybe see if this passes w/ @th3james before changing it

awphi: What about folding it into `ConnectionOptions`?

```go
type ConnectionOptions struct {
	// ...
    Exec execSSH
}

func NewConnection(sshTarget string, opts ConnectionOptions) Connection {
    execFn := ssh.ExecCmd
    if opts.Exec != nil {
        execFn = opts.Exec
    }
	return Connection{
		SSHTarget: ssh.Host(sshTarget),
		exec:      execFn,
		opts:      opts,
	}
}
```

This lets us scrap `getSSHFunction` and `AddMock`

logan-scully-arm: I like the idea of rolling it into exec ops a lot more yeah.

tgonzalezorlandoarm: That gets the best of both worlds, it's a good compromise!

tgonzalezorlandoarm: Will implement

tgonzalezorlandoarm: done: https://github.com/arm/topo/compare/7305c931ee0514545050ce2e849ce4c3e6bfd445..ee72e74f230bde7ee3ea82b54d968e445798118a

tgonzalezorlandoarm: forced push, as it's only one commit. You should be able to review the new commit from scratch

internal/target/connection.go
```diff
@@ -50,14 +50,19 @@ type ConnectionOptions struct {
 	WithLoginShell    bool
 	WithStdin         []byte
 	Multiplex         bool
+	WithMockExec      execSSH
```
logan-scully-arm, line 53: NIT: I'd make this more generic to WithExec -- although there's probably not a reason right now, I wouldn't restrict this just to mocking.

federicobozzini: I agree here. This metioning the mock feels unnecessary specfiic.

tgonzalezorlandoarm: One of the reasons for this PR is to actually restrict that.
The fragmentation that required #24 was caused because we actually allowed to inject more than one function.

`c.exec` would be a method of `Connection` and not an injectable function if it weren't for the mocking logic that we need to support. I think we should make this clear. If the user requires a different execSSH function, they should modify the would-be method, not inject a new one

tgonzalezorlandoarm: Our testing logic is interfering with the production code. That's fine, but let's restrict that and make it clear that we are making this for testing, otherwise we end up with multiple exec functions and then a refactoring PR like #24

internal/target/connection_auth_test.go
tgonzalezorlandoarm: I have to say though, we'll need this pattern to be reusable for other strcutures where we want to inject a mock testing function. This will happen for example when adding Windows support. Do we want to create Options for all of these as well? If not, I suggest the `AddMock` approach.
I'll assume we want the `Options` approach for the rest of the cases, let me know otherwise!

internal/target/connection.go
logan-scully-arm: Yeah that's a fair point. You're right that it's probably best to keep this explicit in its purpose.

internal/target/connection_auth_test.go
awphi: What other structs are you thinking will need to follow this pattern? Imo just an arg (as we had before) is fine if there's no other options passed to `NewX` or folding it into an existing options struct like we've done here is fine either way

tgonzalezorlandoarm: > What other structs are you thinking will need to follow this pattern

Anything that we'd need a mock function for, like for example, mocking `ssh-keygen` to write the files that we need in testing.

Sure, resolving here

awphi (approved):

PR #38, opened 2026-02-21
feat: add Windows Public Key transfer support

tgonzalezorlandoarm:
## Changes

- remove Linux requirements for the related tests.
- replace the SetupKeys operation (which did not merit a separate abstraction) with the newly created SSHKeyGen and PubKeyTransfer operations.
- As PubKeyTransfer makes use of Connection (which uses `ssh`), this allows us to make a transfer of the public to the target without using ssh-copy-id.

### DIFF
--- a/cmd/topo/setup_keys.go
+++ b/cmd/topo/setup_keys.go
@@ -8,7 +8,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
-var setupKeysKeyPath string
+var privateKeysKeyPath string
 
 var setupKeysCmd = &cobra.Command{
 	Use:   "setup-keys",
@@ -30,7 +30,7 @@ Use --dry-run to see what commands would be executed without actually running th
 			return err
 		}
 
-		seq, err := setupkeys.NewKeySetup(resolvedTarget, setupKeysKeyPath)
+		seq, err := setupkeys.NewKeySetup(resolvedTarget, privateKeysKeyPath)
 		if err != nil {
 			return err
 		}
@@ -45,6 +45,6 @@ Use --dry-run to see what commands would be executed without actually running th
 func init() {
 	addTargetFlag(setupKeysCmd)
 	addDryRunFlag(setupKeysCmd)
-	setupKeysCmd.Flags().StringVar(&setupKeysKeyPath, "key-path", "", "Specify the SSH path where the generated key pair will be stored. Default directory: ~/.ssh. Default public key file name: id_ed25519_topo_<target>.pub)")
+	setupKeysCmd.Flags().StringVar(&privateKeysKeyPath, "key-path", "", "Specify the SSH path where the generated key pair will be stored. Default directory: ~/.ssh. Default public key file name: id_ed25519_topo_<target>.pub)")
 	rootCmd.AddCommand(setupKeysCmd)
 }
--- a/cmd/topo/setup_keys.go
+++ b/cmd/topo/setup_keys.go
@@ -3,7 +3,6 @@ package main
 import (
 	"fmt"
 	"os"
-	"runtime"
 
 	"github.com/arm/topo/internal/setupkeys"
 	"github.com/spf13/cobra"
@@ -26,16 +25,12 @@ Use --dry-run to see what commands would be executed without actually running th
 			panic(fmt.Sprintf("internal error: dry-run flag not registered: %v", err))
 		}
 
-		if runtime.GOOS != "linux" {
-			return fmt.Errorf("topo setup-keys currently supports Linux hosts only")
-		}
-
 		resolvedTarget, err := requireTarget(cmd)
 		if err != nil {
 			return err
 		}
 
-		seq, err := setupkeys.NewKeyCreationAndPlacementOnTarget(resolvedTarget, setupKeysKeyPath)
+		seq, err := setupkeys.NewKeyCreateAndPlaceSequence(resolvedTarget, setupKeysKeyPath)
 		if err != nil {
 			return err
 		}
--- a/cmd/topo/setup_keys.go
+++ b/cmd/topo/setup_keys.go
@@ -30,7 +30,7 @@ Use --dry-run to see what commands would be executed without actually running th
 			return err
 		}
 
-		seq, err := setupkeys.NewKeyCreateAndPlaceSequence(resolvedTarget, setupKeysKeyPath)
+		seq, err := setupkeys.NewKeySetup(resolvedTarget, setupKeysKeyPath)
 		if err != nil {
 			return err
 		}
--- a/cmd/topo/setup_keys.go
+++ b/cmd/topo/setup_keys.go
@@ -8,7 +8,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
-var privateKeysKeyPath string
+var privateKeyPath string
 
 var setupKeysCmd = &cobra.Command{
 	Use:   "setup-keys",
@@ -30,7 +30,7 @@ Use --dry-run to see what commands would be executed without actually running th
 			return err
 		}
 
-		seq, err := setupkeys.NewKeySetup(resolvedTarget, privateKeysKeyPath)
+		seq, err := setupkeys.NewKeySetup(resolvedTarget, privateKeyPath)
 		if err != nil {
 			return err
 		}
@@ -45,6 +45,6 @@ Use --dry-run to see what commands would be executed without actually running th
 func init() {
 	addTargetFlag(setupKeysCmd)
 	addDryRunFlag(setupKeysCmd)
-	setupKeysCmd.Flags().StringVar(&privateKeysKeyPath, "key-path", "", "Specify the SSH path where the generated key pair will be stored. Default directory: ~/.ssh. Default public key file name: id_ed25519_topo_<target>.pub)")
+	setupKeysCmd.Flags().StringVar(&privateKeyPath, "key-path", "", "Specify the SSH path where the generated key pair will be stored. Default directory: ~/.ssh. Default public key file name: id_ed25519_topo_<target>.pub)")
 	rootCmd.AddCommand(setupKeysCmd)
 }
--- a/internal/setupkeys/pubkeytransfer/pub_key_transfer_test.go
+++ b/internal/setupkeys/pubkeytransfer/pub_key_transfer_test.go
@@ -1,0 +1,60 @@
+package pubkeytransfer_test
+
+import (
+	"bytes"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"testing"
+
+	"github.com/arm/topo/internal/setupkeys/pubkeytransfer"
+	"github.com/arm/topo/internal/ssh"
+	"github.com/arm/topo/internal/testutil"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPubKeyTransferDryRun(t *testing.T) {
+	tmp := t.TempDir()
+	privKeyPath := filepath.Join(tmp, "id_ed25519_testrun")
+	op := pubkeytransfer.NewPubKeyTransfer("Transfer public key", "hola@chau", privKeyPath, pubkeytransfer.PubKeyTransferOptions{})
+
+	var buf bytes.Buffer
+	require.NoError(t, op.DryRun(&buf))
+	output := buf.String()
+	require.Contains(t, output, "ssh -- hola@chau")
+	require.Contains(t, output, "mkdir -p ~/.ssh && chmod 700 ~/.ssh && cat >> ~/.ssh/authorized_keys && chmod 600 ~/.ssh/authorized_keys")
+}
+
+func TestPubKeyTransferRun(t *testing.T) {
+	tmp := t.TempDir()
+	privKeyPath := filepath.Join(tmp, "id_ed25519_testdryrun")
+	pubKeyPath := privKeyPath + ".pub"
+	pubKeyContent := []byte("ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAItestkey")
+	require.NoError(t, os.WriteFile(pubKeyPath, pubKeyContent, 0o600))
+
+	type call struct {
+		host  ssh.Host
+		cmd   string
+		stdin []byte
+		args  []string
+	}
+	var got call
+
+	opts := pubkeytransfer.PubKeyTransferOptions{WithMockExec: func(h ssh.Host, command string, stdin []byte, args ...string) *exec.Cmd {
+		got = call{host: h, cmd: command, stdin: stdin, args: args}
+		cmd := testutil.CmdWithOutput("ssh invoked", 0)
+		if stdin != nil {
+			cmd.Stdin = bytes.NewReader(stdin)
+		}
+		return cmd
+	}}
+	op := pubkeytransfer.NewPubKeyTransfer("Transfer public key", "thing1@thing2.com", privKeyPath, opts)
+
+	var buf bytes.Buffer
+	require.NoError(t, op.Run(&buf))
+	require.Contains(t, buf.String(), "ssh invoked")
+	require.Equal(t, ssh.Host("thing1@thing2.com"), got.host)
+	require.Contains(t, got.cmd, "mkdir -p ~/.ssh && chmod 700 ~/.ssh && cat >> ~/.ssh/authorized_keys && chmod 600 ~/.ssh/authorized_keys")
+	require.Equal(t, pubKeyContent, got.stdin)
+	require.Empty(t, got.args)
+}
--- a/internal/setupkeys/setup_keys.go
+++ b/internal/setupkeys/setup_keys.go
@@ -16,10 +16,10 @@ func NewKeySetup(target string, privKeyPath string) (goperation.Sequence, error)
 	if privKeyPath == "" {
 		home, err := os.UserHomeDir()
 		if err != nil {
-			return nil, fmt.Errorf("failed to get home directory: %w", err)
+			return nil, fmt.Errorf("failed to determine default key path: %w", err)
 		}
 
-		keyName := fmt.Sprintf("id_ed25519_topo_%s", SlugifyTarget(target))
+		keyName := fmt.Sprintf("id_ed25519_topo_%s", slugifyTarget(target))
 		privKeyPath = filepath.Join(home, ".ssh", keyName)
 	}
 
@@ -30,7 +30,7 @@ func NewKeySetup(target string, privKeyPath string) (goperation.Sequence, error)
 	return goperation.NewSequence(ops...), nil
 }
 
-func SlugifyTarget(target string) string {
+func slugifyTarget(target string) string {
 	var b strings.Builder
 	for _, r := range target {
 		toWrite := '_'
--- a/internal/setupkeys/setup_keys.go
+++ b/internal/setupkeys/setup_keys.go
@@ -19,7 +19,7 @@ func NewKeyCreationAndPlacementOnTarget(target string, keyPath string) (goperati
 			return nil, fmt.Errorf("failed to determine home directory: %w", err)
 		}
 
-		keyName := fmt.Sprintf("id_ed25519_topo_%s", sanitizeTarget(target))
+		keyName := fmt.Sprintf("id_ed25519_topo_%s", SanitizeTarget(target))
 		keyPath = filepath.Join(home, ".ssh", keyName)
 	}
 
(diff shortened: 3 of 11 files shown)


tgonzalezorlandoarm:
Please review #37 first

internal/setup_keys/setup_keys.go
```diff
@@ -0,0 +1,46 @@
...
+	goperation "github.com/arm/topo/internal/deploy/operation"
+	"github.com/arm/topo/internal/setup_keys/pub_key_transfer"
+	"github.com/arm/topo/internal/setup_keys/ssh_keygen"
+)
+
+func NewKeyCreationAndPlacementOnTarget(target string, privKeyPath string) (goperation.Sequence, error) {
+	if privKeyPath == "" {
+		home, err := os.UserHomeDir()
+		if err != nil {
+			return nil, fmt.Errorf("failed to determine home directory: %w", err)
```
Luke-Parkin:
```suggestion
			return nil, fmt.Errorf("failed to get home directory: %w", err)
```

Maybe the nit-pickiest nitpick ever

internal/setup_keys/setup_keys.go
```diff
@@ -0,0 +1,46 @@
...
+	"path/filepath"
+	"strings"
+	"unicode"
+
+	goperation "github.com/arm/topo/internal/deploy/operation"
+	"github.com/arm/topo/internal/setup_keys/pub_key_transfer"
+	"github.com/arm/topo/internal/setup_keys/ssh_keygen"
+)
+
+func NewKeyCreationAndPlacementOnTarget(target string, privKeyPath string) (goperation.Sequence, error) {
```
Luke-Parkin: This function name feels a bit misleading and a bit clunky? We're not returning a 'KeyCreationAndPlacementOnTarget' object. 

Maybe something like


```suggestion
func NewKeyCreateAndPlaceSequence(target string, privKeyPath string) (goperation.Sequence, error) {
```

🤷‍♂️

internal/setup_keys/setup_keys.go
```diff
@@ -0,0 +1,46 @@
...
+	if privKeyPath == "" {
+		home, err := os.UserHomeDir()
+		if err != nil {
+			return nil, fmt.Errorf("failed to determine home directory: %w", err)
+		}
+
+		keyName := fmt.Sprintf("id_ed25519_topo_%s", SanitizeTarget(target))
```
Luke-Parkin: Should the Sanitize logic not be shared with the ssh sanitize logic?aka not reimplemented?

internal/ssh/exec.go
```diff
@@ -35,9 +35,9 @@ func ShellCommand(command string) string {
 	return fmt.Sprintf(`/bin/sh -c "exec ${SHELL:-/bin/sh} -l -c \"%s\""`, escaped)
 }
 
-// Exec runs a command on the target host. If the target is localhost, it runs locally.
+// ExecCmd builds a command to be executed on the target host. If the target is localhost, it will run locally when executed.
 // Pass stdin data as optional parameter, or nil for no stdin.
-func Exec(target Host, command string, stdin []byte, sshArgs ...string) (string, error) {
+func ExecCmd(target Host, command string, stdin []byte, sshArgs ...string) *exec.Cmd {
```
Luke-Parkin, line 40: This may have benefited from being split into another PR... I'm not quite sure why you've done this?

Luke-Parkin: Ah, I see. It's so you can create the dry run commands for every 'ExecCMD'. Hm. This makes it self contained..

I think I'm happy with this, but you'll need to refactor multiplexing back in!! :P

internal/target/connection.go
```diff
@@ -49,24 +52,54 @@ type ConnectionOptions struct {
...
+func NewConnection(sshTarget string, opts ConnectionOptions) Connection {
 	return Connection{
 		SSHTarget: ssh.Host(sshTarget),
-		exec:      exec,
+		exec:      nil,
 		opts:      opts,
 	}
 }
 
+func (c *Connection) AddMock(mockExecSSH execSSH) {
```
Luke-Parkin: This is adding a new pattern to the repo and I think deserves some discussion. 

Codex loves it, it's very well testable, and I do agree it's very useful. Butttt so far in the repo we've tended to stick to testing pure functions and using E2E tests to test action. Mocks provide some false confidence.

`Test behavior, not implementation` is a Bartek principle (I think)

logan-scully-arm: This was removed in the last PR that got merged -- that being said this has been the most mocked area of tests for a while now

tgonzalezorlandoarm: https://github.com/arm/topo/pull/37#discussion_r2840891995

I have to admit I prefer the AddMock pattern, but I was outnumbered

internal/ssh/exec.go
tgonzalezorlandoarm: This was done in another PR, rebased now!

internal/target/connection.go
tgonzalezorlandoarm: The way we are now doing mocking with the Options instead is bad for testing as I cannot assert.Equal() correctly as the pointer to the function may change

internal/setup_keys/setup_keys.go
tgonzalezorlandoarm: done: https://github.com/arm/topo/pull/38/commits/7d1298f933ed07b1e820bf07f539b06d22185353

internal/setup_keys/setup_keys.go
tgonzalezorlandoarm: done

internal/setup_keys/setup_keys.go
tgonzalezorlandoarm: `SanitizeTarget` basically transforms the `target` to something that is file compatible, so that we can create the target file
So "user@host" -> "user_host".
If I understand correctly, what you are referring about the `ssh sanitize logic` is not this, right?

Luke-Parkin: ah gotcha.

It probably should be renamed then. Any sanitization done is a side effect. The main purpose is creating a slug.

logan-scully-arm: `slugify_target`

tgonzalezorlandoarm: Done: https://github.com/arm/topo/pull/38/commits/289c91324374107fac1eb3624c0d45bb8cc39989

tgonzalezorlandoarm:
I recommend reviewing commit by commit on this one

cmd/topo/setup_keys.go
```diff
@@ -3,9 +3,8 @@ package main
 import (
 	"fmt"
 	"os"
-	"runtime"
 
-	"github.com/arm/topo/internal/setupkeys"
+	setup_keys "github.com/arm/topo/internal/setup_keys"
```
yejseo01: Is this necessary?
```suggestion
	"github.com/arm/topo/internal/setup_keys"
```

internal/setup_keys/setup_keys.go
```diff
@@ -0,0 +1,46 @@
...
+		toWrite := '_'
+		if unicode.IsLetter(r) || unicode.IsDigit(r) || r == '-' || r == '_' || r == '.' {
+			toWrite = r
+		}
+
+		b.WriteRune(toWrite)
+	}
+
+	sanitized := b.String()
+	return sanitized
```
yejseo01: nit: I wouldn't call these strings "sanitized" because `SlugifyTarget` won't make the target injection proof. It doesn't detect quotes or spaces. Maybe call it `slugified` instead?

internal/setup_keys/setup_keys_test.go
```diff
@@ -0,0 +1,79 @@
...
+			wantSSH := "ssh -- " + tt.wantTarget
+			wantCmd := "mkdir -p ~/.ssh && chmod 700 ~/.ssh && cat >> ~/.ssh/authorized_keys && chmod 600 ~/.ssh/authorized_keys"
+			require.Contains(t, buf.String(), wantKeygen, "DryRun output should include keygen command")
+			require.Contains(t, buf.String(), wantSSH, "DryRun output should include ssh command with correct target")
+			require.Contains(t, buf.String(), wantCmd, "DryRun output should include correct authorized key addition command to be executed on target")
+		})
+	}
+}
+
+func TestSanitizeTarget(t *testing.T) {
```
yejseo01: same here

cmd/topo/setup_keys.go
th3james: This is a regression on Go idioms
> By convention, packages are given lower case, single-word names; there should be no need for underscores or mixedCaps.

https://go.dev/doc/effective_go#package-names
```suggestion
	"github.com/arm/topo/internal/setupkeys"
```

yejseo01 (commented):
minor comments! Great work!

th3james (changes requested):

cmd/topo/setup_keys.go
tgonzalezorlandoarm: Nooo
This was changed because of another review comment in another PR :(

I think I can just remove the commit that changes this though :)

tgonzalezorlandoarm: @logan-scully-arm FYI

internal/setup_keys/setup_keys.go
tgonzalezorlandoarm: https://github.com/arm/topo/pull/38/commits/238168e830a2bb31fb886e7a7bb86ea88342373a

internal/setup_keys/setup_keys_test.go
tgonzalezorlandoarm: done: https://github.com/arm/topo/pull/38/commits/74de82ddb4b9e1bf9d2b99766f9c62e51ddcc098

cmd/topo/setup_keys.go
tgonzalezorlandoarm: Actually, I think you were referring to the file name Logan, which is different, sorry for the noise.
Addressed

tgonzalezorlandoarm: Addressed

internal/setupkeys/pubkeytransfer/pub_key_transfer_test.go
```diff
@@ -0,0 +1,61 @@
+// Only tests the public APIs of the public key transfer operation
```
awphi:
```suggestion
```

nit: Redundant comment as implied by the separate `_test` package

internal/setupkeys/sshkeygen/ssh_keygen.go
```diff
@@ -0,0 +1,65 @@
...
+	return &SSHKeyGen{
+		description: description,
+		targetHost:  targetHost,
+		keyType:     keyType,
+		keyPath:     keyPath,
+		exec:        execFn,
+	}
+}
+
+func (sshkg *SSHKeyGen) Description() string {
```
awphi: nit: Go convention is normally 1-2 letter var names on receivers and we've mostly adhered to that elsewhere in the codebase so love it or hate it think we should apply it here

awphi (approved):
Very nice work! Couple minor nits is all 🚀

internal/setupkeys/setup_keys.go
```diff
@@ -0,0 +1,45 @@
...
+	"path/filepath"
+	"strings"
+	"unicode"
+
+	goperation "github.com/arm/topo/internal/deploy/operation"
+	"github.com/arm/topo/internal/setupkeys/pubkeytransfer"
+	"github.com/arm/topo/internal/setupkeys/sshkeygen"
+)
+
+func NewKeyCreateAndPlaceSequence(target string, privKeyPath string) (goperation.Sequence, error) {
```
th3james: This name isn't consistent with other sequence generating functions, which don't include the  `Sequence` suffix, and I wonder if it's a bit How rather than What

```suggestion
func NewKeySetup(target string, privKeyPath string) (goperation.Sequence, error) {
```

cmd/topo/setup_keys.go
```diff
@@ -26,16 +25,12 @@ Use --dry-run to see what commands would be executed without actually running th
...
-			return fmt.Errorf("topo setup-keys currently supports Linux hosts only")
-		}
-
 		resolvedTarget, err := requireTarget(cmd)
 		if err != nil {
 			return err
 		}
 
-		seq, err := setupkeys.NewKeyCreationAndPlacementOnTarget(resolvedTarget, setupKeysKeyPath)
+		seq, err := setupkeys.NewKeyCreateAndPlaceSequence(resolvedTarget, setupKeysKeyPath)
```
th3james: I'd suggest renaming the key path argument, I find it a little unclear

```suggestion
		seq, err := setupkeys.NewKeyCreateAndPlaceSequence(resolvedTarget, privateKeyPath)
```

internal/setupkeys/setup_keys.go
```diff
@@ -0,0 +1,45 @@
...
+	if privKeyPath == "" {
+		home, err := os.UserHomeDir()
+		if err != nil {
+			return nil, fmt.Errorf("failed to get home directory: %w", err)
```
th3james: Explain what you were trying to do when the error occurred
```suggestion
			return nil, fmt.Errorf("failed to determine default key path: %w", err)
```

internal/setupkeys/setup_keys.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	ops := []goperation.Operation{
+		sshkeygen.NewSSHKeyGen("Generate SSH key pair for target", target, "ed25519", privKeyPath, sshkeygen.SSHKeyGenOptions{}),
+		pubkeytransfer.NewPubKeyTransfer("Transfer public key to target and set it as an authorized key", target, privKeyPath, pubkeytransfer.PubKeyTransferOptions{}),
+	}
+	return goperation.NewSequence(ops...), nil
+}
+
+func SlugifyTarget(target string) string {
```
th3james: I don't believe this should be exported, but if you do want it to be public it doesn't belong in the setupkeys namespace
```suggestion
func slugifyTarget(target string) string {
```

internal/setupkeys/sshkeygen/ssh_keygen.go
tgonzalezorlandoarm: done: https://github.com/arm/topo/compare/b752f06de81608aebe127234935c93fad43a41ae..4daa33b0275f6bffe943c1d656e6bab587a68e0e

internal/setupkeys/pubkeytransfer/pub_key_transfer_test.go
tgonzalezorlandoarm: done: https://github.com/arm/topo/compare/b752f06de81608aebe127234935c93fad43a41ae..4daa33b0275f6bffe943c1d656e6bab587a68e0e

internal/setupkeys/setup_keys.go
tgonzalezorlandoarm: done: https://github.com/arm/topo/pull/38/commits/e6397a144089d5278fcf1c65757e4d584213a369

cmd/topo/setup_keys.go
tgonzalezorlandoarm: https://github.com/arm/topo/pull/38/commits/2ecfebd9a7ae0d70354aeab91fb423c4027c1e6f

cmd/topo/setup_keys.go
logan-scully-arm: To be fair I was on about both -- wasn't aware of this. We do have a handful of files that follow a snake case `cli_provider`. 

@th3james should be changing these?
(sorry for adding more to this @tgonzalezorlandoarm )

internal/setupkeys/sshkeygen/ssh_keygen_test.go
```diff
@@ -0,0 +1,57 @@
...
+	require.NoError(t, op.Run(&buf))
+	require.True(t, got.called)
+	require.Equal(t, "ed25519", got.keyType)
+	require.Equal(t, keyPath, got.keyPath)
+	require.Equal(t, "user@example.com", got.target)
+	require.Contains(t, buf.String(), "ssh-keygen invoked")
+
+	_, err := os.Stat(filepath.Dir(keyPath))
+	require.NoError(t, err, "expected key directory to be created")
+}
```
th3james, line 37: This test is mostly verifying that the constructor passed its arguments through to the mock — but the test provided those exact arguments two lines above. The `got` struct and its assertions are just round-tripping the test's own inputs.

`Run` only does two things: creates the key directory and executes the command. The test already checks both of those at the bottom, making the argument-capture redundant.

Suggested simplification:

```suggestion
func TestSSHKeyGenRun(t *testing.T) {
	keyPath := filepath.Join(t.TempDir(), "keys", "id_ed25519_test")

	opts := ssh_keygen.SSHKeyGenOptions{
		WithMockKeyGen: func(keyType, keyPath, targetHost string) *exec.Cmd {
			return testutil.CmdWithOutput("ssh-keygen invoked", 0)
		},
	}
	op := ssh_keygen.NewSSHKeyGen("Generate SSH key pair", "user@example.com", "ed25519", keyPath, opts)

	var buf bytes.Buffer
	require.NoError(t, op.Run(&buf))
	require.Contains(t, buf.String(), "ssh-keygen invoked")

	_, err := os.Stat(filepath.Dir(keyPath))
	require.NoError(t, err, "expected key directory to be created")
}
```

The argument-wiring is already covered by `TestSSHKeyGenDryRun`, which asserts the full command string including key type, path, and target.

yejseo01 (approved):
LGTM!

tgonzalezorlandoarm:
@th3james is this good to go? I need your update since the "requested changes" button blocks the merge

internal/testutil/testutil_windows.go
```diff
@@ -3,10 +3,27 @@
...
+	volume := filepath.VolumeName(dir)
+	if volume == "" {
+		return
+	}
+	path := dir[len(volume):]
+	t.Setenv("HOMEDRIVE", volume)
+	if path == "" {
+		path = "\\"
+	}
+	t.Setenv("HOMEPATH", path)
```
th3james, line 24: Is any of this actually necessary?
```suggestion
```

internal/setupkeys/sshkeygen/ssh_keygen_test.go
```diff
@@ -0,0 +1,57 @@
+// Only tests the SSH key generation operation
```
th3james: ?
```suggestion
```

cmd/topo/setup_keys.go
th3james: Yes, we should try to respect Go's idioms here. This PR should seek to avoid making things worse, and we should probably tidy other sites up in a separate PR.

th3james (changes requested):

internal/testutil/testutil_windows.go
tgonzalezorlandoarm: Unfortunately if I just set `HOME` it failed in Windows. From what I understand of this, depending on the Windows system, we need to set other variables instead of just HOME. So this is th eway I found to be safe.
Please do let me know if my limited Windows knowledge is incorrect. I only found out about this when CI started failing

cmd/topo/setup_keys.go
tgonzalezorlandoarm: Just to confirm, is everything alright now?

tgonzalezorlandoarm: Regarding this PR. I'll clean up in another PR

internal/setupkeys/sshkeygen/ssh_keygen_test.go
tgonzalezorlandoarm: done: https://github.com/arm/topo/pull/38/commits/37ebd9ed4140bf8f63daff5204f496a6ce5443bf

internal/setupkeys/sshkeygen/ssh_keygen_test.go
tgonzalezorlandoarm: done: https://github.com/arm/topo/pull/38/commits/eaeee3fd7eb88c460562487695e58661383f2330

tgonzalezorlandoarm: Good point

logan-scully-arm (approved):
This is looking good now -- good job

tgonzalezorlandoarm:
@th3james Are u happy with this now?

th3james (approved):

PR #39, opened 2026-02-23
feat: modify SSH config after SSH key creation

tgonzalezorlandoarm:
## Changes

- Create a topo SSH configuration file to cover hosts and their associated keys for the keys created via the `topo setup-keys` command
- Add a line on the user's SSH config file to include that directory.

### DIFF
--- a/internal/setupkeys/sshconfig/ssh_config.go
+++ b/internal/setupkeys/sshconfig/ssh_config.go
@@ -1,0 +1,125 @@
+package sshconfig
+
+import (
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"slices"
+	"strings"
+
+	"github.com/arm/topo/internal/output/term"
+	"github.com/arm/topo/internal/ssh"
+)
+
+func ModifySSHConfig(targetHost string, privKeyPath string, targetSlug string, dryRun bool, output io.Writer) error {
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return fmt.Errorf("failed to determine home directory for SSH config: %w", err)
+	}
+
+	sshTopoConfigDir := filepath.Join(home, ".ssh", "topo_config")
+	sshTopoConfigPath := filepath.Join(sshTopoConfigDir, fmt.Sprintf("topo_%s.conf", targetSlug))
+
+	mainConfigPath := filepath.Join(home, ".ssh", "config")
+
+	// ssh config parsing expects forward slashes (even on Windows)
+	includeLine := fmt.Sprintf("Include %s", filepath.ToSlash(sshTopoConfigDir+"/*.conf"))
+
+	if dryRun {
+		if output == nil {
+			return errors.New("dry run requested but no output writer provided for SSH config changes")
+		}
+
+		if err := term.PrintHeader(output, "Update local SSH config for key-based authentication"); err != nil {
+			return err
+		}
+		if _, err := fmt.Fprintf(output, "Will update %s to include:\n- %s\n", mainConfigPath, includeLine); err != nil {
+			return err
+		}
+		if _, err := fmt.Fprintf(output, "Will create %s and place the target SSH config there pointing at the key that was just created.\n", sshTopoConfigPath); err != nil {
+			return err
+		}
+
+		return nil
+	}
+
+	if err := os.MkdirAll(sshTopoConfigDir, 0o700); err != nil {
+		return fmt.Errorf("failed to create %s: %w", sshTopoConfigDir, err)
+	}
+
+	existingTopoContent, errTopo := getFileContents(sshTopoConfigPath)
+	if errTopo != nil {
+		return errTopo
+	}
+
+	existingMainContent, errMain := getFileContents(mainConfigPath)
+	if errMain != nil {
+		return errMain
+	}
+
+	configBody := buildSSHConfigFragment(targetHost, privKeyPath)
+
+	if string(existingTopoContent) != configBody {
+		if err := os.WriteFile(sshTopoConfigPath, []byte(configBody), 0o600); err != nil {
+			return fmt.Errorf("failed to write %s: %w", sshTopoConfigPath, err)
+		}
+	}
+
+	if !hasIncludeLine(existingMainContent, includeLine) {
+		updated := slices.Concat([]byte(includeLine+"\n\n"), existingMainContent)
+		return os.WriteFile(mainConfigPath, updated, 0o600)
+	}
+
+	return nil
+}
+
+func getFileContents(filePath string) ([]byte, error) {
+	existingContent, err := os.ReadFile(filePath)
+	if err != nil && !os.IsNotExist(err) {
+		return nil, fmt.Errorf("failed to read %s: %w", filePath, err)
+	} else if os.IsNotExist(err) {
+		existingContent = []byte{}
+	}
+
+	return existingContent, nil
+}
+
+func hasIncludeLine(data []byte, includeLine string) bool {
+	for line := range strings.SplitSeq(string(data), "\n") {
+		trimmed := strings.TrimSpace(line)
+		if trimmed == "" || strings.HasPrefix(trimmed, "#") {
+			continue
+		}
+		if strings.EqualFold(trimmed, includeLine) {
+			return true
+		}
+	}
+	return false
+}
+
+func buildSSHConfigFragment(targetHost string, privKeyPath string) string {
+	user, host, port := ssh.SplitUserHostPort(targetHost)
+	hostAlias := host
+	if hostAlias == "" {
+		hostAlias = targetHost
+	}
+
+	var b strings.Builder
+	fmt.Fprintf(&b, "Host %s\n", hostAlias)
+	if host != "" && (strings.Contains(targetHost, "@") || strings.Contains(targetHost, ":")) {
+		fmt.Fprintf(&b, "  HostName %s\n", host)
+	}
+	if user != "" {
+		fmt.Fprintf(&b, "  User %s\n", user)
+	}
+	if port != "" {
+		fmt.Fprintf(&b, "  Port %s\n", port)
+	}
+
+	// needs to be this way even on Windows to work with ssh config parsing, which generally accepts forward slashes
+	fmt.Fprintf(&b, "  IdentityFile %s\n", filepath.ToSlash(privKeyPath))
+	b.WriteString("  IdentitiesOnly yes\n")
+	return b.String()
+}
--- a/internal/ssh/host.go
+++ b/internal/ssh/host.go
@@ -2,7 +2,9 @@ package ssh
 
 import (
 	"fmt"
+	"net"
 	"strings"
+	"unicode"
 )
 
 type Host string
@@ -15,4 +17,34 @@ func (h Host) IsPlainLocalhost() bool {
 
 func (h Host) AsURI() string {
 	return fmt.Sprintf("ssh://%s", h)
+}
+
+func (h Host) Slugify() string {
+	var b strings.Builder
+	for _, r := range h {
+		toWrite := '_'
+		if unicode.IsLetter(r) || unicode.IsDigit(r) || r == '-' || r == '_' || r == '.' {
+			toWrite = r
+		}
+		b.WriteRune(toWrite)
+	}
+	return b.String()
+}
+
+func SplitUserHostPort(raw string) (user, host, port string) {
+	hostPart := raw
+	if at := strings.LastIndex(raw, "@"); at != -1 {
+		user = raw[:at]
+		hostPart = raw[at+1:]
+	}
+
+	if strings.HasPrefix(hostPart, "[") && strings.HasSuffix(hostPart, "]") {
+		host = strings.TrimSuffix(strings.TrimPrefix(hostPart, "["), "]")
+		return user, host, port
+	}
+
+	if h, p, err := net.SplitHostPort(hostPart); err == nil {
+		return user, h, p
+	}
+	return user, hostPart, port
 }
--- a/cmd/topo/setup_keys.go
+++ b/cmd/topo/setup_keys.go
@@ -5,6 +5,8 @@ import (
 	"os"
 
 	"github.com/arm/topo/internal/setupkeys"
+	"github.com/arm/topo/internal/setupkeys/sshconfig"
+	"github.com/arm/topo/internal/ssh"
 	"github.com/spf13/cobra"
 )
 
@@ -30,15 +32,30 @@ Use --dry-run to see what commands would be executed without actually running th
 			return err
 		}
 
+		targetSlug := ssh.Host(resolvedTarget).Slugify()
+		if privateKeyPath == "" {
+			privateKeyPath, err = setupkeys.GetDefaultPrivateKeyPath(targetSlug)
+			if err != nil {
+				return err
+			}
+		}
+
 		seq, err := setupkeys.NewKeySetup(resolvedTarget, privateKeyPath)
 		if err != nil {
 			return err
 		}
 
 		if dryRun {
-			return seq.DryRun(os.Stdout)
+			err = seq.DryRun(os.Stdout)
+		} else {
+			err = seq.Run(os.Stdout)
 		}
-		return seq.Run(os.Stdout)
+
+		if err != nil {
+			return err
+		}
+
+		return sshconfig.ModifySSHConfig(resolvedTarget, privateKeyPath, targetSlug, dryRun, os.Stdout)
 	},
 }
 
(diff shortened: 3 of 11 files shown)


tgonzalezorlandoarm:
Please review #38 first

Luke-Parkin:
>Please review https://github.com/arm/topo/pull/38 first

Maybe base it off of 38? Looks like it should be pointing towards 38 rather than main?

internal/ssh/host.go
```diff
@@ -16,3 +17,15 @@ func (h Host) IsPlainLocalhost() bool {
 func (h Host) AsURI() string {
 	return fmt.Sprintf("ssh://%s", h)
 }
+
+func (h Host) SlugifyTarget() string {
```
th3james: this doesn't operate on Target, it operates on Host
```suggestion
func (h Host) Slugify() string {
```

th3james: This should be easy to write some tests for

internal/setupkeys/sshconfig/ssh_config.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	if user != "" {
+		fmt.Fprintf(&b, "  User %s\n", user)
+	}
+	if port != "" {
+		fmt.Fprintf(&b, "  Port %s\n", port)
+	}
+
+	// needs to be this way even on Windows to work with ssh config parsing, which generally accepts forward slashes
+	fmt.Fprintf(&b, "  IdentityFile %s\n", filepath.ToSlash(pubKeyPath))
```
th3james: Possible I'm missing something, but... shouldn't this be the private key?

internal/setupkeys/sshconfig/ssh_config.go
```diff
@@ -0,0 +1,166 @@
...
+	b.WriteString("  IdentitiesOnly yes\n")
+	return b.String()
+}
+
+func writeDryRunHeader(w io.Writer, description string) error {
```
th3james: This appears to be a duplicate of `printHeader` in `internal/deploy/operation/sequence.go` - work out where `PrintHeader` belongs as part of a public interface, then call it rather than duplicating the code

internal/setupkeys/sshconfig/ssh_config.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	if _, err := f.WriteString(includeLine + "\n"); err != nil {
+		return fmt.Errorf("failed to update %s: %w", mainConfigPath, err)
+	}
+	return nil
+}
+
+func hasIncludeLine(data []byte, includeLine string) bool {
+	lines := strings.Split(string(data), "\n")
+	for _, line := range lines {
```
th3james: The linter flagged that this is more efficient as it doesn't require allocating the entire lines array
```suggestion
	for line := range strings.SplitSeq(string(data), "\n") {
```

internal/setupkeys/sshconfig/ssh_config.go
```diff
@@ -0,0 +1,149 @@
...
+			return os.WriteFile(mainConfigPath, []byte(includeLine+"\n"), 0o600)
+		}
+		return fmt.Errorf("failed to read %s: %w", mainConfigPath, err)
+	}
+
+	if hasIncludeLine(data, includeLine) {
+		return nil
+	}
+
+	f, err := os.OpenFile(mainConfigPath, os.O_APPEND|os.O_WRONLY|os.O_CREATE, 0o600)
```
th3james: Since `data` is already fully read into memory, you can mutate the slice and write it back in one shot:

```suggestion
    if len(data) > 0 && data[len(data)-1] != '\n' {
        data = append(data, '\n')
    }
    data = append(data, []byte(includeLine+"\n")...)
    return os.WriteFile(mainConfigPath, data, 0o600)
```

internal/setupkeys/sshconfig/ssh_config.go
```diff
@@ -0,0 +1,149 @@
...
+	"net"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+func ModifySSHConfig(targetHost string, privKeyPath string, targetSlug string, dryRun bool, output io.Writer) error {
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return fmt.Errorf("failed to determine topo_config directory: %w", err)
```
th3james:
```suggestion
		return fmt.Errorf("failed to determine home directory for SSH config: %w", err)
```

internal/setupkeys/sshconfig/ssh_config.go
```diff
@@ -0,0 +1,149 @@
...
+
+	mainConfigPath := filepath.Join(home, ".ssh", "config")
+
+	// on Windows, ssh config parsing generally accepts forward slashes.
+	// sshConfigPath will be backslashes, which can be flaky
+	includeLine := fmt.Sprintf("Include %s", filepath.ToSlash(sshTopoConfigDir+"/*.conf"))
+
+	if dryRun {
+		if output != nil {
+			if _, err := fmt.Fprintf(output, "\n┌─ Update local SSH config for key-based authentication\n"); err != nil {
```
th3james: This should be calling the `printHeader` function. Outside of the scope of this PR, but we should probably have a CommandOutput with methods like `CommandOutput.WriteHeader` and `WriteLine`

internal/setupkeys/sshconfig/ssh_config.go
```diff
@@ -0,0 +1,149 @@
...
+	sshTopoConfigPath := filepath.Join(sshTopoConfigDir, fmt.Sprintf("topo_%s.conf", targetSlug))
+
+	mainConfigPath := filepath.Join(home, ".ssh", "config")
+
+	// on Windows, ssh config parsing generally accepts forward slashes.
+	// sshConfigPath will be backslashes, which can be flaky
+	includeLine := fmt.Sprintf("Include %s", filepath.ToSlash(sshTopoConfigDir+"/*.conf"))
+
+	if dryRun {
+		if output != nil {
```
th3james: This being false would be a programming error and should panic, rather than silently continuing

internal/setupkeys/sshconfig/ssh_config.go
```diff
@@ -0,0 +1,149 @@
...
+			if _, err := fmt.Fprintf(output, "Will create %s and place the target SSH config there pointing at the key that was just created.\n", sshTopoConfigPath); err != nil {
+				return err
+			}
+		}
+		return nil
+	}
+	if err := os.MkdirAll(sshTopoConfigDir, 0o700); err != nil {
+		return fmt.Errorf("failed to create %s: %w", sshTopoConfigDir, err)
+	}
+	if err := ensureIncludeLine(mainConfigPath, includeLine); err != nil {
```
th3james: Feels like we should do this _after_ writing the config, otherwise we're writing an include for a file that doesn't exist. Also, we should consider how we tear this down in case of subsequent commands failing. It might be worth doing more work in memory to compute all the state changes, then doing all the writes at the end so this is easier to guarantee atomically.

th3james (changes requested):

internal/setupkeys/sshconfig/ssh_config.go
tgonzalezorlandoarm: You are not missing anything, good catch!
It was passing because my target was already on the .ssh/config , I may need to remove it first.

internal/setupkeys/sshconfig/ssh_config.go
tgonzalezorlandoarm: I think this is an outdated comment, I've removed this as we probably are just fine with one print line. Will check printHeader though and come back to this

internal/setupkeys/sshconfig/ssh_config.go
th3james: Github has helpfully mangled the range of my suggestion - it should apply to the rest of this function.

Another note:
https://man.openbsd.org/ssh_config#Include
> Include directive may appear inside a Match or Host block to perform conditional inclusion.

Does this not create a problem for where we're inserting this entry? If we just append, the include will be conditionally scoped to the last `Host` block, right? As I understand it, if we want this to be global is needs to be prepended to the config prior to any Host directives, or we need to append it after the Host directive for Target

internal/setupkeys/sshconfig/ssh_config.go
```diff
@@ -0,0 +1,149 @@
...
+		fmt.Fprintf(&b, "  Port %s\n", port)
+	}
+
+	// needs to be this way even on Windows to work with ssh config parsing, which generally accepts forward slashes
+	fmt.Fprintf(&b, "  IdentityFile %s\n", filepath.ToSlash(pubKeyPath))
+	b.WriteString("  IdentitiesOnly yes\n")
+	return b.String()
+}
+
+func SplitUserHostPort(raw string) (user, host, port string) {
```
awphi: We have an extremely similar function called `splitHostPort` in `ssh_tunnel.go` but yours has a more descriptive name. Could we merge the two and move it into the `ssh` package? Happy if you want to do this in a follow up ofc

internal/ssh/host.go
tgonzalezorlandoarm: done: https://github.com/arm/topo/pull/39/commits/5ca83b1931fd293e399ea302ee9ad880885e2f29

tgonzalezorlandoarm: https://github.com/arm/topo/pull/39/commits/118af613bdce54b9d596d19564022ff7253b65b1

internal/setupkeys/sshconfig/ssh_config.go
tgonzalezorlandoarm: https://github.com/arm/topo/pull/39/commits/83b83da8875c707a2035676dacb6de958c1e45db

internal/setupkeys/sshconfig/ssh_config.go
tgonzalezorlandoarm: https://github.com/arm/topo/pull/39/commits/be87f7e573c2e739b7232830f00c819d146aaa5d

internal/setupkeys/sshconfig/ssh_config.go
tgonzalezorlandoarm: https://github.com/arm/topo/pull/39/commits/19bedc7bbfe148c637857690f9feb41088166ddf

internal/setupkeys/sshconfig/ssh_config.go
tgonzalezorlandoarm: https://github.com/arm/topo/pull/39/commits/9d9e3e3d54d600eb21f6b670cd6a9c2ab8686867

internal/setupkeys/sshconfig/ssh_config.go
logan-scully-arm: I'd probably say that instead of styling this as is , It might be good to create a logger and output these messages in the format `LOG: xyz`

internal/setupkeys/sshconfig/ssh_config.go
```diff
@@ -0,0 +1,141 @@
...
+				return err
+			}
+			if _, err := fmt.Fprintf(output, "Will update %s to include:\n- %s\n", mainConfigPath, includeLine); err != nil {
+				return err
+			}
+			if _, err := fmt.Fprintf(output, "Will create %s and place the target SSH config there pointing at the key that was just created.\n", sshTopoConfigPath); err != nil {
+				return err
+			}
+		} else {
+			return errors.New("dry run requested but no output writer provided for SSH config changes")
```
logan-scully-arm, line 32: as a secondary to what @th3james said, it would be nicer to do this as the `if` rather than have an `if-else` flow

internal/setupkeys/sshconfig/ssh_config.go
```diff
@@ -0,0 +1,141 @@
...
+		return fmt.Errorf("failed to determine home directory for SSH config:: %w", err)
+	}
+
+	sshTopoConfigDir := filepath.Join(home, ".ssh", "topo_config")
+	sshTopoConfigPath := filepath.Join(sshTopoConfigDir, fmt.Sprintf("topo_%s.conf", targetSlug))
+
+	mainConfigPath := filepath.Join(home, ".ssh", "config")
+
+	// on Windows, ssh config parsing generally accepts forward slashes.
+	// sshConfigPath will be backslashes, which can be flaky
```
awphi: So if I've understood correctly this comment is saying we convert the backslash-delimited path from `filepath.Join` to a forward-slash one on windows because ssh config prefers UNIX-style paths? Would suggest clarifying the comment in this case


```suggestion
	// ssh config parsing expects forward slashes (even on Windows)
```

awphi (commented):
Few small comments but yet more nice work

internal/setupkeys/sshconfig/ssh_config.go
tgonzalezorlandoarm: done: https://github.com/arm/topo/pull/39/commits/5604352e404fdbbb93ee0d207ed6b993359eedc1

internal/setupkeys/sshconfig/ssh_config.go
tgonzalezorlandoarm: done

internal/setupkeys/sshconfig/ssh_config.go
tgonzalezorlandoarm: Moving the printHeader discussion here and resolving this thread
https://github.com/arm/topo/pull/39#discussion_r2847688634

internal/setupkeys/sshconfig/ssh_config.go
tgonzalezorlandoarm: Moved the PrintHeader function as indicated: https://github.com/arm/topo/pull/39/commits/aaa5a9bfeb910f9b818eec85efcda6ae2b38358b

internal/setupkeys/sshconfig/ssh_config.go
tgonzalezorlandoarm: Very miuch related to this https://github.com/arm/topo/pull/39#discussion_r2847851279 so resolving here and moving the conversation there

internal/setupkeys/sshconfig/ssh_config.go
tgonzalezorlandoarm: done: https://github.com/arm/topo/pull/39/commits/00b0568c68dd3802760be4374cc380bbfb121207

internal/setupkeys/sshconfig/ssh_config.go
tgonzalezorlandoarm: I wonder why the linter hasn't flagged me about this

internal/setupkeys/sshconfig/ssh_config.go
tgonzalezorlandoarm: > needs to be prepended to the config prior to any Host directives, or we need to append it after the Host directive for Target

You are right. I'm working on the latter

tgonzalezorlandoarm: I ended up going for the prepend line after all (way easier and if the user is not using ssh keys I don't expect us to be intruding with their config by adding the config at the beginning):
https://github.com/arm/topo/pull/39/commits/39eb6d47f928198eac58359cf13b9a9bd6748a33

internal/setupkeys/sshconfig/ssh_config.go
tgonzalezorlandoarm: https://github.com/arm/topo/pull/39/commits/1c036bb23adbcbfc4c71ad58b33e22cd48148f12

tgonzalezorlandoarm: I'm now working on this:

> It might be worth doing more work in memory to compute all the state changes, then doing all the writes at the end so this is easier to guarantee atomically.

so hold tight

tgonzalezorlandoarm: ok, https://github.com/arm/topo/pull/39/commits/0d91c7419de58933d143dcfdfbb9fdd78ea27ffe

I've done this now.
I'm not strictly guaranteeing atomicity, but I think this is good enough.

tgonzalezorlandoarm:
I think I've addressed all commnets here now , please lmk otherwise

internal/setupkeys/sshconfig/ssh_config.go
```diff
@@ -0,0 +1,125 @@
...
+	"strings"
+
+	"github.com/arm/topo/internal/output/term"
+	"github.com/arm/topo/internal/ssh"
+)
+
+func ModifySSHConfig(targetHost string, privKeyPath string, targetSlug string, dryRun bool, output io.Writer) error {
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return fmt.Errorf("failed to determine home directory for SSH config:: %w", err)
```
th3james: Typo
```suggestion
		return fmt.Errorf("failed to determine home directory for SSH config: %w", err)
```

th3james (approved):

PR #42, opened 2026-02-23
Sort out the ID-Name confusion

logan-scully-arm:
## Changes

- Sorts the ID-Name confusion by making everything name.
- Removes an unnecessary test from `catalog_test`


internal/catalog/data/templates.json
```diff
@@ -1,39 +1,39 @@
 [
   {
-    "id": "Hello-World",
+    "Name": "Hello-World",
```
awphi: nit: Why is Name title case here rather than lower case like all the other keys?

scripts/update_templates/main.go
```diff
@@ -9,14 +9,14 @@ import (
...
-	"Arm-Examples/topo-template-simd-visual-benchmark#main",
+	"Arm-Examples/topo-welcome#main",
+	"Arm-Examples/topo-lightbulb-moment#main",
+	"Arm-Examples/topo-v9-cpu-chat#main",
+	"Arm-Examples/topo-simd-visual-benchmark#main",
 }
 
 type Template struct {
-	ID          string   `json:"id"`
+	Name        string   `json:"Name"`
```
awphi: Ah now I see why it's capitalised - could we go with lower case pls?

scripts/update_templates/main.go
```diff
@@ -54,12 +54,12 @@ func main() {
 		}
 		tmpl.Ref = ref
 
-		if _, exists := seenIDs[tmpl.ID]; exists {
-			fmt.Fprintf(os.Stderr, "duplicate template id %q from %s; skipping\n", tmpl.ID, spec)
+		if _, exists := seenNamess[tmpl.Name]; exists {
```
awphi: typo on `seenNames`

awphi (changes requested):
Couple nits but this seems like an effective solution to the name/ID confusion

scripts/update_templates/parse.go
```diff
@@ -20,7 +20,7 @@ func BuildTemplate(repoURL string, compose io.Reader) (Template, error) {
 	}
 
 	return Template{
-		ID:          makeTemplateID(metadata.Name),
+		Name:        makeTemplateID(metadata.Name),
```
federicobozzini: Should this become `makeTemplateName`? Also maybe it should not return an empty string in case of failure.

scripts/update_templates/main.go
```diff
@@ -54,12 +54,12 @@ func main() {
...
+			fmt.Fprintf(os.Stderr, "duplicate template name %q from %s; skipping\n", tmpl.Name, spec)
```
federicobozzini: Rather than just skipping, should we stop the process and return an error here?

federicobozzini (commented):
Please remember to also change the template format docs.

yejseo01 (commented):
Just to note: this will break vscode-topo, and requires changes in vscode topo's usage of ids.

internal/catalog/catalog_test.go
```diff
@@ -9,19 +9,6 @@ import (
...
-		require.NoError(t, err)
-		assert.Equal(t, &catalog.Repo{
-			Id:          "Lightbulb-moment",
-			Description: "Reads a switch over GPIO pins on an M class cpu, reports switch state over Remoteproc Message, then a web application on the A class reads this and displays a lightbulb in either the on or off state. The lightbulb state is described by an LLM in any user-specified style.",
-			Features:    []string{"SVE", "NEON"},
-			Url:         "git@github.com:Arm-Examples/topo-template-lightbulb-moment.git",
-			Ref:         "main",
-		}, template)
-	})
-
```
yejseo01, line 24: why are you getting rid of this test?

logan-scully-arm: Because it relies on information from the JSON -- every time the json changes this needs changing. Feels to me more like its testing the contents of the json rather than the actual function -- I can put it back if you think it's needed.

tgonzalezorlandoarm: I agree with Yejin that this test should stay

yejseo01: I think it's best to keep this here. Catalog information and this test both exist in the same repo, so any changes to the json shouldn't be a problem.

logan-scully-arm: Brought it back :+1:

tgonzalezorlandoarm (approved):

awphi (approved):

PR #43, opened 2026-02-23
Add gosec to CI - Implement changes in current code

logan-scully-arm:
## Changes

- Adds a Gosec action to the GH-CI
- Adds `nosec` flags where it would have warned for validated inputs.
- Implements Gosec changes where possible to ensure current codebase passes checks

Currently removed the `G304` error, as it would like us to scope instead of using `os.Open()`. I'd like this to be done in an alternate PR, as it will take a large amount of refactor


.github/workflows/ci.yaml
```diff
@@ -92,4 +92,28 @@ jobs:
...
+
+    permissions:
+      contents: read
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Setup Go
+        uses: actions/setup-go@v5
```
tgonzalezorlandoarm: We are setting up go upstairs (liine 79) with a bigger version

.github/workflows/ci.yaml
```diff
@@ -92,4 +92,28 @@ jobs:
...
+          cache: true
+
+      - name: Install gosec
+        run: go install github.com/securego/gosec/v2/cmd/gosec@latest
+
+      - name: Add Go bin to PATH
+        run: echo "$(go env GOPATH)/bin" >> $GITHUB_PATH
+
+      - name: Run gosec (fail on findings)
+        run: gosec -severity high -confidence medium ./...
```
tgonzalezorlandoarm: How did you choose the severity and the confidence? I'd drop the severity as low as possible personally

e2e/bin.go
```diff
@@ -17,7 +17,7 @@ func buildBinary(t *testing.T) string {
 		binName = "topo.exe"
 	}
 	bin := filepath.Join(tmp, binName)
-	cmd := exec.Command("go", "build", "-o", bin, "../cmd/topo")
+	cmd := exec.Command("go", "build", "-o", bin, "../cmd/topo") // #nosec
```
tgonzalezorlandoarm: what warnings are we getting? Why is it ok to ignore? This should be written down

internal/health/health.go
```diff
@@ -8,6 +8,7 @@ import (
 	"github.com/arm/topo/internal/target"
 )
 
+// #nosec
 const passwordAuthErrorMessage = `note: Topo does not support SSH password-based authentication. To connect, either:
```
tgonzalezorlandoarm, line 13: what warning are we  getting here?

tgonzalezorlandoarm (changes requested):
We probably don't want to ignore every warning we get with a nosec, and we should justify why we are ignoring when we are

e2e/bin.go
logan-scully-arm: Yeah fair point, I'll write down on each one why the warning is ignored.

From trial end error I believe things like this will return warnings regardless a lot of the time due to how gosec works. I need to have another look tomorrow in case I was missing something glaringly obvious

.github/workflows/ci.yaml
logan-scully-arm: Something I need to look at tomorrow I think. 

When going through it seemed to warn about an excessive amount of stuff which would throw warnings.
For example `cmd := exec.Command(userInput)` would throw a warning regardless if you had checked it prior -- meaning you had to `nosec` it. This level of warning seemed the best to not litter nosec all over the codebase.

tgonzalezorlandoarm: I understand that POV for sure.
On the other hand, it's quite nice to have something that warns us when we are executing user input, because we may add some functionality that executes user input and we are not handling correctly and may not be aware of

logan-scully-arm: I'll set it to medium -- there will probably be a lot of `nosec` to review. 

But I guess it's a one and done problem

internal/health/health.go
```diff
@@ -8,6 +8,7 @@ import (
 	"github.com/arm/topo/internal/target"
 )
 
+// #nosec G101 -- Does not contain hardcoded credentials
```
yejseo01, line 12: Is this comment a way to wildcard out of gosec check?

internal/install/install.go
yejseo01, line 1: Are these changes based on the recommendation of the gosec?? it doesn't seem like anything to do with adding the gosec itself?

internal/testutil/container.go
yejseo01, line 1: same here

yejseo01 (commented):
Just confused if this PR is just adding gosec, or adding gosec along with implementing its advice

internal/health/health.go
logan-scully-arm: Yeah! This will block the `G101 ` check for example

internal/install/install.go
logan-scully-arm: Yeah I should probably add in the description. By adding it to the CI it means that it will fail unless these are all adhered to. 


Maybe there's a better way of doing that?

awphi (commented):
`gosec` should be [supported as a linter by golanglint-ci](https://golangci-lint.run/docs/linters/configuration/#gosec) which we're already using and has the advantage of just working with all our current tooling - should also require less/no CI yamling.

I mainly just want red squigglies in my editor where possible;

Image

logan-scully-arm:
> I mainly just want red squigglies in my editor where possible;

~~I'd like a second opinion on this. The ticket [[DEML-1317](https://jira.arm.com/secure/RapidBoard.jspa?rapidView=11701&projectKey=DEML&view=detail&selectedIssue=DEML-1317#)](https://jira.arm.com/secure/RapidBoard.jspa?rapidView=11701&projectKey=DEML&view=detail&selectedIssue=DEML-1317#) asks for CI, but I'm happy to go with this route if people would prefer that?~~

th3james:
@logan-scully-arm as I understand it, @awphi is suggesting that we should enable gosec using golangci-lint by adding:
```yaml
  linters:
    enable:
      - gosec
```
 to `.golangci.yaml`. This would then run it as part of the CI.

This is perhaps not intuitive unless you understand that golangci-lint is not actually a "linter", but a meta-linter which orchestrates the running of other linters

logan-scully-arm:
Oh thank you @th3james and @awphi, I think I completely misunderstood :laughing:

internal/deploy/project_checks/project_checks_test.go
awphi, line 1: WDYT about excluding security-related linters on files ending in `_test.go` project wide instead of case by case? https://github.com/golangci/golangci-lint/issues/2527#issuecomment-1023707262

internal/describe/describe.go
```diff
@@ -26,6 +26,7 @@ func GenerateTargetDescription(conn target.Connection) (target.HardwareProfile,
 
 func WriteTargetDescriptionToFile(dir string, report target.HardwareProfile) (string, error) {
 	outputFile := filepath.Join(dir, TargetDescriptionFilename)
+	// #nosec G304 -- dir and filename are valid
```
awphi: Is this required given the fact you've disable G304 across the whole project in the ci config?

awphi (commented):
Looks good + happy to have my squigglies 😄 Couple small comments still

internal/deploy/project_checks/project_checks_test.go
tgonzalezorlandoarm: I agree with this

logan-scully-arm: TY for sending that :+1: I'll do that now

internal/describe/describe.go
logan-scully-arm: Thanks, I think this must've been before I excluded it everywhere

tgonzalezorlandoarm:
We talked about this in offline but leaving this here as a record:
The general feeling I get is that as the codebase progresses, an issue that was previously discarded using #nosec should not be discarded anymore. In that case, we are losing that information.
As suggested by @logan-scully-arm , the approach may be to produce warnings instead and not error out. Then @th3james suggested using some kind of baseline comparison in CI to check whether there are new warnings that should be paid attention to.

I think the baseline comparison should probably be left for a different PR, but maybe we need the warnings here?

logan-scully-arm:
@tgonzalezorlandoarm I have brought back the severity level-check, this will silence the error but not ignore the warning. It will not remove the squiggles @awphi. Best of all worlds?

awphi:
> @tgonzalezorlandoarm I have brought back the severity level-check, this will silence the error but not ignore the warning. It will not remove the squiggles @awphi. Best of all worlds?

Tbf if CI isn't going to fail I'm indifferent to squigglies so sounds like a good approach!

internal/compose/compose_test.go
```diff
@@ -1,3 +1,4 @@
+// #nosec
```
awphi: Still needed?

awphi (approved):
One redundant-looking linter directive but looks good to me now!

tgonzalezorlandoarm (approved):

internal/compose/compose_test.go
logan-scully-arm: thanks thought I removed them all :laughing:

PR #44, opened 2026-02-23
ci: add Github Action to lint jobs and check pinned versions

tgonzalezorlandoarm:
https://jira.arm.com/browse/DEML-1318 required a simple check for the pinned actions

First ask yourself, do we actually need this?  I think not

## Changes

Signed-off-by: Tomás González <tomasagustin.gonzalezorlando@arm.com>

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,10 +12,10 @@ jobs:
     runs-on: [ ubuntu-latest ]
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v6
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       - name: Set up Go
-        uses: actions/setup-go@v6
+        uses: actions/setup-go@7a3fe6cf4cb3a834922a1244abfce67bcef6a0c5 # v6.2.0
         with:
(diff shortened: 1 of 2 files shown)


federicobozzini (commented):
Some of these actions you have pinned using the SHA are the official ones from GitHub. I think they might not need any SHA pinning?

tgonzalezorlandoarm:
> Some of these actions you have pinned using the SHA are the official ones from GitHub. I think they might not need any SHA pinning?

The tool does not distinguish unfortunately. I would have to make a separate script to handle these, which I didn't deem worthy

yejseo01:
Will review properly when CI passes?

Luke-Parkin:
I agree I don't think we need it

tgonzalezorlandoarm:
> Will review properly when CI passes?

@yejseo01 I guess what I want to get clear before I go through the effort of fixing the CI is whether or not we need the feature

tgonzalezorlandoarm:
Four people in the team already agree that this may not be needed
We are probably good with just reviewing correctly when an Action is changed.
We probably don't want to run a separate job (that we have to maintain and fix) and replicate it accross repos (although arguably this could be a re-usable workflow but with separate lockfiles)

For future reference, this may be what we are looking for as well: https://gh-actions-lockfile.net/

PR #46, opened 2026-02-24
Improved var naming consistency

federicobozzini:
## Changes

- Use `Model` consistenly instead of `ModelName`

### DIFF
--- a/internal/target/probe.go
+++ b/internal/target/probe.go
@@ -17,9 +17,9 @@ var armCpuFeatures = map[string]string{
 }
 
 type HostProcessor struct {
-	Model     string   `yaml:"model"`
-	Cores     int      `yaml:"cores"`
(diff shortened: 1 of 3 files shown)


internal/target/probe.go
```diff
@@ -17,9 +17,9 @@ var armCpuFeatures = map[string]string{
 }
 
 type HostProcessor struct {
-	ModelName string   `yaml:"model"`
-	Cores     int      `yaml:"cores"`
-	Features  []string `yaml:"features"`
+	Model    string   `yaml:"model"`
```
federicobozzini, line 20: This is the reason behind the change, ensuring the output (mentioning `model`) matches the internal field (previously `modelName`).

awphi (approved):

PR #47, opened 2026-02-25
feat: detection of gatewayport's enablement on dropbear and openssh o…

yejseo01:
Changes suggested while addressing this ticket - [DEML-1319](https://jira.arm.com/browse/DEML-1319)

## Changes

- Implement logic to detect the target's ssh server program: Currently detects for OpenSSH and dropbear
- Implement logic to detect if the gatewayport is open
- Warn if we cannot be sure that gatewayport is disabled
- Error if we can be sure if gatewayport is enabled

### DIFF
--- a/internal/deploy/project_checks/project_checks.go
+++ b/internal/deploy/project_checks/project_checks.go
@@ -6,6 +6,9 @@ import (
 	"strings"
 
 	"github.com/arm/topo/internal/compose"
+	"github.com/arm/topo/internal/output/logger"
+	"github.com/arm/topo/internal/ssh"
+	targetpkg "github.com/arm/topo/internal/target"
 )
 
 const linuxArm64Platform = "linux/arm64"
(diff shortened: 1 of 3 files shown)


yejseo01:
Apologies in the bad test format. It will be updated separately as a refactoring PR here - https://github.com/arm/topo/pull/48

internal/deploy/project_checks/project_checks.go
```diff
@@ -48,3 +51,169 @@ func EnsureProjectIsLinuxArm64Ready(composePath string) error {
 
 	return nil
 }
+
+func EnsureSSHGatewayPortsAreDisabled(target ssh.Host, connectionOption targetpkg.ConnectionOptions) []logger.Entry {
```
tgonzalezorlandoarm, line 55: These functions should probably be located elsewhere (ssh/Connection)
This module is for project definition checks

internal/deploy/project_checks/project_checks.go
```diff
@@ -48,3 +51,169 @@ func EnsureProjectIsLinuxArm64Ready(composePath string) error {
...
+	}
+
+	if hasOpenSSH {
+		return sshServerOpenSSH
+	}
+	if hasDropbear {
+		return sshServerDropbear
+	}
+	return sshServerUnknown
+}
```
tgonzalezorlandoarm, line 178:
```suggestion
func detectSSHServer(conn targetpkg.Connection) sshServerType {
	output, err := conn.Run("ps -eo comm")
	if err != nil {
		return sshServerUnknown
	}

	for line := range strings.SplitSeq(output, "\n") {
		switch strings.TrimSpace(line) {
		case "sshd":
			return sshServerOpenSSH
		case "dropbear":
			return sshServerDropbear
		}
	}

	return sshServerUnknown
}
...

File: internal/deploy/project_checks/project_checks.go [hunk 3]
```diff
@@ -48,3 +51,169 @@ func EnsureProjectIsLinuxArm64Ready(composePath string) error {
...
+	case "yes", "clientspecified":
+		return true
+	default:
+		return false
+	}
+}
+
+func parseGatewayPortsFromSSHD(sshdConfig string) (string, bool) {
+	lines := strings.Split(sshdConfig, "\n")
+	for i := len(lines) - 1; i >= 0; i-- {
```
tgonzalezorlandoarm, line 191: Why do we iterate backwards?

internal/deploy/project_checks/project_checks.go
```diff
@@ -48,3 +51,169 @@ func EnsureProjectIsLinuxArm64Ready(composePath string) error {
...
+		fields := strings.Fields(lines[i])
+		if len(fields) < 2 {
+			continue
+		}
+		if strings.EqualFold(fields[0], "gatewayports") {
+			return fields[1], true
+		}
```
tgonzalezorlandoarm, line 198:
```suggestion
		if len(fields) >= 2 && strings.EqualFold(fields[0], "gatewayports") {
			return fields[1], true
		}
```

internal/deploy/project_checks/project_checks.go
```diff
@@ -48,3 +51,169 @@ func EnsureProjectIsLinuxArm64Ready(composePath string) error {
...
+	}
+	return "", false
+}
+
+func parseGatewayPortsFromConfig(sshdConfig string) (value string, uncertain bool) {
+	lines := strings.Split(sshdConfig, "\n")
```
tgonzalezorlandoarm, line 204: Another opportunity to use SplitSeq?

internal/deploy/project_checks/project_checks.go
```diff
@@ -48,3 +51,169 @@ func EnsureProjectIsLinuxArm64Ready(composePath string) error {
...
+			})
+		} else if isGatewayPortsEnabled(gatewayPorts) {
+			logs = append(logs, logger.Entry{
+				Level:   logger.Err,
+				Message: fmt.Sprintf("SSH GatewayPorts must be disabled on %s: expected \"no\", got %q. Update sshd_config and restart sshd", target, gatewayPorts),
+			})
+		}
+
+		return logs
+	} else {
```
logan-scully-arm, line 114: This else is unnecessary -- you return in both blocks so you only need the initial if

yejseo01:
this way of resolving the problem is dismissed. Will make a different PR that attempts connection rather than being dependent on ssh server program's configuration

yejseo01:
different approach is selected and implemented here - https://github.com/arm/topo/pull/54

PR #48, opened 2026-02-25
refactor: make test matrix and reduce over-repeated function names in tests' names

yejseo01:
tests in `project_checks.go` were sore in the eyes because the names of the tests were trying to say everything what they were doing in camel cases. These descriptions also came after the statement of the function name the test is testing combined with _. Made the test matrix, and preserved the test description in more readable way

### DIFF
--- a/internal/deploy/project_checks/project_checks_test.go
+++ b/internal/deploy/project_checks/project_checks_test.go
@@ -13,76 +13,88 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestEnsureProjectIsLinuxArm64Ready_SucceedsWithValidRemoteProc(t *testing.T) {
-	composeFile := writeComposeFile(t, `
+func TestEnsureProjectIsLinuxArm64Ready(t *testing.T) {
+	cases := []struct {
+		name        string
+		compose     string
+		wantErr     bool
+		errContains string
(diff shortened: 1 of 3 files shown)


yejseo01:
Will merge this after https://github.com/arm/topo/pull/47 has been merged

internal/deploy/project_checks/project_checks_test.go
```diff
@@ -91,3 +107,135 @@ func writeComposeFile(t *testing.T, contents string) string {
...
+	cases := []struct {
+		name        string
+		target      ssh.Host
+		responses   map[string]mockedCommand
+		wantEmpty   bool
+		wantLevel   string
+		wantMessage string
+	}{
+		{
+			name:   "openssh allows when gateway ports disabled",
```
logan-scully-arm, line 126: I'm tend to be against this pattern. I believe in most cases where this was valid for tests, we have still defaulted to handwriting every test individually in the `t.Run(...` format.

However, I believe we do use this here and there, and I know a few people prefer this; so I'm not set on either method.

logan-scully-arm: Sorry I just realized this was a draft 😆

yejseo01:
not necessary anymore

PR #51, opened 2026-02-25
Check /proc/meminfo for total memory.

Luke-Parkin:
## Changes

- Adds a `totalmemory` field in the target-info.
We fetch this info from /proc/meminfo, as lsmem lists the exact same information.

### DIFF
--- a/internal/target/probe.go
+++ b/internal/target/probe.go
@@ -1,6 +1,7 @@
 package target
 
 import (
+	"bufio"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -29,6 +30,7 @@ type RemoteprocCPU struct {
 type HardwareProfile struct {
 	HostProcessor []HostProcessor `yaml:"host"`
 	RemoteCPU     []RemoteprocCPU `yaml:"remoteprocs"`
+	TotalMemory   int64
 }
 
 type LscpuOutputField struct {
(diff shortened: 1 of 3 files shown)


internal/target/probe.go
```diff
@@ -29,6 +30,7 @@ type RemoteprocCPU struct {
 type HardwareProfile struct {
 	HostProcessor []HostProcessor `yaml:"host"`
 	RemoteCPU     []RemoteprocCPU `yaml:"remoteprocs"`
+	TotalMemory   int64
```
awphi: Worth adding a yaml tag? If we're following compose yaml case convention should probably be `total_memory`

awphi: Would also be keen to get the unit in the same somehow

awphi (commented):
Looks nice - couple minor bits!

internal/target/probe.go
Luke-Parkin: should remoteprocs not then be remote_procs?

awphi: Very maybe but I think it's less clear since remoteproc is a term typically stylised as one word to refer to the kernel feature we're using. I guess it should be `remoteproc_drivers` but that seems a bit verbose + limiting imo

Luke-Parkin: https://github.com/arm/topo/pull/51/commits/0210c9c9f95f9ed617122e5b84c3f371f03d9cec

Luke-Parkin: https://github.com/arm/topo/pull/51/commits/0210c9c9f95f9ed617122e5b84c3f371f03d9cec

awphi (approved):

PR #52, opened 2026-02-26
Add govulncheck to the CI

logan-scully-arm:
## Changes

- Add Govulncheck to CI (I don't believe I can use golangci-lint)
- Update go version to deal with vulnerabilities.


.github/workflows/ci.yaml
```diff
@@ -22,6 +23,11 @@ jobs:
           cache-dependency-path: |
             **/go.sum
 
+      - name: Run govulncheck
+        run: |
+          go install golang.org/x/vuln/cmd/govulncheck@latest
```
awphi: I think we probably have to pin this to a known good version like we do our actions?

tgonzalezorlandoarm: I have the same feeling. Also, we probably don['t want to update go to a later version because of th9is

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/arm/topo
 
-go 1.25.0
```
tgonzalezorlandoarm, line 3: I don't think we want to be doing this

.github/workflows/ci.yaml
logan-scully-arm: Yeah good point.

.github/workflows/ci.yaml
```diff
@@ -22,6 +23,11 @@ jobs:
           cache-dependency-path: |
             **/go.sum
 
+      - name: Run govulncheck
+        run: |
+          go install golang.org/x/vuln/cmd/govulncheck@1.1.4
+          govulncheck ./...
```
tgonzalezorlandoarm, line 29: 1. Have you confirmed that it actually fails when there's a vulnerability in one of our dependencies?

logan-scully-arm: Yes, which is why I updated the GO version. Updating to `25.7` fixed the problems

logan-scully-arm:
@awphi and @tgonzalezorlandoarm for some reason all of my responses are stuck on `pending`. So here's a brief response.
- I couldn't seem to find a specific version to go with so have hadcoded the latest (1.1.4). This will still allow us to update in our own time.
- I tested this, and have updated the GO version as a response to vulnerable packages.

awphi (approved):

PR #53, opened 2026-02-26
fix: topo health works with unreachable target

awphi:
## Problem

- Basic problem is we had two probes doing similar things - `ProbeAuthentication` which runs `ssh <target> true` and makes some assertions about error messages and `ProbeConnection` which does the same but just treats any error as a failure
  - Since probing the connection was gated behind probing authentication but only when a specific connection option was set, our unit tests didn't catch this regression

## Solution  

- Turns out, afaict, the `AuthProbeEnabled` option is not necessary - we can run the probe fine and just treat the lack of ssh as a `ConnectionError`. Then, since we are separately checking for SSH on the path, you get exactly the same health report as you did before without the complex logic. This change on its own would mean our unit tests would now catch this bug
- Dropped `ProbeConnection` from health entirely as it's done by virtue of performing the auth check. This lets us avoid separating the errors and go back to treating any non-password-auth-related error as a `ConnectionError` and not crashing out the whole program.

End result is less code + more resilient tests, hence why no new ones are introduced here.

### DIFF
--- a/internal/health/health.go
+++ b/internal/health/health.go
@@ -110,11 +110,11 @@ func Check(sshTarget string, acceptNewHostKeys bool) (Report, error) {
 	conn := target.NewConnection(sshTarget, opts)
 	targetStatus := ProbeHealthStatus(conn)
 	report := GenerateReport(dependencyStatuses, targetStatus)
-	if err := targetStatus.AuthError; err != nil {
+	if err := targetStatus.ConnectionError; err != nil {
 		if errors.Is(err, target.ErrPasswordAuthentication) {
 			return report, fmt.Errorf(passwordAuthErrorMessage, sshTarget)
 		}
-		return report, err
+		return report, nil
 	}
 	return report, nil
 }
--- a/internal/health/health.go
+++ b/internal/health/health.go
@@ -92,16 +92,7 @@ func GenerateReport(hostDependencies []DependencyStatus, targetStatus Status) Re
 
 func Check(sshTarget string, acceptNewHostKeys bool) (Report, error) {
 	dependencyStatuses := CheckInstalled(HostRequiredDependencies, BinaryExistsLocally)
-
-	authProbeEnabled := false
-	for _, s := range dependencyStatuses {
-		if s.Dependency.Name == "ssh" {
-			authProbeEnabled = s.Installed
-			break
-		}
-	}
 	opts := target.ConnectionOptions{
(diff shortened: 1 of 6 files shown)


tgonzalezorlandoarm:
Nice catch!
Doesn't this mean we can remove `ProbeConnection` entirely though?
It's probably an unnecessary extra step in `GenerateTargetDescription` at this point

awphi:
> Nice catch! Doesn't this mean we can remove `ProbeConnection` entirely though? It's probably an unnecessary extra step in `GenerateTargetDescription` at this point

Probably true, the first hardware probe will fail which is probably useful enough context! Was thinking about replacing it with `ProbeAuthentication` but we probably don't want to check auth on every single command

tgonzalezorlandoarm:
> Probably true, the first hardware probe will fail which is probably useful enough context! Was thinking about replacing it with `ProbeAuthentication` but we probably don't want to check auth on every single command

Agreed, let's remove `ProbeConnection` entirely and this is good to go IMO, thanks Adam!

internal/health/health.go
```diff
@@ -110,11 +101,11 @@ func Check(sshTarget string, acceptNewHostKeys bool) (Report, error) {
 	conn := target.NewConnection(sshTarget, opts)
 	targetStatus := ProbeHealthStatus(conn)
 	report := GenerateReport(dependencyStatuses, targetStatus)
-	if err := targetStatus.AuthError; err != nil {
+	if err := targetStatus.ConnectionError; err != nil {
 		if errors.Is(err, target.ErrPasswordAuthentication) {
 			return report, fmt.Errorf(passwordAuthErrorMessage, sshTarget)
```
logan-scully-arm, line 106: Since the password auth error is now a part of connection error, do we even need to return the error as a part of this function. 

IMO we could just return the report as is and alter the PrintableHealthReport to also log the error as a part of the printing process?

tgonzalezorlandoarm (approved):
:rocket:

internal/health/health.go
awphi: I like the sound of that since you'd still get host health info even if your target is inaccessible due to password-based auth. I feel like it should be made as a separate change though since atm we don't surface the nature of `ConnectionError` to the user at all, just whether it's nil/non-nil. WDYT?

logan-scully-arm: Yeah ofc can be a different PR :+1:

PR #54, opened 2026-02-26
feat: security check by attempting to sniff the port

yejseo01:
Mitigating risk for [DEML-1319](https://jira.arm.com/browse/DEML-1319)

Attempt to sniff the open port from target after the SSH tunnel is established. If it succeeds, fail the deployment.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -120,7 +120,7 @@ jobs:
           OWNER: Arm-debug
           REPO: signer
           WORKFLOW_FILE: upload_archives_to_artifactory.yml
-          REF: pythonize
+          REF: main
           UPSTREAM_RUN_ID: ${{ github.run_id }}
         run: |
           gh workflow run "$WORKFLOW_FILE" \
@@ -138,7 +138,7 @@ jobs:
           OWNER: Arm-debug
           REPO: signer
           WORKFLOW_FILE: sign-and-post-macos.yml
-          REF: pythonize
+          REF: main
           UPSTREAM_RUN_ID: ${{ github.run_id }}
         run: |
           gh workflow run "$WORKFLOW_FILE" \
@@ -157,7 +157,7 @@ jobs:
           OWNER: Arm-debug
           REPO: signer
           WORKFLOW_FILE: sign-and-post-windows.yml
-          REF: pythonize
+          REF: main
           UPSTREAM_RUN_ID: ${{ github.run_id }}
         run: |
           gh workflow run "$WORKFLOW_FILE" \
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -120,7 +120,7 @@ jobs:
           OWNER: Arm-debug
           REPO: signer
           WORKFLOW_FILE: upload_archives_to_artifactory.yml
-          REF: main
+          REF: pythonize
           UPSTREAM_RUN_ID: ${{ github.run_id }}
         run: |
           gh workflow run "$WORKFLOW_FILE" \
@@ -138,7 +138,7 @@ jobs:
           OWNER: Arm-debug
           REPO: signer
           WORKFLOW_FILE: sign-and-post-macos.yml
-          REF: main
+          REF: pythonize
           UPSTREAM_RUN_ID: ${{ github.run_id }}
         run: |
           gh workflow run "$WORKFLOW_FILE" \
@@ -157,7 +157,7 @@ jobs:
           OWNER: Arm-debug
           REPO: signer
           WORKFLOW_FILE: sign-and-post-windows.yml
-          REF: main
+          REF: pythonize
           UPSTREAM_RUN_ID: ${{ github.run_id }}
         run: |
           gh workflow run "$WORKFLOW_FILE" \
--- a/internal/ssh/host.go
+++ b/internal/ssh/host.go
@@ -15,6 +15,14 @@ func (h Host) IsPlainLocalhost() bool {
 	return strings.EqualFold(string(h), "localhost") || h == "127.0.0.1"
 }
 
+func (h Host) IsLocalhost() bool {
+	if h.IsPlainLocalhost() {
+		return true
+	}
+	_, host, _ := SplitUserHostPort(string(h))
+	return Host(host).IsPlainLocalhost()
+}
+
 func (h Host) AsURI() string {
 	return fmt.Sprintf("ssh://%s", h)
 }
--- a/internal/ssh/ssh_config_resolve.go
+++ b/internal/ssh/ssh_config_resolve.go
@@ -24,7 +24,7 @@ func resolveSSHConfigHost(raw string) (string, string, string) {
 	}
 
 	values := parseSSHConfigOutput(output)
-	return values.host, values.user, values.port
+	return values.user, values.host, values.port
 }
 
 func isExplicitSSHHost(raw string) bool {
--- a/internal/ssh/ssh_config_resolve.go
+++ b/internal/ssh/ssh_config_resolve.go
@@ -1,0 +1,81 @@
+package ssh
+
+import (
+	"bufio"
+	"bytes"
+	"os/exec"
+	"strings"
+)
+
+type sshConfigValues struct {
+	host string
+	user string
+	port string
+}
+
+func resolveSSHConfigHost(raw string) string {
+	if raw == "" || isExplicitSSHHost(raw) {
+		return raw
+	}
+
+	output, err := exec.Command("ssh", "-G", raw).Output()
+	if err != nil {
+		return raw
+	}
+
+	values := parseSSHConfigOutput(output)
+	return buildResolvedHost(raw, values)
+}
+
+func isExplicitSSHHost(raw string) bool {
+	if strings.HasPrefix(raw, "ssh://") {
+		return true
+	}
+	if strings.Contains(raw, "@") || strings.Contains(raw, ":") {
+		return true
+	}
+	if strings.HasPrefix(raw, "[") {
+		return true
+	}
+	return false
+}
+
+func parseSSHConfigOutput(output []byte) sshConfigValues {
+	var values sshConfigValues
+	scanner := bufio.NewScanner(bytes.NewReader(output))
+	for scanner.Scan() {
+		fields := strings.Fields(scanner.Text())
+		if len(fields) < 2 {
+			continue
+		}
+		switch strings.ToLower(fields[0]) {
+		case "hostname":
+			values.host = fields[1]
+		case "user":
+			values.user = fields[1]
+		case "port":
+			values.port = fields[1]
+		}
+	}
+	return values
+}
+
+func buildResolvedHost(raw string, values sshConfigValues) string {
+	hostPart := values.host
+	if hostPart == "" {
+		hostPart = raw
+	}
+
+	if strings.Contains(hostPart, ":") {
+		hostPart = "[" + hostPart + "]"
+	}
+
+	if values.port != "" {
+		hostPart = hostPart + ":" + values.port
+	}
+
+	if values.user == "" {
+		return hostPart
+	}
+	return values.user + "@" + hostPart
+}
--- a/internal/ssh/ssh_config_resolve.go
+++ b/internal/ssh/ssh_config_resolve.go
@@ -13,18 +13,19 @@ type sshConfigValues struct {
 	port string
 }
 
-func resolveSSHConfigHost(raw string) string {
+func resolveSSHConfigHost(raw string) (string, string, string) {
 	if raw == "" || isExplicitSSHHost(raw) {
-		return raw
+		return SplitUserHostPort(raw)
 	}
 
 	output, err := exec.Command("ssh", "-G", raw).Output()
 	if err != nil {
-		return raw
+
+		return "", "", ""
 	}
 
 	values := parseSSHConfigOutput(output)
-	return buildResolvedHost(raw, values)
+	return values.host, values.user, values.port
 }
 
 func isExplicitSSHHost(raw string) bool {
@@ -58,24 +59,4 @@ func parseSSHConfigOutput(output []byte) sshConfigValues {
 		}
 	}
 	return values
-}
-
-func buildResolvedHost(raw string, values sshConfigValues) string {
-	hostPart := values.host
-	if hostPart == "" {
-		hostPart = raw
-	}
-
-	if strings.Contains(hostPart, ":") {
-		hostPart = "[" + hostPart + "]"
-	}
-
-	if values.port != "" {
-		hostPart = hostPart + ":" + values.port
-	}
-
-	if values.user == "" {
-		return hostPart
-	}
-	return values.user + "@" + hostPart
 }
--- a/internal/ssh/ssh_config_resolve.go
+++ b/internal/ssh/ssh_config_resolve.go
@@ -20,7 +20,6 @@ func resolveSSHConfigHost(raw string) (string, string, string) {
 
 	output, err := exec.Command("ssh", "-G", raw).Output()
 	if err != nil {
-
 		return "", "", ""
 	}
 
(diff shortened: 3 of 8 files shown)


internal/ssh/ssh_tunnel.go
```diff
@@ -106,6 +107,52 @@ func (s *SSHTunnelStart) DryRun(w io.Writer) error {
...
+		return fmt.Errorf("SSH tunnel to %s is not secure: able to access registry port without authentication", c.TargetHost)
+	} else {
+		_, _ = fmt.Fprintln(w, "Your SSH tunnel appears to be secure: unable to access registry port without authentication")
+	}
+
+	return nil
+}
+
+func (c *CheckSSHTunnelSecurity) DryRun(w io.Writer) error {
+	_, _ = fmt.Fprintln(w, strings.Join(c.Command().Args, " "))
```
tgonzalezorlandoarm: 1. Doesn't `c.Command()` return nil on `localhost`? That would break this here.

2. I also think we should return the `err` returned `by Fprintln`

internal/ssh/ssh_config_resolve.go
```diff
@@ -0,0 +1,81 @@
...
+		return raw
+	}
+
+	output, err := exec.Command("ssh", "-G", raw).Output()
+	if err != nil {
+		return raw
+	}
+
+	values := parseSSHConfigOutput(output)
+	return buildResolvedHost(raw, values)
```
tgonzalezorlandoarm: I don't understand why is `buildResolvedHost` needed?
The only time we call `resolveSSHConfigHost`, we end up calling `SplitUserHostPort`, which splits what `buildResolvedHost` has constructed

internal/ssh/host.go
```diff
@@ -15,6 +15,14 @@ func (h Host) IsPlainLocalhost() bool {
 	return strings.EqualFold(string(h), "localhost") || h == "127.0.0.1"
 }
 
+func (h Host) IsLocalhost() bool {
+	if h.IsPlainLocalhost() {
```
tgonzalezorlandoarm, line 19: Do we need the `IsPlainLocalhost` in the codebase or can we just use `IsLocalhost` all the time now? If so, should we merge them into one?

internal/ssh/ssh_tunnel.go
```diff
@@ -106,6 +107,52 @@ func (s *SSHTunnelStart) DryRun(w io.Writer) error {
 	return nil
 }
 
+type CheckSSHTunnelSecurity struct {
+	TargetHost Host
+	Port       string
+}
+
+func (c *CheckSSHTunnelSecurity) Description() string {
```
tgonzalezorlandoarm: nit: would call this `ct`

internal/ssh/ssh_tunnel.go
```diff
@@ -35,16 +35,18 @@ func formatSSHHost(raw string, user string, host string) string {
 	return user + "@" + hostPart
 }
 
-func NewSSHTunnel(targetHost Host, port string, useControlSockets bool) (operation.Operation, operation.Operation) {
+func NewSSHTunnel(targetHost Host, port string, useControlSockets bool) (operation.Operation, operation.Operation, operation.Operation) {
 	start := NewSSHTunnelStart(targetHost, port, useControlSockets)
+	securityCheck := NewCheckSSHTunnelSecurity(targetHost, port)
```
awphi, line 40: It doesn't look there's a reason to construct this in `NewSSHTunnel` - the only reason start and stop are coupled together is so stop can refer to start when not using control sockets.

Might be cleaner to hoist it out?

internal/ssh/host.go
awphi: Merging them would break registry-based deployment in the e2e tests since we deploy to localhost:port there

internal/ssh/ssh_tunnel.go
```diff
@@ -106,6 +107,52 @@ func (s *SSHTunnelStart) DryRun(w io.Writer) error {
...
+
+func NewCheckSSHTunnelSecurity(targetHost Host, port string) *CheckSSHTunnelSecurity {
+	return &CheckSSHTunnelSecurity{TargetHost: targetHost, Port: port}
+}
+
+func (c *CheckSSHTunnelSecurity) Command() *exec.Cmd {
+	if !c.TargetHost.IsLocalhost() {
+		rawHost := resolveSSHConfigHost(string(c.TargetHost))
+		_, host, _ := SplitUserHostPort(rawHost)
+		return exec.Command("curl", fmt.Sprintf("%s:%s", host, c.Port), "--max-time", "5")
```
awphi: I guess we'll need to add curl as a host machine dependency to the README and `topo health` too

internal/ssh/ssh_tunnel.go
```diff
@@ -106,6 +107,52 @@ func (s *SSHTunnelStart) DryRun(w io.Writer) error {
...
+	}
+	return nil
+}
+
+func (c *CheckSSHTunnelSecurity) Run(w io.Writer) error {
+	cmd := c.Command()
+	if cmd == nil {
+		_, _ = fmt.Fprintln(w, "SSH tunnel target is localhost, skipping security check")
+		return nil
+	}
```
awphi, line 142: I wonder if we should implement this as a conditional operation (i.e. `operation.NewConditional`) so it's hidden completely from output when deploying to localhost? 

No strong feelings since localhost deployment isn't the champion use case

internal/ssh/ssh_config_resolve.go
```diff
@@ -0,0 +1,81 @@
...
+	if strings.Contains(raw, "@") || strings.Contains(raw, ":") {
+		return true
+	}
+	if strings.HasPrefix(raw, "[") {
+		return true
+	}
+	return false
+}
+
+func parseSSHConfigOutput(output []byte) sshConfigValues {
```
awphi, line 43: Instead of parsing ssh config on the host did you consider ssh-ing into the target and checking if `0.0.0.0:$PORT` is bound? 

This has the same effect without the need for config parsing and I think we could drop the dependency on curl since if we're just sniffing ports we can use `netstat` or similar on the target

internal/ssh/ssh_tunnel.go
logan-scully-arm: 100% we should return the error here. DryRun itself is a printing operation, if we fail at printing dry run has surely failed

internal/ssh/ssh_config_resolve.go
yejseo01: Good point. Will pass the host directly

internal/ssh/ssh_config_resolve.go
yejseo01: I’m a bit hesitant to rely on ss/netstat in embedded environments in the target. Those tools aren’t always available in some embedded linux, and /proc access can be restricted. The current curl‑style check  validates the actual exposure from the network, which is the security outcome we care about, and doesn’t depend on target tooling.

internal/ssh/ssh_tunnel.go
yejseo01: I thought it's quite clean the security check operation is generated when SSH tunnel is initialized in line 43 at `internal/deploy/docker/deploy.go`. Especially when the parameters(SSH host information necessary) are the same as what's required to open the SSH tunnel. 

I can make it into another operation if you think that's not enough justification of this implementation being in here

internal/ssh/ssh_tunnel.go
yejseo01: Is this section already made? or is it something new we should make?

internal/ssh/ssh_tunnel.go
yejseo01: done in 3684ad1ed6c53832b3029845808da53c4be5cea9

internal/ssh/ssh_config_resolve.go
yejseo01: done in 8e3da148cc3873ab395175d18363286905ef38ba

internal/ssh/ssh_tunnel.go
yejseo01: I think I will decide on this based on the conclusion we have on this comment - https://github.com/arm/topo/pull/54/changes#r2864175625

I think I would move the CheckSecurity operation to `operation`, and I can implement the conditional of the localhost within there if we decide to move the check outside the SSHNewTunnel. 

Since it's not particularly necessary, will not change just for this

internal/ssh/ssh_tunnel.go
yejseo01: This is a good point, but all other dry-run operation here returns `nil` regardless of the error of the `Fprintln` in this file. Will make a different PR for fixing this across this file

yejseo01: fix for this part in particular is in 506ea62b4a69a07e866076865da8443b0b37cfb4. making a new PR for the whole file

yejseo01: new PR raised here - https://github.com/arm/topo/pull/69 will open it for review when this PR is merged

.github/workflows/release.yaml
```diff
@@ -120,7 +118,7 @@ jobs:
           OWNER: Arm-debug
           REPO: signer
           WORKFLOW_FILE: upload_archives_to_artifactory.yml
-          REF: main
+          REF: pythonize
```
th3james: These changes shouldn't be in this PR, right?
```suggestion
          REF: main
```

internal/ssh/ssh_tunnel.go
```diff
@@ -106,6 +107,57 @@ func (s *SSHTunnelStart) DryRun(w io.Writer) error {
...
+	return &CheckSSHTunnelSecurity{TargetHost: targetHost, Port: port}
+}
+
+func (ct *CheckSSHTunnelSecurity) Command() *exec.Cmd {
+	if !ct.TargetHost.IsLocalhost() {
+		_, host, _ := resolveSSHConfigHost(string(ct.TargetHost))
+		if host == "" {
+			return nil
+		}
+		return exec.Command("curl", fmt.Sprintf("%s:%s", host, ct.Port), "--max-time", "5")
```
th3james: This is going to add 5 seconds of overhead in situations where everything is secure. At minimum we should reduce this to the smallest workable number
```suggestion
		return exec.Command("curl", fmt.Sprintf("%s:%s", host, ct.Port), "--max-time", "1")
```

internal/ssh/ssh_tunnel.go
```diff
@@ -106,6 +107,57 @@ func (s *SSHTunnelStart) DryRun(w io.Writer) error {
...
+		return nil
+	}
+	cmd.Stdout = w
+	cmd.Stderr = w
+
+	err := cmd.Run()
+	if err == nil {
+		return fmt.Errorf("SSH tunnel to %s is not secure: able to access registry port without authentication", ct.TargetHost)
+	} else {
+		_, _ = fmt.Fprintln(w, "Your SSH tunnel appears to be secure: unable to access registry port without authentication")
```
th3james: I think this is also just noise
```suggestion
```

internal/ssh/ssh_tunnel.go
```diff
@@ -106,6 +107,57 @@ func (s *SSHTunnelStart) DryRun(w io.Writer) error {
...
+
+func (ct *CheckSSHTunnelSecurity) Command() *exec.Cmd {
+	if !ct.TargetHost.IsLocalhost() {
+		_, host, _ := resolveSSHConfigHost(string(ct.TargetHost))
+		if host == "" {
+			return nil
+		}
+		return exec.Command("curl", fmt.Sprintf("%s:%s", host, ct.Port), "--max-time", "5")
+	}
+	return nil
```
th3james, line 132: I would recommend moving this check one layer up the stack
```suggestion
	_, host, _ := resolveSSHConfigHost(string(ct.TargetHost))
	if host == "" {
		return nil, fmt.Errorf("failed to resolve SSH host %q", ct.TargetHost)
	}
	return exec.Command("curl", fmt.Sprintf("%s:%s", host, ct.Port), "--max-time", "5"), nil
```

internal/ssh/ssh_tunnel.go
```diff
@@ -106,6 +107,57 @@ func (s *SSHTunnelStart) DryRun(w io.Writer) error {
...
+}
+
+func (ct *CheckSSHTunnelSecurity) Run(w io.Writer) error {
+	cmd := ct.Command()
+	if cmd == nil {
+		_, _ = fmt.Fprintln(w, "SSH tunnel target is localhost, skipping security check")
+		return nil
+	}
```
th3james, line 142: * Do the localhost check first and early return
* Treat failure to resolve the host as a programmer error
```suggestion
	if ct.TargetHost.IsLocalhost() {
		return nil
	}
	cmd, err := ct.Command()
	if err != nil {
		panic(fmt.Sprintf("BUG: security check called for unresolvable host %q; caller must validate host before invoking", ct.TargetHost))
	}
```

internal/ssh/ssh_tunnel.go
```diff
@@ -106,6 +107,57 @@ func (s *SSHTunnelStart) DryRun(w io.Writer) error {
...
+
+	return nil
+}
+
+func (ct *CheckSSHTunnelSecurity) DryRun(w io.Writer) error {
+	if !ct.TargetHost.IsLocalhost() {
+		_, err := fmt.Fprintln(w, strings.Join(ct.Command().Args, " "))
+		return err
+	}
```
th3james: Invert this
```suggestion
	if ct.TargetHost.IsLocalhost() {
		return nil
	}
	cmd, err := ct.Command()
	if err != nil {
		return err
	}
	_, err = fmt.Fprintln(w, strings.Join(cmd.Args, " "))
```

th3james (changes requested):

internal/ssh/ssh_tunnel.go
th3james: Oh good point, this is probably not really acceptable as a dependency on Windows?

awphi: I think `curl` is as widely available on windows as `ssh` so probably OK - not sure if the CLI interface is different so needs some testing

internal/ssh/ssh_config_resolve.go
awphi: That's fair. I guess the complexity of parsing ssh config output is required then and adding a dependency on the host machine is more acceptable than adding one on the target. Feel free to resolve 😄

internal/ssh/ssh_tunnel.go
yejseo01: this comes down to the discussion on whether we are keeping the dry-run. I think I find it useful, so maybe we keep it like this so we can have curl?

.github/workflows/release.yaml
yejseo01: done - d88e51f20c4108f1647fdb8ed8c63dc5dab236bc

internal/ssh/ssh_tunnel.go
yejseo01: I'll check if 1 second would be fine before committing this

yejseo01: done -b98ce2b8b7e2f1c73bbe0b3293e40ea8557289de

internal/ssh/ssh_tunnel.go
yejseo01: do you think this should be in this PR? it would change the input parameters of irrelevant functions. I prefer to make a different PR for this.

yejseo01: I made a different PR here to address this - https://github.com/arm/topo/pull/78/changes

internal/ssh/ssh_tunnel.go
yejseo01: done - e5d21688da7ccc39aa858869020fd9d94d0a07b2

internal/ssh/ssh_tunnel.go
yejseo01: done - e3d532b6bce6e758e294adb292b5a0031b3eba9a

internal/ssh/ssh_tunnel.go
yejseo01: done - 1ab93f870a922352055bb237ee4cea900bfd9471

th3james (approved):

PR #55, opened 2026-02-26
chore: camelCase keys in topo health json output

awphi:
## Changes

- Does what it says on the tin. Allows us to have more standard-looking JS objects in the vscode extension + rename props on these structs without breaking the extension in the future

Why is the branch called snake-case but the PR is in camelCase? I changed my mind after pushing the branch


logan-scully-arm (approved):
LGTM!

PR #56, opened 2026-02-26
feat: Enable darwin-x64 builds

thegecko:
## Changes

- Stop not building the darwin x64 release

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -91,7 +91,9 @@ jobs:
         uses: actions/upload-artifact@v6
         with:
           name: topo-v${{ steps.version.outputs.version }}-macos-unsigned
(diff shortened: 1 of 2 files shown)


tgonzalezorlandoarm (approved):

awphi (commented):
We'll also need to pass this new binary to the signer

tgonzalezorlandoarm:
> We'll also need to pass this new binary to the signer

@yejseo01 maybe?

thegecko:
I've added this new build to the artefact uploads, looks like magic signing may work from there?

awphi (approved):
Looks right to me!

thegecko:
Hmm, the new build artefact doesn't seem to be appearing, am I missing something?

cc @yejseo01

PR #57, opened 2026-02-27
Removed vscode config metadata

federicobozzini:
## Changes

- This functionality is no longer used by the VS Code extension


awphi (approved):
🎉

PR #59, opened 2026-02-27
Use login shell for checking binaries

Luke-Parkin:
## Changes

- Fixes bug in connection resulting in binary not checking


tgonzalezorlandoarm:
I wonder if this issue should have been discovered by testing on the install side? Maybe an opportunity for an e2e test?

logan-scully-arm (approved):
LGTM! 

I agree that Topo probably needs a lot more e2e tests to prevent issues like this.

PR #60, opened 2026-02-27
Create e2e Tests for the Install Command

logan-scully-arm:
## Changes

- Add test to check regular install.
- Test for reinstall (ie new version)
- Test for failing to install to a fake target

### DIFF
--- a/e2e/install_test.go
+++ b/e2e/install_test.go
@@ -15,7 +15,9 @@ func TestInstall(t *testing.T) {
 
 	t.Run("installs the binary", func(t *testing.T) {
 		targetURL := fmt.Sprintf("ssh://%s", target.SSHConnectionString)
+
 		out, err := installRemoteproc(topo, targetURL)
+
 		require.NoError(t, err, out)
 		requireInstalled(t, "remoteproc-runtime", targetURL)
 	})
@@ -23,13 +25,16 @@ func TestInstall(t *testing.T) {
 	t.Run("replaces the binary when installing a new version", func(t *testing.T) {
 		targetURL := fmt.Sprintf("ssh://%s", target.SSHConnectionString)
 		requireInstalled(t, "remoteproc-runtime", targetURL)
+
 		out, err := installRemoteproc(topo, targetURL)
+
 		require.NoError(t, err, out)
 		requireInstalled(t, "remoteproc-runtime", targetURL)
 	})
 
 	t.Run("fails to install when the target cannot connect", func(t *testing.T) {
 		_, err := installRemoteproc(topo, "fake@target")
+
 		require.Error(t, err)
 	})
 }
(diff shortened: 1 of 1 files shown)


internal/install/install.go
```diff
@@ -134,7 +134,7 @@ func getLatestReleaseTarAddress(repoURL string) (*url.URL, error) {
 
 	base.Path = path.Join(
 		"repos",
-		url.PathEscape(repoURL),
+		repoURL,
```
awphi, line 137: Could swear we fixed this #58

e2e/install_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+func TestInstall(t *testing.T) {
+	target := testutil.StartTargetContainer(t)
+	topo := buildBinary(t)
+
+	t.Run("installs the binary", func(t *testing.T) {
+		targetURL := fmt.Sprintf("ssh://%s", target.SSHConnectionString)
+		out, err := installRemoteproc(topo, targetURL)
+		require.NoError(t, err, out)
+		requireInstalled(t, "remoteproc-runtime", targetURL)
```
awphi: nit: Could we line-delimit the three steps of AAA


```suggestion
		targetURL := fmt.Sprintf("ssh://%s", target.SSHConnectionString)
		
		out, err := installRemoteproc(topo, targetURL)
		
		require.NoError(t, err, out)
		requireInstalled(t, "remoteproc-runtime", targetURL)
```

e2e/install_test.go
```diff
@@ -0,0 +1,56 @@
...
+	})
+
+	t.Run("replaces the binary when installing a new version", func(t *testing.T) {
+		targetURL := fmt.Sprintf("ssh://%s", target.SSHConnectionString)
+		requireInstalled(t, "remoteproc-runtime", targetURL)
+		out, err := installRemoteproc(topo, targetURL)
+		require.NoError(t, err, out)
+		requireInstalled(t, "remoteproc-runtime", targetURL)
+	})
```
awphi: Two things that could be improved here:
1. This test doesn't assert that the binary is replaced at all, just that installing remoteproc a second time works
2. It relies on the previous test in this suite to run first. This is kind of down to the stateful nature of the target but I wonder if we can do better?

Would it be possible to "install" a dummy binary called `remoteproc-runtime` such that we can assert it gets replaced? If not suggest renaming the test to better reflect the assertions it's making

awphi (approved):
Nothing I'd want to block this over but a nit and a suggestion

internal/install/install.go
logan-scully-arm: Yeah, We did :+1:

e2e/install_test.go
logan-scully-arm: Yeah you're right. I'll see if I can install a fake 'remoteproc-runtime' then assert it's not the same.

If not I might just scrap the test

muchzill4:
@logan-scully-arm aargh, I thought you are no longer working on this and started to fiddle locally. 😅 Are you planning to wrap it up or shall I?

logan-scully-arm:
@muchzill4 I'm so sorry I didn't realize 😅 . I don't mind, I'm happy to finish this, but if you'd like to that's alright too :+1:

logan-scully-arm:
As discussed offline @muchzill4 will carry on with the changes to this PR

PR #61, opened 2026-02-27
Implement some basic e2e health tests

logan-scully-arm:
## Changes

- Adds some basic e2e tests for the health command.
- Validates host information
- Validates target connection
- Ensures proper handling when failing a connection

### DIFF
--- a/internal/testutil/test-container/Dockerfile
+++ b/internal/testutil/test-container/Dockerfile
@@ -1,6 +1,6 @@
 FROM docker:dind
 
-RUN apk add --no-cache openssh-server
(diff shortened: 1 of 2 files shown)


internal/testutil/test-container/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM docker:dind
 
-RUN apk add --no-cache openssh-server
+RUN apk add --no-cache openssh-server util-linux
```
awphi, line 3: Interesting - we should look into what distro this is based on since it's the first we've come across without `util-linux`

awphi (approved):
LGTM!

PR #64, opened 2026-03-01
fix:Organize signing

yejseo01:
## Changes

- Fix the tag name for pre-releases so that it's obvious it's tagged by the date
- delete the build folders after the archives are made
- decompress the final archives before sending them off to be signed
- use the decompressed archives to sign not the build folder
- use the updated workflow file names in the signer
- Update the archive name format in goreleaser
- Update the go version so that the CI wouldn't fail on govulncheck

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/arm/topo
 
-go 1.26.0
+go 1.26.1
 
 require (
 	github.com/compose-spec/compose-go/v2 v2.10.1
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/arm/topo
 
-go 1.26.0
+go 1.26.1
```
Luke-Parkin, line 3: 🤨

Luke-Parkin (approved):
Looks fine! I assume you've tested?

go.mod
muchzill4: https://github.com/arm/topo/pull/83 please don't be lazy - a change like this warrants it's own PR, which explains why this change was made.

PR #66, opened 2026-03-03
chore: Gosec cleanup

muchzill4:
## Changes

- Removes a bunch of `G204` ignores, which we exclude by `severity: high` entry in `.golangci.yaml`
- Maybe improves `#nosec G703` explanation


tgonzalezorlandoarm (approved):
WDYT about lowering the `severity` ?

muchzill4:
> WDYT about lowering the `severity` ?

Right now it'd mean having to add `# nosec` in a bunch of non-issues, so I'd rather not. 😉 
Can be convinced otherwise though, if people find lowering severity legitimately useful and will sleep better. 👍

PR #67, opened 2026-03-03
feat: filter compatible templates by target-description file

tgonzalezorlandoarm:
* Use the target-description file to show whether or not a template is compatible with the given target.

This helps users choose templates with confidence before cloning or deploying, using the same hardware understanding produced by topo describe.

## Changes

- Added `--target-description` flag to `topo templates` to load a local target profile file and annotate compatibility.
- Extended catalog repo model with Compatibility metadata and optional min_ram_kb
- Checks for SVE, NEON and remoteproc

### DIFF
--- a/cmd/topo/templates.go
+++ b/cmd/topo/templates.go
@@ -8,6 +8,7 @@ import (
 	"github.com/arm/topo/internal/describe"
 	"github.com/arm/topo/internal/output/printable"
 	"github.com/arm/topo/internal/output/templates"
+	"github.com/arm/topo/internal/target"
 	"github.com/spf13/cobra"
 )
 
@@ -44,9 +45,12 @@ var templatesCmd = &cobra.Command{
 			return fmt.Errorf("could not filter templates: %w", err)
 		}
 
-		profile, err := describe.ReadTargetDescriptionFromFile(targetDescriptionPath)
-		if err != nil {
-			return err
+		var profile *target.HardwareProfile
+		if targetDescriptionPath != "" {
+			profile, err = describe.ReadTargetDescriptionFromFile(targetDescriptionPath)
+			if err != nil {
+				return err
+			}
 		}
 
 		reposWithCompatibility := catalog.AnnotateCompatibility(profile, repos)
--- a/cmd/topo/templates.go
+++ b/cmd/topo/templates.go
@@ -5,11 +5,10 @@ import (
 	"os"
 
 	"github.com/arm/topo/internal/catalog"
+	"github.com/arm/topo/internal/describe"
 	"github.com/arm/topo/internal/output/printable"
 	"github.com/arm/topo/internal/output/templates"
-	"github.com/arm/topo/internal/target"
 	"github.com/spf13/cobra"
-	"gopkg.in/yaml.v3"
 )
 
 var (
@@ -45,21 +44,12 @@ var templatesCmd = &cobra.Command{
 			return fmt.Errorf("could not filter templates: %w", err)
 		}
 
-		reposWithCompatibility := catalog.WithCompatibility(repos)
-
-		if targetDescriptionPath != "" {
-			description, err := os.ReadFile(targetDescriptionPath)
-			if err != nil {
-				return fmt.Errorf("failed to read target description file %q: %w", targetDescriptionPath, err)
-			}
-
-			var profile target.HardwareProfile
-			if err := yaml.Unmarshal(description, &profile); err != nil {
-				return fmt.Errorf("failed to parse target description file %q: %w", targetDescriptionPath, err)
-			}
-			reposWithCompatibility = catalog.AnnotateCompatibility(profile, reposWithCompatibility)
+		profile, err := describe.ReadTargetDescriptionFromFile(targetDescriptionPath)
+		if err != nil {
+			return err
 		}
 
+		reposWithCompatibility := catalog.AnnotateCompatibility(profile, repos)
 		return printable.Print(templates.RepoCollection(reposWithCompatibility), os.Stdout, outputFormat)
 	},
 }
--- a/cmd/topo/templates.go
+++ b/cmd/topo/templates.go
@@ -7,10 +7,15 @@ import (
 	"github.com/arm/topo/internal/catalog"
 	"github.com/arm/topo/internal/output/printable"
 	"github.com/arm/topo/internal/output/templates"
+	"github.com/arm/topo/internal/target"
 	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v3"
 )
 
-var templateFilters catalog.TemplateFilters
+var (
+	templateFilters       catalog.TemplateFilters
+	targetDescriptionPath string
+)
 
 var templatesCmd = &cobra.Command{
 	Use:   "templates",
@@ -22,9 +27,12 @@ var templatesCmd = &cobra.Command{
 			return err
 		}
 
-		resolvedTarget, exists := lookupTarget(cmd)
-		if exists {
-			templateFilters.Target = resolvedTarget
+		// even if the target flag was not used, TOPO_TARGET may be set, so we check if the resolved target is non-empty
+		if targetDescriptionPath == "" {
+			resolvedTarget, exists := lookupTarget(cmd)
+			if exists && targetDescriptionPath == "" {
+				templateFilters.Target = resolvedTarget
+			}
 		}
 
 		repos, err := catalog.ParseRepos(catalog.TemplatesJSON)
@@ -36,7 +44,23 @@ var templatesCmd = &cobra.Command{
 		if err != nil {
 			return fmt.Errorf("could not filter templates: %w", err)
 		}
-		return printable.Print(templates.RepoCollection(repos), os.Stdout, outputFormat)
+
+		reposWithCompatibility := catalog.WithCompatibility(repos)
+
+		if targetDescriptionPath != "" {
+			description, err := os.ReadFile(targetDescriptionPath)
+			if err != nil {
+				return fmt.Errorf("failed to read target description file %q: %w", targetDescriptionPath, err)
+			}
+
+			var profile target.HardwareProfile
+			if err := yaml.Unmarshal(description, &profile); err != nil {
+				return fmt.Errorf("failed to parse target description file %q: %w", targetDescriptionPath, err)
+			}
+			reposWithCompatibility = catalog.AnnotateCompatibility(profile, reposWithCompatibility)
+		}
+
+		return printable.Print(templates.RepoCollection(reposWithCompatibility), os.Stdout, outputFormat)
 	},
 }
 
@@ -48,6 +72,13 @@ func init() {
 		[]string{},
 		"Only show templates that use the indicated arm feature (NEON, SVE, SME, SVE2, SME2)",
 	)
+	templatesCmd.Flags().StringVar(
+		&targetDescriptionPath,
+		"target-description",
+		"",
+		"Path to the target description file used to show template compatibility",
+	)
 	templatesCmd.MarkFlagsMutuallyExclusive("target", "feature")
+	templatesCmd.MarkFlagsMutuallyExclusive("target", "target-description")
 	rootCmd.AddCommand(templatesCmd)
 }
--- a/cmd/topo/templates.go
+++ b/cmd/topo/templates.go
@@ -30,7 +30,7 @@ var templatesCmd = &cobra.Command{
 		// even if the target flag was not used, TOPO_TARGET may be set, so we check if the resolved target is non-empty
 		if targetDescriptionPath == "" {
 			resolvedTarget, exists := lookupTarget(cmd)
-			if exists && targetDescriptionPath == "" {
+			if exists {
 				templateFilters.Target = resolvedTarget
 			}
 		}
--- a/internal/catalog/catalog.go
+++ b/internal/catalog/catalog.go
@@ -23,6 +23,7 @@ type Repo struct {
 	Name        string   `json:"name"`
 	Description string   `json:"description"`
 	Features    []string `json:"features"`
+	MinRAMKb    int64    `json:"min_ram_kb,omitempty"`
 	URL         string   `json:"url"`
 	Ref         string   `json:"ref"`
 }
--- a/internal/catalog/compatibility.go
+++ b/internal/catalog/compatibility.go
@@ -61,13 +61,8 @@ func isRepoSupported(profile target.HardwareProfile, supportedFeatures map[strin
 		}
 	}
 
-	if repo.MinRAMKb > 0 {
-		if profile.TotalMemoryKb == 0 {
-			return false
-		}
-		if profile.TotalMemoryKb < repo.MinRAMKb {
-			return false
-		}
+	if repo.MinRAMKb > 0 && profile.TotalMemoryKb < repo.MinRAMKb {
+		return false
 	}
 
 	return true
--- a/internal/catalog/compatibility.go
+++ b/internal/catalog/compatibility.go
@@ -19,14 +19,6 @@ type RepoWithCompatibility struct {
 	Compatibility CompatibilityStatus `json:"compatibility,omitempty"`
 }
 
-func withCompatibility(repos []Repo) []RepoWithCompatibility {
-	withCompatibility := make([]RepoWithCompatibility, len(repos))
-	for i, repo := range repos {
-		withCompatibility[i] = RepoWithCompatibility{Repo: repo}
-	}
-	return withCompatibility
-}
-
 func AnnotateCompatibility(profile *target.HardwareProfile, repos []Repo) []RepoWithCompatibility {
 	annotated := withCompatibility(repos)
 	if profile == nil {
@@ -42,6 +34,14 @@ func AnnotateCompatibility(profile *target.HardwareProfile, repos []Repo) []Repo
 	}
 
 	return annotated
+}
+
+func withCompatibility(repos []Repo) []RepoWithCompatibility {
+	withCompatibility := make([]RepoWithCompatibility, len(repos))
+	for i, repo := range repos {
+		withCompatibility[i] = RepoWithCompatibility{Repo: repo}
+	}
+	return withCompatibility
 }
 
 func compatibilityStatus(profile target.HardwareProfile, supportedFeatures map[string]struct{}, repo Repo) CompatibilityStatus {
--- a/internal/catalog/compatibility.go
+++ b/internal/catalog/compatibility.go
@@ -27,6 +27,19 @@ func AnnotateCompatibility(profile target.HardwareProfile, repos []RepoWithCompa
 	annotated := make([]RepoWithCompatibility, len(repos))
 	copy(annotated, repos)
 
+	supportedFeatures := extractSupportedFeatures(profile)
+
+	for r := range annotated {
+		repo := &annotated[r]
+		repo.Compatibility = &Compatibility{
+			Supported: isRepoSupported(profile, supportedFeatures, repo.Repo),
+		}
+	}
+
+	return annotated
+}
+
+func extractSupportedFeatures(profile target.HardwareProfile) map[string]struct{} {
 	supportedFeatures := map[string]struct{}{}
 	for _, proc := range profile.HostProcessor {
 		for _, feature := range proc.ExtractArmFeatures() {
@@ -37,29 +50,25 @@ func AnnotateCompatibility(profile target.HardwareProfile, repos []RepoWithCompa
 		supportedFeatures["remoteproc"] = struct{}{}
 		supportedFeatures["remoteproc-runtime"] = struct{}{}
 	}
+	return supportedFeatures
+}
 
-	for r := range annotated {
-		repo := &annotated[r]
-		supported := true
-
-		for _, feature := range repo.Features {
-			normalized := strings.ToLower(strings.TrimSpace(feature))
-			if _, ok := supportedFeatures[normalized]; !ok {
-				supported = false
-				break
-			}
+func isRepoSupported(profile target.HardwareProfile, supportedFeatures map[string]struct{}, repo Repo) bool {
+	for _, feature := range repo.Features {
+		normalized := strings.ToLower(strings.TrimSpace(feature))
+		if _, ok := supportedFeatures[normalized]; !ok {
+			return false
 		}
-
-		if repo.MinRAMKb > 0 {
-			if profile.TotalMemoryKb == 0 {
-				supported = false
-			} else if profile.TotalMemoryKb < repo.MinRAMKb {
-				supported = false
-			}
-		}
-
-		repo.Compatibility = &Compatibility{Supported: supported}
 	}
 
-	return annotated
+	if repo.MinRAMKb > 0 {
+		if profile.TotalMemoryKb == 0 {
+			return false
+		}
+		if profile.TotalMemoryKb < repo.MinRAMKb {
+			return false
+		}
+	}
+
+	return true
 }
--- a/internal/catalog/compatibility.go
+++ b/internal/catalog/compatibility.go
@@ -1,0 +1,65 @@
+package catalog
+
+import (
+	"strings"
+
+	"github.com/arm/topo/internal/target"
+)
+
+type Compatibility struct {
+	Supported bool `json:"supported"`
+}
+
+type RepoWithCompatibility struct {
+	Repo
+	Compatibility *Compatibility `json:"compatibility,omitempty"`
+}
+
+func WithCompatibility(repos []Repo) []RepoWithCompatibility {
+	withCompatibility := make([]RepoWithCompatibility, len(repos))
+	for i, repo := range repos {
+		withCompatibility[i] = RepoWithCompatibility{Repo: repo}
+	}
+	return withCompatibility
+}
+
+func AnnotateCompatibility(profile target.HardwareProfile, repos []RepoWithCompatibility) []RepoWithCompatibility {
+	annotated := make([]RepoWithCompatibility, len(repos))
+	copy(annotated, repos)
+
+	supportedFeatures := map[string]struct{}{}
+	for _, proc := range profile.HostProcessor {
+		for _, feature := range proc.ExtractArmFeatures() {
+			supportedFeatures[strings.ToLower(feature)] = struct{}{}
+		}
+	}
+	if len(profile.RemoteCPU) > 0 {
+		supportedFeatures["remoteproc"] = struct{}{}
+		supportedFeatures["remoteproc-runtime"] = struct{}{}
+	}
+
+	for r := range annotated {
+		repo := &annotated[r]
+		supported := true
+
+		for _, feature := range repo.Features {
+			normalized := strings.ToLower(strings.TrimSpace(feature))
+			if _, ok := supportedFeatures[normalized]; !ok {
+				supported = false
+				break
+			}
+		}
+
+		if repo.MinRAMKb > 0 {
+			if profile.TotalMemoryKb == 0 {
+				supported = false
+			} else if profile.TotalMemoryKb < repo.MinRAMKb {
+				supported = false
+			}
+		}
+
+		repo.Compatibility = &Compatibility{Supported: supported}
+	}
+
+	return annotated
+}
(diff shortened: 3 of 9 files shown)


tgonzalezorlandoarm:
Question for this PR:
Do we want to have these 2 functionalities?
1. Template `Filtering` via `feature`
2. Template `Compatibility` via `target-description`

~~I need to do some refactoring to leverage on the common logic, this will happen soon~~

tgonzalezorlandoarm:
> Question for this PR: Do we want to have these 2 functionalities?
> 
> 1. Template `Filtering` via `feature`
> 2. Template `Compatibility` via `target-description`
> 
> ~I need to do some refactoring to leverage on the common logic, this will happen soon~

As discussed offline, we will remove the `filtering` feature in a follow-up PR. The reason behind this is that there are a lot of ways to grep through a list so the functionality may not be as useful as compatibility reporting

internal/output/templates/templates.go
```diff
@@ -4,23 +4,46 @@ import (
...
+	}
+	return "❌"
+}
+
+func colorCompatibilityMark(c *catalog.Compatibility) string {
+	if c == nil {
+		return ""
+	}
+	if c.Supported {
+		return term.Color(term.Cyan, plainCompatibilityMark(c))
```
muchzill4: Are we colouring emojis? I'm a bit confused. 🤔

internal/output/templates/templates_list_test.go
```diff
@@ -197,4 +197,42 @@ name-of-other-project | url.git | main
 
 		assert.Equal(t, want, got)
 	})
+
+	t.Run("prints compatibility markers when provided", func(t *testing.T) {
```
muchzill4: Test description could use improvement: What is provided? Compatibility markers?

internal/catalog/compatibility.go
```diff
@@ -0,0 +1,54 @@
...
+			}
+		}
+
+		if repo.MinRAMKb > 0 {
+			if profile.TotalMemoryKb == 0 {
+				supported = false
+			} else if profile.TotalMemoryKb < repo.MinRAMKb {
+				supported = false
+			}
+		}
```
muchzill4: Should this be part of this PR? Where is memory from?

internal/catalog/catalog.go
```diff
@@ -20,11 +20,13 @@ type TemplateFilters struct {
...
-	Features    []string `json:"features"`
-	URL         string   `json:"url"`
-	Ref         string   `json:"ref"`
+	Name          string         `json:"name"`
+	Description   string         `json:"description"`
+	Features      []string       `json:"features"`
+	MinRAMKb      int64          `json:"min_ram_kb,omitempty"`
+	URL           string         `json:"url"`
+	Ref           string         `json:"ref"`
+	Compatibility *Compatibility `json:"compatibility,omitempty"`
```
muchzill4: I don't think `Compatibility` is a property of a repo in the catalog, thus it shouldn't be modelled on that type. Repo in the catalog is the raw metadata of the repo. The compatibility is an add-on feature, which should be modelled separately.

Idea 1: Can we use a simple slice to group compatible vs incompatible repos? 
Idea 2: Add a new type `RepoWithCompatibility` or such, that has this new data.

This of course means changing `templates.RepoCollection`.

internal/output/templates/templates_list_test.go
```diff
@@ -197,4 +197,42 @@ name-of-other-project | url.git | main
...
+		for _, tc := range []struct {
+			name      string
+			supported bool
+			want      string
+		}{
+			{
+				name:      "prints compatibility marker when provided",
+				supported: true,
+				want:      "✅ name-of-project | url.git | main\n\n",
```
muchzill4: I'm not a fan of the fact the `want` precedes the input data, which is hidden in a loop. It's kinda backwards.

I think that maybe this is complex enough to warrant not using a table test? Also, there's a bunch of data, like "url.git" and "main" that's not really relevant here - it's just noise that makes my tired brain consume more precious thought tokens.

Perhaps we could do something like this:

```go
t.Run("prints compatibility marker if project is compatible and vice versa", func() {
   compatibleRepo := Repo{Name: "lasagne", Compatible: true}
   incompatibleRepo := Repo{Name: "spaghetti", Compatible: false}
   repos := []catalog.Repo{compatibleRepo, incompatibleRepo}

   // printing shenanigans

   assert.Contains("✅ compatible repo")
   assert.Contains("❌ incompatible repo")
})
```

internal/catalog/catalog.go
muchzill4: Having a second look at this, we already have `FilterTemplateRepos` which works just like I'd expect for filtering compatible repos from a collection of repos (in essence producing two collections of repos). I think we should use similar pattern here. I'm of course assuming that we don't need to show which property of the repo makes it compatible.

If in near future we'll need to say which property of the repo makes it compatible and which doesn't, then we definitely need a new type. The representation of raw repo itself shouldn't know about the compatibility layer.

internal/output/templates/templates_list_test.go
tgonzalezorlandoarm: https://github.com/arm/topo/pull/67/commits/deff06925ee7829b21dcfbc4c62382e0b51d88e2

tgonzalezorlandoarm: https://github.com/arm/topo/pull/67/commits/31cae20db2db8ddabb12c9ee9e6e4b5de27c1148

internal/output/templates/templates.go
tgonzalezorlandoarm: Ooppsiiie

internal/catalog/compatibility.go
tgonzalezorlandoarm: The target-description.yaml for the DXG, for example, has:

```
(...)
        - dgh
        - bti
        - ecv
        - afp
        - wfxt
remoteprocs: []
totalmemory_kb: 125487760
```

So with this we add capability of describing a mininum RAM required for the project by filtering the target description file, which is something I understood we want.
I can leave it for another PR if you want?

tgonzalezorlandoarm: Basically, here, if you edit the `templates.json` for any project, like the `simd-visual-benchmark`, to for example:

```
  {
    "name": "simd-visual-benchmark",
    "description": "Visual demonstration of SIMD performance benefits on Arm processors.\nCompare scalar (no SIMD), NEON (128-bit), and SVE (scalable vector)\nimplementations running identical image processing workloads side-by-side.\n\nThis demo shows real hardware acceleration through three C++ services\ncompiled with different architecture flags, processing the same box blur\nalgorithm on images. Performance differences are measured in real-time\nand displayed in an interactive web dashboard.\n\nPerfect for demonstrating to non-technical audiences the concrete benefits\nof SIMD optimizations, with visual results and quantified speedups.\n",
    "features": [
      "NEON",
      "SVE"
    ],
    "url": "git@github.com:Arm-Examples/topo-simd-visual-benchmark.git",
    "ref": "main",
    "min_ram_kb": 125487761
  }
```

then the project will be marked as incompatible.

tgonzalezorlandoarm: (for the DGX)

muchzill4: I see, thanks for clarifying. I'd probably try to do this separately, because smaller PR gets reviewed faster and integrated quickly. However it's already here, so perhaps ok to leave it.

internal/catalog/compatibility.go
```diff
@@ -0,0 +1,54 @@
...
+	"strings"
+
+	"github.com/arm/topo/internal/target"
+)
+
+type Compatibility struct {
+	Supported bool `json:"supported"`
+}
+
+func AnnotateCompatibility(profile target.HardwareProfile, repos []Repo) []Repo {
```
muchzill4: This function does a lot. Can we try to extract named pieces of functionality, so that it is easier to digest for a human reader? These could be simple private functions.

`extractSupportedFeatures()`
`areRepoFeaturesCompatible(repo.Features, extractSupportedFeatures(profile))`
`isMemoryRequirementFulfilled(repo.MinRamKb, profile.TotalMemoryKb)`

What I'm aiming to achieve is:
```go
func DoAllThings() {
   supportedFeatures := extractSupportedFeatures()
   for _, repo := range repos {
       if areRepoFeaturesCompatible() && isMemoryRequirementFulfilled() {
           append...
       }
   }
}
```

internal/catalog/catalog.go
tgonzalezorlandoarm: Working on the changes.
Btw, we are no longer going to be filtering templates (and the `--feature` flag is going to go), we are just going to be showing compatibility

internal/catalog/compatibility.go
tgonzalezorlandoarm: https://github.com/arm/topo/pull/67/commits/e84872a155bca5744583331d98bd01e0e274fce0

tgonzalezorlandoarm:
@muchzill4 hopefully I've resolved everything now! LMK. Thanks a lot for the reviews so far!

internal/catalog/catalog.go
tgonzalezorlandoarm: I now added the RepoWithCompatibility type.
FilterTemplateRepos is going to go as the filtering functionality is no longer needed as far as I'm aware (will remove this in a separate PR).

I personally don't like having two list of repos for this, so I went with Idea 2.

internal/catalog/compatibility.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/arm/topo/internal/target"
+)
+
+type Compatibility struct {
+	Supported bool `json:"supported"`
+}
+
+type RepoWithCompatibility struct {
+	Repo
+	Compatibility *Compatibility `json:"compatibility,omitempty"`
```
muchzill4: Do I understand correctly that this is a pointer to structure to represent three states:
- Was checked and is compatible
- Was checked and is not compatible
- Wasn't checked

If so, we should probably use an enum, to make it clearer.

internal/catalog/compatibility.go
```diff
@@ -0,0 +1,74 @@
...
+
+func WithCompatibility(repos []Repo) []RepoWithCompatibility {
+	withCompatibility := make([]RepoWithCompatibility, len(repos))
+	for i, repo := range repos {
+		withCompatibility[i] = RepoWithCompatibility{Repo: repo}
+	}
+	return withCompatibility
+}
+
+func AnnotateCompatibility(profile target.HardwareProfile, repos []RepoWithCompatibility) []RepoWithCompatibility {
```
muchzill4: Why does this take `[]RepoWithCompatibility` and not `[]Repo`?

internal/output/templates/templates_list.go
```diff
@@ -24,7 +24,7 @@ const repoTemplate = `
 {{- end }}
 
 {{- define "repoRow" }}
-{{- cyan .Name }} | {{ blue .URL }} | {{ yellow .Ref }}
+{{- if .Compatibility }}{{ compatibilityMark .Compatibility }} {{ end }}{{ cyan .Name }} | {{ blue .URL }} | {{ yellow .Ref }}
```
muchzill4, line 27: Another reason to do boolean here. `if .Compatible` would read better. 😉

cmd/topo/templates.go
```diff
@@ -22,9 +27,12 @@ var templatesCmd = &cobra.Command{
...
 
-		resolvedTarget, exists := lookupTarget(cmd)
-		if exists {
-			templateFilters.Target = resolvedTarget
+		// even if the target flag was not used, TOPO_TARGET may be set, so we check if the resolved target is non-empty
+		if targetDescriptionPath == "" {
+			resolvedTarget, exists := lookupTarget(cmd)
+			if exists && targetDescriptionPath == "" {
+				templateFilters.Target = resolvedTarget
+			}
```
muchzill4, line 35: What does this do exactly, I'm a bit confused. Why is this required now?

Also, we check `targetDescriptionPath == ""` twice.

cmd/topo/templates.go
```diff
@@ -36,7 +44,23 @@ var templatesCmd = &cobra.Command{
...
+			if err != nil {
+				return fmt.Errorf("failed to read target description file %q: %w", targetDescriptionPath, err)
+			}
+
+			var profile target.HardwareProfile
+			if err := yaml.Unmarshal(description, &profile); err != nil {
+				return fmt.Errorf("failed to parse target description file %q: %w", targetDescriptionPath, err)
+			}
+			reposWithCompatibility = catalog.AnnotateCompatibility(profile, reposWithCompatibility)
+		}
```
muchzill4, line 54: I wonder if we could call `catalog.AnnotateCompatibility` unconditionally, sometimes with `nil` profile, thus making the flow a bit simpler. This is a tradeoff, which makes that function a bit more complex, but makes call sites nicer. `nil` profile is useful, unless we introduce a method on `HardwareProfile` that can determine if it's "empty".

```go
profile := buildHardwareProfile(targetDescriptionPath)
reposWithCompatibility := catalog.AnnotateCompatibility(profile, repos)
return printable.Print(...)
```

Where:

```go
func buildHardwareProfile(targetDescriptionPath string) (*profile, error) {
    if targetDescriptionPath == "" {
       return nil, nil
    }
    ...
}
```

internal/catalog/compatibility.go
```diff
@@ -0,0 +1,74 @@
...
+		normalized := strings.ToLower(strings.TrimSpace(feature))
+		if _, ok := supportedFeatures[normalized]; !ok {
+			return false
+		}
+	}
+
+	if repo.MinRAMKb > 0 {
+		if profile.TotalMemoryKb == 0 {
+			return false
+		}
```
muchzill4: Is this necessary given below condition will also be true in that case?

cmd/topo/templates.go
tgonzalezorlandoarm: At the moment, we don't want to be  marking compatibility of templates by using the `target` flag yet.
Problem is that `target` may be exposed either by the `target` flag or by the `TOPO_TARGET` flag.

So just by doing:
`templatesCmd.MarkFlagsMutuallyExclusive("target", "target-description")` we don't actually guarantee the mutual exclusion of target and the rest.

internal/catalog/compatibility.go
tgonzalezorlandoarm: good catch: https://github.com/arm/topo/pull/67/commits/5cbaaa463b93f4578d52e8759654307c0ae51d3d

cmd/topo/templates.go
tgonzalezorlandoarm: https://github.com/arm/topo/pull/67/commits/7edc7bf42da9a74d522dd01d609bff120b122e50

cmd/topo/templates.go
tgonzalezorlandoarm: This makes a lot of sense,, and I should probably handle this in the `describe` package as well!
Done: https://github.com/arm/topo/pull/67/commits/0b22177badd8241f794ade8841a9bb4132d8ecd7

tgonzalezorlandoarm: plus tests, of course: https://github.com/arm/topo/pull/67/commits/df496a106b8403e5ea9394487bfd69d9a2a1379e

internal/catalog/compatibility.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+func AnnotateCompatibility(profile *target.HardwareProfile, repos []Repo) []RepoWithCompatibility {
+	annotated := withCompatibility(repos)
+	if profile == nil {
+		return annotated
+	}
+
+	hardwareProfile := *profile
+	supportedFeatures := extractSupportedFeatures(hardwareProfile)
```
muchzill4, line 28:
```suggestion
	supportedFeatures := extractSupportedFeatures(*profile)
```

Is that worse or better? 🤔

internal/catalog/compatibility.go
```diff
@@ -0,0 +1,81 @@
...
+
+func withCompatibility(repos []Repo) []RepoWithCompatibility {
+	withCompatibility := make([]RepoWithCompatibility, len(repos))
+	for i, repo := range repos {
+		withCompatibility[i] = RepoWithCompatibility{Repo: repo}
+	}
+	return withCompatibility
+}
+
+func AnnotateCompatibility(profile *target.HardwareProfile, repos []Repo) []RepoWithCompatibility {
```
muchzill4, line 22: Nit: public functions should go first in the file.

internal/catalog/compatibility_test.go
```diff
@@ -0,0 +1,109 @@
...
+	t.Run("marks template unsupported when SVE is missing", func(t *testing.T) {
+		repos := []catalog.Repo{{Name: "sve-template", Features: []string{"SVE"}}}
+		profile := target.HardwareProfile{
+			HostProcessor: []target.HostProcessor{
+				{Features: []string{"asimd"}},
+			},
+		}
+
+		got := catalog.AnnotateCompatibility(&profile, repos)
+		assert.Equal(t, catalog.CompatibilityUnsupported, got[0].Compatibility)
```
muchzill4:
```suggestion
		got := catalog.AnnotateCompatibility(&profile, repos)
		
		assert.Equal(t, catalog.CompatibilityUnsupported, got[0].Compatibility)
```

NIT: Arrange Act Assert. Same comment about other tests, where `A A A` format is not respected.

internal/catalog/compatibility_test.go
```diff
@@ -0,0 +1,109 @@
...
+import (
+	"testing"
+
+	"github.com/arm/topo/internal/catalog"
+	"github.com/arm/topo/internal/target"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestAnnotateCompatibility(t *testing.T) {
+	t.Run("supports template requiring SVE when target has SVE", func(t *testing.T) {
```
muchzill4, line 12: Template or repo? 🤔

internal/catalog/compatibility_test.go
```diff
@@ -0,0 +1,109 @@
...
+	t.Run("supports template with no requirements and does not mutate input", func(t *testing.T) {
+		repos := []catalog.Repo{
+			{Name: "plain"},
+		}
+		profile := target.HardwareProfile{}
+
+		got := catalog.AnnotateCompatibility(&profile, repos)
+		assert.Len(t, got, 1)
+		assert.Equal(t, catalog.CompatibilitySupported, got[0].Compatibility)
+		assert.Equal(t, "plain", repos[0].Name)
```
muchzill4: You could simplify assertions and avoid magic numbers if tests were like this:

```suggestion
        repo := {Name: "plain"}
		repos := []catalog.Repo{repo}
		profile := target.HardwareProfile{}

		got := catalog.AnnotateCompatibility(&profile, repos)
		
		want := []catalog.RepoWithCompatibility{
		   {repo: repo, Compatibility: catalog.CompatibilitySupported}
		}
		assert.Equal(t, want, got)
```

internal/describe/describe.go
```diff
@@ -33,3 +34,20 @@ func WriteTargetDescriptionToFile(dir string, report target.HardwareProfile) (st
 	}
 	return outputFile, f.Close()
 }
+
+func ReadTargetDescriptionFromFile(filePath string) (*target.HardwareProfile, error) {
+	if filePath == "" {
+		return nil, nil
+	}
```
muchzill4: Unsure if this belongs here. I'd think this function _implicitly_ only accepts valid file paths, wdyt?

internal/describe/describe.go
```diff
@@ -33,3 +34,20 @@ func WriteTargetDescriptionToFile(dir string, report target.HardwareProfile) (st
...
+	description, err := os.ReadFile(filePath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read target description file %q: %w", filePath, err)
+	}
+
+	var profile target.HardwareProfile
+	if err := yaml.Unmarshal(description, &profile); err != nil {
+		return nil, fmt.Errorf("failed to parse target description file %q: %w", filePath, err)
+	}
+	return &profile, nil
```
muchzill4:
```suggestion
	var profile *target.HardwareProfile
	if err := yaml.Unmarshal(description, profile); err != nil {
		return nil, fmt.Errorf("failed to parse target description file %q: %w", filePath, err)
	}
	return profile, nil
```

Less pretzels. 🤷

internal/describe/describe_test.go
```diff
@@ -114,3 +114,61 @@ func TestWriteTargetDescriptionFile(t *testing.T) {
...
+    cores: 2
+remoteprocs:
+  - name: remoteproc0
+totalmemory_kb: 16384
+`
+		require.NoError(t, os.WriteFile(filePath, []byte(content), 0o644))
+
+		profile, err := describe.ReadTargetDescriptionFromFile(filePath)
+		require.NoError(t, err)
+		require.NotNil(t, profile)
```
muchzill4: If your `assert.Equal` prepended `want` value with `&` you wouldn't need this `NotNil` check.

internal/describe/describe_test.go
```diff
@@ -114,3 +114,61 @@ func TestWriteTargetDescriptionFile(t *testing.T) {
...
+			},
+			RemoteCPU:     []target.RemoteprocCPU{{Name: "remoteproc0"}},
+			TotalMemoryKb: 16384,
+		}, *profile)
+	})
+
+	t.Run("returns error when target description file does not exist", func(t *testing.T) {
+		profile, err := describe.ReadTargetDescriptionFromFile("/no/such/file.yaml")
+		require.Error(t, err)
+		assert.Nil(t, profile)
```
muchzill4: 99.9% of time we don't care about first return value if error is not nil, thus I think we can skip this assertion.

internal/describe/describe_test.go
```diff
@@ -114,3 +114,61 @@ func TestWriteTargetDescriptionFile(t *testing.T) {
...
+	})
+
+	t.Run("returns error when target description yaml is invalid", func(t *testing.T) {
+		dir := t.TempDir()
+		filePath := dir + "/invalid.yaml"
+		require.NoError(t, os.WriteFile(filePath, []byte("host_processor: ["), 0o644))
+
+		profile, err := describe.ReadTargetDescriptionFromFile(filePath)
+		require.Error(t, err)
+		assert.Nil(t, profile)
```
muchzill4:
```suggestion
```

`assert.ErrorContains` is enough.

muchzill4 (commented):
Thanks for making the changes.

Couple of nits about AAA and structure, but I think we're good to go.

internal/catalog/compatibility.go
```diff
@@ -0,0 +1,81 @@
...
+	annotated := withCompatibility(repos)
```
muchzill4: Just noticed that we're always looping twice. I think we could avoid doing `withCompatibility` if `profile != nil`.

tgonzalezorlandoarm: I don't think we can avoid this if we do the profile check inside AnnotateCompatibility, which always returns RepoWithCompatibility.

The other option is having an `if/else` situation in `cmd/templates` which is what we had before through checking the target Description path. We can do it by checking profile instead, but it ends up being hte same as before after all

tgonzalezorlandoarm: Let's revisit this once we've removed the template filtering

internal/catalog/compatibility.go
tgonzalezorlandoarm: Used the enum

internal/catalog/compatibility.go
tgonzalezorlandoarm: Done

internal/catalog/compatibility.go
tgonzalezorlandoarm: That's worse since we also have to pass it to `compatibilityStatus`

internal/catalog/compatibility.go
muchzill4: What I mean (pseudocode)

```
if profile == nil {
   return withUnknownCompatibility(repos)
}

checked := make([]RepoWithCompatibility, len(repos))
for i, repo := range repos {
   checked[i].Repo = Repo
   checked[i].Compatibility = checkCompatibility(...) 
}
```

Note, that we avoid calling `withUnknownCompatiblity` if profile is not nil, thus we don't loop twice in that case.

internal/catalog/compatibility.go
muchzill4: Right, but can't we dereference there too or do you think it's expensive?

tgonzalezorlandoarm: I guess I don't see what the advantage is of doing `*profile` everytime? I find it more aesthetically pleasant without the *profile every time, but I guess that's just preference. Any particular reason why that would be better?

internal/catalog/compatibility.go
tgonzalezorlandoarm: https://github.com/arm/topo/pull/67/commits/474f6d2e945f7e2f3c59703dae226690faafc9ea

internal/catalog/compatibility.go
muchzill4: I'm not arguing for either, just discussing. :D On one hand you introduce a variable, that's simply a dereferenced value, one could argue you can dereference inline. Both are unpleasant.

internal/describe/describe.go
tgonzalezorlandoarm: https://github.com/arm/topo/pull/67/commits/da9e3f1dbf1f147641e163eed068449a906205ef

internal/catalog/compatibility_test.go
tgonzalezorlandoarm: Good question. Is that structure correctly named?

```
type Repo struct {
	Name        string   `json:"name"`
	Description string   `json:"description"`
	Features    []string `json:"features"`
	MinRAMKb    int64    `json:"min_ram_kb,omitempty"`
	URL         string   `json:"url"`
	Ref         string   `json:"ref"`
}
```

muchzill4: Mmm. Can of worms I see. We could say "supports template repo requiring SVE ...", though this is a NIT. Fine as is. 👍

internal/catalog/compatibility.go
tgonzalezorlandoarm: Agreed

internal/catalog/compatibility_test.go
tgonzalezorlandoarm: The suggestion seems to be the same as before but with a newline in the middle?

muchzill4: Yes, it's about visually spacing your tests using [Arrange Act Assert pattern](https://thoughtbot.com/blog/the-arrange-act-assert-pattern).

internal/catalog/compatibility_test.go
tgonzalezorlandoarm: Done:
 https://github.com/arm/topo/pull/67/commits/0bbac3e4ec0ffd80349cff6bea6843d091b2de38

internal/describe/describe_test.go
tgonzalezorlandoarm: https://github.com/arm/topo/pull/67/commits/6b0ed91c61a3c3526c6a113f05133a4e7a6d1726

internal/describe/describe.go
tgonzalezorlandoarm: https://github.com/arm/topo/pull/67/commits/253a0c59ef276af2217bb0ba74953030c66dd457

internal/describe/describe_test.go
tgonzalezorlandoarm: done

internal/catalog/compatibility.go
tgonzalezorlandoarm: https://github.com/arm/topo/pull/67/commits/519880f16cea9c4c04005184fd90ea76895fa0d5

internal/describe/describe.go
tgonzalezorlandoarm: Less pretzels is not possible after all https://github.com/arm/topo/pull/67/commits/d6ddc0e6345a6ca796f267449cd05d6f2ac6e016

muchzill4: I guess it's good, because who doesn't like pretzels?

muchzill4 (approved):
🚢 🇮🇹

PR #68, opened 2026-03-04
chore: Rename `install remoteproc` to `install remoteproc-runtime`

muchzill4:
## Changes

- Makes `install` command less confusing - we're not installing remoteproc, but installing the runtime.
- Moves the `remotproc-runtime` subcommand logic to a separate file, which is consistent with other sub-commands (`service remove` lives in `service_remove.go`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,12 +138,12 @@ topo stop --target my-board
 
 The Getting Started walkthrough above covers the core flow. These additional commands are available:
 
-| Command              | When to use it                                                                                              |
-| -------------------- | ----------------------------------------------------------------------------------------------------------- |
-| `init`               | Scaffold a new empty project instead of cloning a template                                                  |
-| `extend`             | Add services from a template into an existing project                                                       |
-| `service remove`     | Remove a service from your compose file                                                                     |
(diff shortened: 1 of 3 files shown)


tgonzalezorlandoarm (approved):
🚀

tgonzalezorlandoarm:
In the health report, do we want to report on  `remoteproc` or `remoteproc-runtime`?

muchzill4:
> In the health report, do we want to report on `remoteproc` or `remoteproc-runtime`?

Can you point me towards the one you're thinking about? There's `remoteproc` as a capability of the system, which we probably should keep as is. Dependencies already mention `remoteproc-runtime`.

tgonzalezorlandoarm:
> > In the health report, do we want to report on `remoteproc` or `remoteproc-runtime`?
> 
> Can you point me towards the one you're thinking about? There's `remoteproc` as a capability of the system, which we probably should keep as is. Dependencies already mention `remoteproc-runtime`.

You are right! This LGTM

tgonzalezorlandoarm (approved):
🚀

PR #69, opened 2026-03-04
fix:dry run returns errors properly

yejseo01:
## Changes

- fix the dry run function so that it returns the error when print failed


Luke-Parkin:
Don't worry about dry run; we're deprecating it soon.

PR #71, opened 2026-03-04
refactor: Explicitly shared operation interfaces

muchzill4:
## Changes

`package operation`, originally defined for docker deployment, has also been leveraged outside of that realm. This PR pulls it from `internal/deploy/docker/operation` to  `internal/operation` so it is shared more explicitly.

Where both `internal/operation` and `internal/deploy/docker/operation` are imported in the same file, the global package is aliased as `goperation` (`g` = global).


tgonzalezorlandoarm (approved):
🚀

PR #75, opened 2026-03-05
feat: setup-keys - support selecting SSH key type

tgonzalezorlandoarm:
Not all targets support ed25519, but this key type should be the default one for modern systems.
However, we do need to offer support for setting up keys in older systems.

 * Add key-type selection to the setup-keys flow. Allow the user to use RSA as their key type.

### DIFF
--- a/internal/setupkeys/setup_keys.go
+++ b/internal/setupkeys/setup_keys.go
@@ -47,6 +47,6 @@ func ParseKeyType(s string) (KeyType, error) {
 	case KeyTypeRSA:
 		return KeyTypeRSA, nil
 	default:
-		return "", fmt.Errorf("unsupported key type %q, supported types: ed25519, rsa", s)
+		return "", fmt.Errorf("unsupported key type %q, supported types: %s, %s", s, KeyTypeED25519, KeyTypeRSA)
 	}
 }
--- a/internal/setupkeys/setup_keys.go
+++ b/internal/setupkeys/setup_keys.go
@@ -18,13 +18,8 @@ const (
 )
 
 func NewKeySetup(target string, privKeyPath string, keyType string) (operation.Sequence, error) {
-	parsedKeyType, err := ParseKeyType(keyType)
-	if err != nil {
-		return nil, err
-	}
-
 	ops := []operation.Operation{
-		sshkeygen.NewSSHKeyGen("Generate SSH key pair for target", target, string(parsedKeyType), privKeyPath, sshkeygen.SSHKeyGenOptions{}),
+		sshkeygen.NewSSHKeyGen("Generate SSH key pair for target", target, keyType, privKeyPath, sshkeygen.SSHKeyGenOptions{}),
 		pubkeytransfer.NewPubKeyTransfer("Transfer public key to target and set it as an authorized key", target, privKeyPath, pubkeytransfer.PubKeyTransferOptions{}),
 	}
 	return operation.NewSequence(ops...), nil
--- a/internal/setupkeys/setup_keys.go
+++ b/internal/setupkeys/setup_keys.go
@@ -10,9 +10,14 @@ import (
 	"github.com/arm/topo/internal/setupkeys/sshkeygen"
 )
 
+const (
+	KeyTypeED25519 string = "ed25519"
+	KeyTypeRSA     string = "rsa"
+)
+
 func NewKeySetup(target string, privKeyPath string, keyType string) (operation.Sequence, error) {
-	if keyType != "ed25519" && keyType != "rsa" {
-		return nil, fmt.Errorf("unsupported key type: %s", keyType)
+	if err := isValidKeyType(keyType); err != nil {
+		return nil, err
 	}
 
 	ops := []operation.Operation{
@@ -30,4 +35,13 @@ func GetDefaultPrivateKeyPath(targetSlug string) (string, error) {
 	keyName := fmt.Sprintf("id_ed25519_topo_%s", targetSlug)
 	privKeyPath := filepath.Join(home, ".ssh", keyName)
 	return privKeyPath, nil
+}
+
+func isValidKeyType(s string) error {
+	switch s {
+	case KeyTypeED25519, KeyTypeRSA:
+		return nil
+	default:
+		return fmt.Errorf("unsupported key type %q, supported types: %s, %s", s, KeyTypeED25519, KeyTypeRSA)
+	}
 }
--- a/internal/setupkeys/setup_keys.go
+++ b/internal/setupkeys/setup_keys.go
@@ -17,9 +17,9 @@ const (
 	KeyTypeRSA     KeyType = "rsa"
 )
 
-func NewKeySetup(target string, privKeyPath string, keyType string) (operation.Sequence, error) {
+func NewKeySetup(target string, privKeyPath string, keyType KeyType) (operation.Sequence, error) {
 	ops := []operation.Operation{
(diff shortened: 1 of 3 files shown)


internal/setupkeys/setup_keys.go
```diff
@@ -10,9 +10,13 @@ import (
 	"github.com/arm/topo/internal/setupkeys/sshkeygen"
 )
 
-func NewKeySetup(target string, privKeyPath string) (operation.Sequence, error) {
+func NewKeySetup(target string, privKeyPath string, keyType string) (operation.Sequence, error) {
```
muchzill4: We could tighten the signature if `keyType` was an actual type.

```go
type KeyType string

const (
    KeyTypeED25519 KeyType = "ed25519"
    KeyTypeRSA     KeyType = "rsa"
)

func ParseKeyType(s string) (KeyType, error) {
    switch KeyType(s) {
    case KeyTypeED25519:
        return KeyTypeED25519, nil
    case KeyTypeRSA:
        return KeyTypeRSA, nil
    default:
        return "", fmt.Errorf("unsupported key type %q, supported types: ed25519, rsa", s)
    }
}
```

This establishes a pattern where `ParseKeyType` is the single entrypoint for unvalidated input, which is also a nice separation of concerns. `NewKeySetup` doesn't deal with validation, it deals only with the operation setup.

internal/setupkeys/setup_keys_test.go
muchzill4, line 77: Perhaps not a subject of this PR, but I want to point out that the responsibilities of this test are way too much, making it fragile and hard to understand. The func name says it's `TestDryRun`, but in reality it tests a bunch of things that are outside of scope of `DryRun`, like operation types and their descriptions.

Ideally, the functions returning sequences, would be tested by either executing `Run()` on returned sequence and asserting the state of the system. Failing that - if it's not convenient - we could compare returned sequence with an expected sequence. Less ideal, because test ends up duplicating the implementation, but can still serve as warning light.

For `DryRun` and `Description` specifically, I'd test each operation in isolation.

internal/setupkeys/setup_keys.go
tgonzalezorlandoarm: IMO, this was too much abstraction for what it needs right now (I was going `YAGNI`), but I don't feel strongly about it.
I don't like the fact that it's a type because this implies casting to the type for validation and then casting back to string to actually emit the command.


Hopefully this is a good compromise? https://github.com/arm/topo/pull/75/commits/be21dceed4f9d9766cb813527357b803e14839e9

tgonzalezorlandoarm: Lmk otherwise, happy to change to exactly what you suggested if you don't like it

internal/setupkeys/setup_keys_test.go
tgonzalezorlandoarm: https://github.com/arm/topo/pull/75/commits/4216c0c82f82b2e8fa8b5a93798cda55b1f217aa

tgonzalezorlandoarm: All good points, thanks!

tgonzalezorlandoarm: I went rogue and did this in this PR

internal/setupkeys/setup_keys.go
muchzill4: I don't think the compromise is worth while, as it doesn't really achieve much. It introduces constants that aren't used outside of the `setupkeys` package. It'd make more sense if those constants where the "agreed upon" only entry point to calling functions in `setupkeys`, which gets us much closer to the type.

I think the type has big benefit compared to a string. String can be anything. With a concrete type, you are sure that the data that's flowing down the stack is validated. Yes, you'd have to cast it in the end, but frankly, that could be done in a similar vein to `ssh.Host`, which is given to `command.Docker`. In other words, `command.SSHKeyGen` could take typed `KeyType`.

I'm happy for this PR to go as is, but also think it's worth exploring tightening the types in a subsequent patch.

tgonzalezorlandoarm: https://github.com/arm/topo/pull/75/commits/3d949aefd3dd02b2540034ba6ffafb3b78e81ce2

tgonzalezorlandoarm: https://github.com/arm/topo/pull/75/commits/955b1c25e56cef98cc2e71046b47917958b4dec5

internal/setupkeys/setup_keys_test.go
tgonzalezorlandoarm: ... and reverted back in favour of a new PR as discussed offline

tgonzalezorlandoarm: Created #79

yejseo01 (approved):
LGTM! 🚀

internal/setupkeys/setup_keys.go
```diff
@@ -10,9 +10,21 @@ import (
...
 
-func NewKeySetup(target string, privKeyPath string) (operation.Sequence, error) {
+type KeyType string
+
+const (
+	KeyTypeED25519 KeyType = "ed25519"
+	KeyTypeRSA     KeyType = "rsa"
+)
+
+func NewKeySetup(target string, privKeyPath string, keyType string) (operation.Sequence, error) {
```
muchzill4:
```suggestion
func NewKeySetup(target string, privKeyPath string, keyType KeyType) (operation.Sequence, error) {
```

We want to use the custom type as input to this function, that's the whole point.

tgonzalezorlandoarm: https://github.com/arm/topo/pull/75/commits/76f4d0625a7c95850d159f09b0b18c3d94c1cce4

internal/setupkeys/setup_keys_test.go
```diff
@@ -68,3 +78,8 @@ func TestNewKeySetupDryRun(t *testing.T) {
 		})
 	}
 }
+
+func TestNewKeySetupUnsupportedKeyType(t *testing.T) {
```
muchzill4: `TestNewKeySetupUnsupportedKeyType` is not a correct name.

internal/setupkeys/setup_keys_test.go
```diff
@@ -68,3 +78,8 @@ func TestNewKeySetupDryRun(t *testing.T) {
...
+	_, err := setupkeys.ParseKeyType("ecdsa")
+	require.EqualError(t, err, "unsupported key type \"ecdsa\", supported types: ed25519, rsa")
```
muchzill4:
```suggestion
	require.EqualError(t, err, `unsupported key type "ecdsa", supported types: ed25519, rsa`)
```

Less escape-y.

internal/setupkeys/setup_keys.go
muchzill4: https://github.com/arm/topo/commit/76f4d0625a7c95850d159f09b0b18c3d94c1cce4 doesn't address my comment. You're still passing string `keyType`. to `NewKeySetup`.

muchzill4 (changes requested):

internal/setupkeys/setup_keys.go
tgonzalezorlandoarm: absolutelty right, apologieS: https://github.com/arm/topo/pull/75/commits/4651d50c716bbd80ba20b81dc2c206470fe17fa6

internal/setupkeys/setup_keys_test.go
tgonzalezorlandoarm: https://github.com/arm/topo/pull/75/commits/4651d50c716bbd80ba20b81dc2c206470fe17fa6

muchzill4 (approved):

PR #77, opened 2026-03-05
Use docker digest to pull, rather than tags.

Luke-Parkin:
The docker pull by tag on the remote can fail with "not found" or it can use stale cached content when Docker's tag resolution doesn't match what was just pushed to the registry. 

Pulling by the exact digest that docker push reports means the remote always gets the same image that was pushed - and this is cryptographically validated.

FAQ:
Q: **Why did you forego the r.buildtransfercommands feature? This messes up the dryrun! Dryrun and run have now diverged!**
A: We have decided to get rid of dryrun. I want to fix this bug now so I can continue with my actual ticket. We will later delete dryrun. For now, this leaves the buildTransferCommands so that dryrun almost works- it just has a <digest> placeholder.

### DIFF
--- a/internal/deploy/docker/operation/registry_transfer.go
+++ b/internal/deploy/docker/operation/registry_transfer.go
@@ -1,15 +1,19 @@
 package operation
 
 import (
+	"bytes"
 	"fmt"
 	"io"
 	"os/exec"
+	"regexp"
 	"sort"
 	"strings"
 
 	"github.com/arm/topo/internal/command"
 	"github.com/arm/topo/internal/ssh"
 )
+
+var digestRegexp = regexp.MustCompile(`digest: (sha256:[a-f0-9]+)`)
 
 type RegistryTransfer struct {
 	composeFile string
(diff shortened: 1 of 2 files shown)


tgonzalezorlandoarm:
Even if the dry run is removed, isn't the build transfer command still useful? Should we deviate from it? Why do we need to stop using it?

tgonzalezorlandoarm:
I prefer having the build command abstraction tbh, it makes things nicer with (almost) no extra cost

Luke-Parkin:
> Even if the dry run is removed, isn't the build transfer command still useful? Should we deviate from it? Why do we need to stop using it?

I think it's abstraction for abstractions sake, if we're not using dry run.. there's no other place that's calling it. We're adding complexity in the form of the buildCmd instead of just.. running it.

tgonzalezorlandoarm (approved):
I recorded my opinion, but not blocking the PR, happy for this to be merged as is

internal/deploy/docker/operation/registry_transfer.go
```diff
@@ -70,31 +74,76 @@ func (r *RegistryTransfer) getImagesFromCompose(w io.Writer) ([]string, error) {
...
+
+	retagCmd := command.Docker(r.targetHost, "tag", digestRef, image)
+	if err := runCmd(retagCmd, w); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func runCmd(cmd *exec.Cmd, w io.Writer) error {
```
muchzill4, line 122: We could push this to `command` package for future reuse:

```
command.Run(cmd *exec.Cmd, stdOut, stdErr io.Writer) error
command.RunWithOutput(cmd *exec.Cmd, stdOut, stdErr io.Writer) (string, error)
```

internal/deploy/docker/operation/registry_transfer_test.go
```diff
@@ -16,6 +16,45 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestParseDigestFromPushOutput(t *testing.T) {
```
muchzill4, line 19: Lovely tests.

muchzill4 (approved):
Nice work.

Indifferent about `buildTransferCommands`, though I think I'm leaning towards less abstraction here and agree with @Luke-Parkin that there's little value in retaining it in `transferImage()`.

_Kind of_ supporting dry-run with a placeholder is a nice compromise.

internal/deploy/docker/operation/registry_transfer.go
Luke-Parkin: Will follow up with that!

PR #78, opened 2026-03-05
refactor: SSH alias and raw SSH address are gathered by an interface

yejseo01:
## Changes

Before this change, the SSH Alias and raw SSH address was treated as two different struct that the user of the struct need to detect and choose which to use. Now both are gathered by an interface


Luke-Parkin:
Your PR title and description seem to disagree with the actual code as to what this PR's main purpose is?

yejseo01:
This is a PR that came up while addressing comments in this PR - https://github.com/arm/topo/pull/54. Will mark it draft for now until the other one is merged.

internal/ssh/ssh_tunnel.go
```diff
@@ -15,29 +15,31 @@ import (
 
 const TunnelPIDPlaceholder = "<ssh tunnel pid>"
 
-func ControlSocketPath(targetHost string) string {
-	hash := sha256.Sum256([]byte(targetHost))
+func ControlSocketPath(targetHost SSHConfigValues) string {
+	var hash [32]byte
+	if targetHost.configName != "" {
```
yejseo01: inverse the condition

yejseo01:
This is ready to be reviewed

internal/ssh/ssh_config_resolve.go
```diff
@@ -7,24 +7,25 @@ import (
 	"strings"
 )
 
-type sshConfigValues struct {
-	host string
-	user string
-	port string
+type SSHConfigValues struct {
+	host       Host
```
muchzill4: `Host` isn't really a `Host` sadly. It's really _destination_, which is `ssh` terms means the full connection string. (`ssh://user@host:123` or `user@host`)

I think we should stick to `string` here.

internal/ssh/ssh_config_resolve.go
```diff
@@ -7,24 +7,25 @@ import (
...
 	if raw == "" || isExplicitSSHHost(raw) {
-		return SplitUserHostPort(raw)
+		user, host, port := SplitUserHostPort(raw)
+		return SSHConfigValues{user: user, host: Host(host), port: port, configName: ""}
 	}
 
 	output, err := exec.Command("ssh", "-G", raw).Output()
 	if err != nil {
-		return "", "", ""
+		return SSHConfigValues{user: "", host: "", port: "", configName: ""}
```
muchzill4:
```suggestion
		return SSHConfigValues{}
```

Shouldn't this also propagate error up the stack?

internal/ssh/ssh_config_resolve.go
```diff
@@ -7,24 +7,25 @@ import (
 	"strings"
 )
 
-type sshConfigValues struct {
-	host string
-	user string
-	port string
+type SSHConfigValues struct {
```
muchzill4: We could improve naming here. I'm looking at `man ssh_config` and trying to not be weird:

```go
type SSHConfigEntry struct {
   Host string // <- a bit confusing, but this is the "entry name"
   HostName string
   User string
   Port string
}
```

internal/ssh/ssh_tunnel.go
```diff
@@ -50,35 +53,38 @@ func NewSSHTunnel(targetHost Host, port string, useControlSockets bool) (operati
...
+func NewSSHTunnelStart(targetHost SSHConfigValues, registryPort string, useControlSockets bool) *SSHTunnelStart {
+	return &SSHTunnelStart{TargetHost: targetHost, RegistryPort: registryPort, UseControlSockets: useControlSockets}
 }
 
 func (s *SSHTunnelStart) Command() *exec.Cmd {
-	rawHost := string(s.TargetHost)
-	user, host, port := SplitUserHostPort(rawHost)
-	hostArg := formatSSHHost(rawHost, user, host)
+	var hostArg string
+	if s.TargetHost.configName == "" {
```
muchzill4: This is not ideal, because the `SSHTunnelStart` now couples the cli building decisions to the shape of `s.TargetHost` (`SSHConfigValues`). 

Imho, we could make `SSHConfigValues` responsible for producing "cli-ready" host.

```go
func (s SSHConfigValues) CliHost() string {
   if s.TargetHost.configName ...
}
```

Heck, in many places we have this `if thing.configName == ""`, which screams coupling to me. We could leverage same idea there.

internal/ssh/ssh_tunnel.go
```diff
@@ -109,31 +115,31 @@ func (s *SSHTunnelStart) DryRun(w io.Writer) error {
...
 		}
-		return exec.Command("curl", fmt.Sprintf("%s:%s", host, ct.Port), "--max-time", "1")
+		return exec.Command("curl", fmt.Sprintf("%s:%s", hostPartStringified, ct.Port), "--max-time", "1")
 	}
 	return nil
 }
 
 func (ct *CheckSSHTunnelSecurity) Run(w io.Writer) error {
-	if ct.TargetHost.IsLocalhost() {
+	if ct.TargetHost.host.IsLocalhost() {
```
muchzill4: Why not implement `IsLocalhost` on `TargetHost` and proxy the call?

muchzill4 (commented):
Not a fan of the fact we use same type (`SSHConfigValues`) for something that comes from ssh config and something that is parsed from cli input, such as `user@host`.

I think we should model this better:

```go
type sshConfigEntry struct {
    host string // <- entry name
    sshConnectionDetails
}

type sshConnectionDetails struct {
    port string
    user string
    hostName string
}
```

Then, instead of coupling each function to `SSHConfigEntry`, we introduce an interface, which both structures defined above implement:

```
type SSHThingy interface {
    CliArgs() []string // <- returns what `ssh` invocation would need in one shot
    ControlSocketPath() string
}
```

internal/ssh/ssh_config_resolve.go
yejseo01: I think it should.

yejseo01:
@muchzill4 I changed a lot of this PR so that we don't have confusing names for different SSH targets. I also made two different structs for SSHConnection interface: SSHAlias and SSHDestination.

muchzill4:
> @muchzill4 I changed a lot of this PR so that we don't have confusing names for different SSH targets. I also made two different structs for SSHConnection interface: SSHAlias and SSHDestination.

Can we split this? 

I think the PR does too much now. I merely suggested using `string` instead of `ssh.Host` where it wasn't appropriate. I think the big refactor of introducing `SSHDestination` should be done separately, as it also includes a bunch of variable renames, which aren't even yet covered here.

> "Aim for ~250 LoC; split anything over ~500"

yejseo01:
not necessary anymore on latest main

PR #79, opened 2026-03-05
chore: make setup-keys tests follow the rest of the project

tgonzalezorlandoarm:


cmd/topo/setup_keys.go
```diff
@@ -10,7 +10,10 @@ import (
 	"github.com/spf13/cobra"
 )
 
-var privateKeyPath string
+var (
+	privateKeyPath string
+	keyType        string
+)
```
muchzill4, line 16: Seems unrelated to this PR? I assume it builds on top of another one?

tgonzalezorlandoarm: Fixed

muchzill4 (commented):
Thanks for taking time to tweak this. 

I might've not explained that well, so I'll try to do this with code snippets instead of words:

```go
// foo.go
package foo

type Foo {}

func NewFoo() Foo {}

func (f Foo) DoFooThings() {}

// -----------------

// foo_test.go
// 1. If `NewFoo` is complex, it might warrant `t.Run()`
func TestNewFoo() {
  t.Run("does certain thing under certain circumstances", ...)
  t.Run("does other thing under certain circumstances", ...)
}

// 2. If `NewFoo` is really dumb and not interesting, this might be sufficient.
// However, w/o `t.Run()` we lack the facility to describe in a nice _sentence_ what we're testing. Always tradeoffs.
func TestNewFoo() {
   ...
}

// 3. Here we're testing methods of `Foo`. Same rule applies. `t.Run` is usually nice for description's sake.
func TestFoo() {
   t.Run("DoFooThings", func() {
       t.Run("does certain thing under certain circumstances", ...)
       t.Run("does other thing under certain circumstances", ...)
   })
}
```

- Note, that `foo_test.go` is only allowed to test structs/functions defined in `foo.go`. This makes it nice and consistent to find "unit" tests for the code you're looking at.
- The main benefit of (ab)using `t.Run`, other than logical grouping and readability, is that you can easily run groups of tests if your editor/ide does "run tests under cursor" type thing.

internal/setupkeys/setup_keys_test.go
```diff
@@ -13,73 +13,87 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestNewKeySetupDryRun(t *testing.T) {
```
yejseo01, line 16: In all other folders in setupkeys, it seems like the test files implement a function for one public function in the corresponding file that they are testing. So `setup_keys_test.go` should have `TestNewKeySetup` and `TestGetDefaultPrivateKeyPath` as two test functions representing each function in `setup_keys.go`.

internal/setupkeys/setup_keys_test.go
```diff
@@ -13,73 +13,87 @@ import (
...
+		targetSlug := ssh.Host(target).Slugify()
+
+		got, err := setupkeys.GetDefaultPrivateKeyPath(targetSlug)
+		require.NoError(t, err)
+		require.Equal(t, filepath.Join(tmp, ".ssh", "id_ed25519_topo_user_some1thing.com"), got)
+	})
+
+	t.Run("SSHKeyGen operation dry run and description", func(t *testing.T) {
+		keyPath := filepath.Join(t.TempDir(), "custom_keys", "id_ed25519_custom")
+		op := sshkeygen.NewSSHKeyGen(
```
yejseo01: This function doesn't exist in `setup_keys.go`, and the function is already tested in `sshkeygen/ssh_keygen_test.go`

internal/setupkeys/setup_keys_test.go
```diff
@@ -13,73 +13,87 @@ import (
...
+		tmp := t.TempDir()
+		testutil.SetHomeDir(t, tmp)
+
+		target := "user@some1thing.com"
+		targetSlug := ssh.Host(target).Slugify()
+
+		got, err := setupkeys.GetDefaultPrivateKeyPath(targetSlug)
+		require.NoError(t, err)
+		require.Equal(t, filepath.Join(tmp, ".ssh", "id_ed25519_topo_user_some1thing.com"), got)
+	})
```
yejseo01: Can we put this inside `TestGetDefaultPrivateKeyPath`

internal/setupkeys/setup_keys_test.go
```diff
@@ -13,73 +13,87 @@ import (
...
+
+	t.Run("PubKeyTransfer operation dry run and description", func(t *testing.T) {
+		privKeyPath := filepath.Join(t.TempDir(), "custom_keys", "id_ed25519_custom")
+		op := pubkeytransfer.NewPubKeyTransfer(
+			"Transfer public key to target and set it as an authorized key",
+			"user@some3thing.com",
+			privKeyPath,
+			pubkeytransfer.PubKeyTransferOptions{},
+		)
```
yejseo01: this is already tested in a different file(setupkeys/pubkeytransfer/pub_key_transfer_test.go)

internal/setupkeys/setup_keys_test.go
tgonzalezorlandoarm: Done!

internal/setupkeys/setup_keys_test.go
tgonzalezorlandoarm: done!

tgonzalezorlandoarm:
I think I've addressed all your comments now!

internal/setupkeys/setup_keys_test.go
```diff
@@ -13,87 +12,80 @@ import (
...
-			inputKeyPath: filepath.Join("custom_keys", "id_rsa_custom"),
-			wantTarget:   "user@some3thing.com",
-			wantKeyPath:  filepath.Join("custom_keys", "id_rsa_custom"),
-		},
-	}
-	for _, tt := range tests {
-		t.Run(tt.name, func(t *testing.T) {
-			tmp := t.TempDir()
-			testutil.SetHomeDir(t, tmp)
+func TestSetupKeys(t *testing.T) {
```
muchzill4: Does this follow the pattern?

tgonzalezorlandoarm: Since the comment wasn't specific, I've assumed it meant that there should be no for loop and that `TestSetupKeys` should actually be called `TestNewKeySetup`.  Let me know if something else is needed or I've assumed incorreclt.y

muchzill4: Sorry, I should've been more explicit - I meant `TestSetupKeys` should be `TestNewKeySetup`. 👍

muchzill4 (approved):

PR #80, opened 2026-03-06
feat: explain why check has failed

muchzill4:
## Changes

- Provides explainers why a health check has failed as opposed to just rendering ❌.

Screenshot 2026-03-06 at 18 44 29


internal/health/dependencies.go
tgonzalezorlandoarm: We are ignoring this error

internal/health/health.go
```diff
@@ -68,15 +76,21 @@ func generateTargetReport(targetStatus Status) TargetReport {
...
+	remoteCPUs := targetStatus.Hardware.RemoteCPU
+	if len(remoteCPUs) > 0 {
+		names := make([]string, len(remoteCPUs))
+		for i, remoteProc := range remoteCPUs {
+			names[i] = remoteProc.Name
+		}
+		report.SubsystemDriver.Healthy = true
+		report.SubsystemDriver.Value = strings.Join(names, ", ")
+	} else {
+		report.SubsystemDriver.Healthy = false
```
tgonzalezorlandoarm, line 90: We are dropping the error in ProbeRemoteproc

internal/health/health.go
```diff
@@ -37,17 +37,25 @@ type Report struct {
...
+		for _, dep := range deps {
+			if dep.Error == nil {
+				installedNames = append(installedNames, dep.Dependency.Name)
+			} else {
+				errorMessages = append(errorMessages, dep.Error.Error())
+			}
+		}
+		var value string
+		if len(installedNames) > 0 {
+			value = strings.Join(installedNames, ", ")
```
tgonzalezorlandoarm, line 51: what happens if we have installed names > 0 but there's one error? This one gets ignored, right?

internal/health/health.go
```diff
@@ -37,17 +37,25 @@ type Report struct {
 
 func generateDependencyReport(statuses []DependencyStatus) []HealthCheck {
 	res := []HealthCheck{}
-	availableDepsByCategory := CollectAvailableByCategory(statuses)
-
-	for category, installedDependencies := range availableDepsByCategory {
-		names := make([]string, len(installedDependencies))
-		for i, dep := range installedDependencies {
-			names[i] = dep.Dependency.Name
+	for category, deps := range groupByCategory(statuses) {
```
tgonzalezorlandoarm, line 40: Are we guaranteed to have the same ordering everytime?
Would it be nicer to have strict ordering somehow for error reporting?

internal/health/dependencies_test.go
tgonzalezorlandoarm, line 19: typo here.

tgonzalezorlandoarm, line 14: This is an incorrect binary name, and the test actually expects it to fail.

internal/health/health.go
tgonzalezorlandoarm: Do we need a test for ProbeRemoteproc?

internal/health/health.go
muchzill4: Yes, because then we consider the "overall" check healthy.

internal/health/dependencies_test.go
muchzill4: I'm so confused. Not only this is not subject of this PR, but also the test is.. correct in its assertion? 🤔

muchzill4: Explain.

internal/health/dependencies.go
muchzill4: Good shout. I'll see if I can make `binaryExists` behave so that we can use the error. Right now it's returning a boolean indicating if binary exists and the error doesn't indicate "why" it doesn't exist.

internal/health/dependencies_test.go
tgonzalezorlandoarm: on

tgonzalezorlandoarm: `binary regex fails on incorrect binary name`

tgonzalezorlandoarm: no?

tgonzalezorlandoarm: Yes, sorry about not being the subject of the PR, I got distracted by the surroundings when reviewing

tgonzalezorlandoarm: As there's extra things I need to undersatnd

internal/health/health.go
muchzill4: Am I crazy thinking the random ordering is a pre-existing issue? If so, I'd rather fix it in a subsequent PR.

Needs deciding if it's the presentation layer that's responsible for "ordering" the health check. One might want to display unhealthy items first, not necessarily ordered by name. 🤔

internal/health/dependencies_test.go
tgonzalezorlandoarm: As to whether it's a correct assetion... shouldn't binary names be all together (no whitespaces in the middle)? It's asserting False, so the name should be `binary regex rejects an incorrect binary name`, right?

muchzill4: I think it's correct as is, in other words: "binary regex will fail an incorrect binary name".

internal/health/health.go
tgonzalezorlandoarm: yes, that's fine, I'm just raising questions more than giving answers, but happy for it to be for another PR if it leads to something!

internal/health/health.go
tgonzalezorlandoarm: That makes sense, thanks!

internal/health/dependencies_test.go
muchzill4: Yes, it's backwards. https://github.com/arm/topo/pull/81

internal/health/health.go
Luke-Parkin: >Am I crazy thinking the random ordering is a pre-existing issue? If so, I'd rather fix it in a subsequent PR.

It is a pre-existing issue and it's quite annoying!

awphi: +1 the vscode extension manually sorts them to counteract this, would be nice if the CLI handled it!

muchzill4: > It is a pre-existing issue and it's quite annoying!

I will make your life better in a separate PR @Luke-Parkin

tgonzalezorlandoarm:
nit: 
`explain` instead of `Explain` as per https://www.conventionalcommits.org/en/v1.0.0/

internal/health/dependencies.go
muchzill4: https://github.com/arm/topo/pull/80/commits/ebfe64f7df6539d8714d119fad063e59a94135e0

internal/health/health.go
muchzill4: This one is less obvious to me and probably best addressed separately. For now, it's safe to assume that lack of detected remoteprocs (for any reason!) means there's no remoteprocs.

muchzill4:
Updated screenshot, do you think it's a bit much or just enough?

tgonzalezorlandoarm (approved):
LGTM 🚀

PR #82, opened 2026-03-09
fix: ensure predictable order or `topo health` output

muchzill4:
## Changes

- Introduce reusable `collections.GroupBy`, which retains the order of grouped items
- Use the above when aggregating health check results

### DIFF
--- a/internal/collections/collections.go
+++ b/internal/collections/collections.go
@@ -5,6 +5,8 @@ type Group[I any, K comparable] struct {
 	Members []I
 }
 
+// GroupBy partitions items into groups by applying keyFn to each element.
+// Groups are returned in the order their keys first appear in items.
 func GroupBy[I any, K comparable](items []I, keyFn func(I) K) []Group[I, K] {
 	order, groupMap := groupBy(items, keyFn)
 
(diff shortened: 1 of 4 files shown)


muchzill4:
~Depends on #83, which fixes CI.~

internal/collections/collections.go
```diff
@@ -0,0 +1,38 @@
...
+		groups[i] = Group[I, K]{
+			Key:     key,
+			Members: groupMap[key],
+		}
+	}
+
+	return groups
+}
+
+func groupBy[I any, K comparable](slice []I, keyFn func(I) K) ([]K, map[K][]I) {
```
th3james: Bit uncomfortable with the same name being used for two behaviours (admittedly with different return types). How about

```suggestion
func groupAndCollectOrder[I any, K comparable](slice []I, keyFn func(I) K) ([]K, map[K][]I) {
```

th3james (approved):

internal/collections/collections.go
muchzill4: https://github.com/arm/topo/pull/82/commits/367687404fd5e93216e34005e85c507fae0a845e

PR #85, opened 2026-03-09
feat!: missing remoteproc endpoint is a warning, not an error

muchzill4:
## Changes

- Introduce enum to represent health check status in place of boolean healthy/unhealthy
- When no remoteproc devices are found on the target, it's a warning not an error
- ⚠️ This is a breaking change for consumers of `health --output=json`, as the field is no longer boolean but a string

Screenshot 2026-03-09 at 15 19 09


tgonzalezorlandoarm (approved):

federicobozzini (approved):
Great change 👍

PR #86, opened 2026-03-09
feat: support --target for template compatibility markers

tgonzalezorlandoarm:
## Changes

Use `--target` on topo templates to generate a target description on the fly and annotate compatibility markers without requiring `--target-description`.
Keep --target-description behavior unchanged, make the flags mutually exclusive.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,11 @@ topo templates --target-description target-description.yaml
 ```
 
 This lists available templates and indicates compatibility with your target hardware.
+If you don't already have a target description file for your board, you can still use:
+
+```sh
+topo templates --target my-board
+```
 
 ### 4. Clone a template into a new project
 
--- a/cmd/topo/templates.go
+++ b/cmd/topo/templates.go
@@ -34,6 +34,16 @@ var templatesCmd = &cobra.Command{
 			if err != nil {
 				return err
 			}
+		} else {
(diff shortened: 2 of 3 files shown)


muchzill4 (approved):
LGTM.

---

Perhaps for another PR: 
I do wonder if we should have a: 
```go
retrieveTargetDescription(cmd *cobra.Command) (*target.HardwareDescription, error)
addTargetDescriptionFlag(cmd *cobra.Command) // adds `--target-description`
addTargetDescriptionFlags(cmd *cobra.Command) // or `addTargetFlags()` - this would add both `--target` and `--target-description`
```
It'd essentially do the flow you're doing here, but be reusable for future commands.

tgonzalezorlandoarm:
> LGTM.
> 
> Perhaps for another PR: I do wonder if we should have a:
> 
> ```go
> retrieveTargetDescription(cmd *cobra.Command) (*target.HardwareDescription, error)
> addTargetDescriptionFlag(cmd *cobra.Command) // adds `--target-description`
> addTargetDescriptionFlags(cmd *cobra.Command) // or `addTargetFlags()` - this would add both `--target` and `--target-description`
> ```
> 
> It'd essentially do the flow you're doing here, but be reusable for future commands.

I've created a draft PR for this so that I don't forget: #92

PR #87, opened 2026-03-09
refactor: use custom marshaller to ensure non-null slices in json

muchzill4:
## Changes

- Replaces _hacky_ way of ensuring absence of `null` in json for `nil` slice values with custom marshaller

### DIFF
--- a/internal/health/health.go
+++ b/internal/health/health.go
@@ -1,6 +1,7 @@
 package health
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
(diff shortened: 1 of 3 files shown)


internal/health/health.go
```diff
@@ -39,13 +40,29 @@ type HostReport struct {
...
+	if r.Dependencies == nil {
+		r.Dependencies = []HealthCheck{}
+	}
+	return json.Marshal(Alias(r))
+}
+
 type TargetReport struct {
 	IsLocalhost     bool          `json:"isLocalhost"`
 	Connectivity    HealthCheck   `json:"connectivity"`
 	Dependencies    []HealthCheck `json:"dependencies"`
```
Luke-Parkin, line 54: (Just to ensure I understand this)

This can be Nil, because slices in go are nil by defaut

internal/health/health.go
```diff
@@ -39,13 +40,29 @@ type HostReport struct {
...
 	SubsystemDriver HealthCheck   `json:"subsystemDriver"`
 }
 
+func (r TargetReport) MarshalJSON() ([]byte, error) {
+	type Alias TargetReport
+	if r.Dependencies == nil {
+		r.Dependencies = []HealthCheck{}
+	}
+	return json.Marshal(Alias(r))
```
Luke-Parkin, line 63: So.. if we try to json.Marshal a struct with a nil field without checking it will blow up.. somewhere?

muchzill4: It won't blow up, but a `nil` slice will serialize to `null`. We want `[]` (for some reason, I presume JS ergonomics).

This is covered by json rendering tests @ `internal/templates/health_test.go`

Is it worth me adding some tests in `internal/health/health_test.go`? 🤔

internal/health/health.go
muchzill4: Yes.

internal/health/health.go
Luke-Parkin: It's probablly fineee but I'll leave it to your judgement.

Luke-Parkin (approved):

Luke-Parkin:
Good catch!

internal/health/health.go
muchzill4: https://github.com/arm/topo/pull/87/commits/c82c0b36b7454fef598e4efeb9e9149c59a52228

Luke-Parkin:
Nice :)

PR #88, opened 2026-03-10
refactor: decouple host and target reporting

muchzill4:
## Changes

- Split `Check` into `CheckHost` and `CheckTarget`, which I'm planning to subsequently use in the cli layer to conditionally generate target report only when `--target` has been specified
- Reorder definitions in `health.go` (public api first, private last)
- Split existing `GenerateReport` tests into `GenerateHostReport` and `GenerateTargetReport`

### DIFF
--- a/internal/health/health.go
+++ b/internal/health/health.go
@@ -34,6 +34,7 @@ type HealthCheck struct {
 	Status CheckStatus `json:"status"`
 	Value  string      `json:"value"`
 }
+
 type HostReport struct {
 	Dependencies []HealthCheck `json:"dependencies"`
 }
--- a/internal/health/health.go
+++ b/internal/health/health.go
@@ -51,31 +51,33 @@ type Report struct {
 	Target TargetReport `json:"target"`
 }
 
-func generateDependencyReport(statuses []DependencyStatus) []HealthCheck {
-	res := []HealthCheck{}
-	for _, group := range groupByCategory(statuses) {
-		hc := HealthCheck{Name: group.Key}
+func GenerateReport(hostDependencies []DependencyStatus, targetStatus Status) Report {
+	report := Report{}
+	report.Host = generateHostReport(hostDependencies)
+	report.Target = generateTargetReport(targetStatus)
 
-		var installedNames, errorMessages []string
-		for _, dep := range group.Members {
-			if dep.Error == nil {
-				installedNames = append(installedNames, dep.Dependency.Name)
-			} else {
-				errorMessages = append(errorMessages, dep.Error.Error())
-			}
-		}
+	return report
+}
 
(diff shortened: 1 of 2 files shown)


internal/health/health.go
```diff
@@ -50,41 +51,54 @@ type Report struct {
...
-	return res
+	conn := target.NewConnection(sshTarget, opts)
+	targetStatus := ProbeHealthStatus(conn)
+	if errors.Is(targetStatus.ConnectionError, target.ErrPasswordAuthentication) {
+		return TargetReport{}, fmt.Errorf(passwordAuthErrorMessage, sshTarget)
+	}
+	return GenerateTargetReport(targetStatus), nil
+}
+
+func GenerateReport(hostDependencies []DependencyStatus, targetStatus Status) Report {
```
Luke-Parkin: There's a slight asymmetry here between `[]DependencyStatus' and 'Status' And hostDependencies and targetStatus.

But actually you didn't change this so please ignore

internal/health/health.go
```diff
@@ -50,41 +51,54 @@ type Report struct {
...
-			if dep.Error == nil {
-				installedNames = append(installedNames, dep.Dependency.Name)
-			} else {
-				errorMessages = append(errorMessages, dep.Error.Error())
-			}
-		}
+func Check(sshTarget string, acceptNewHostKeys bool) (Report, error) {
+	targetReport, err := CheckTarget(sshTarget, acceptNewHostKeys)
+	if err != nil {
+		return Report{}, nil
```
Luke-Parkin: Error swallowed? This does actually return an error as well; so we should really pass it through.

internal/health/health.go
Luke-Parkin: Wait- this function isn't even being used?

muchzill4: Can you elaborate? 🤔

internal/health/health.go
muchzill4: Nice catch. https://github.com/arm/topo/pull/88/commits/f45c3aa87f058650baee93c64531be78b4c260eb

internal/health/health.go
muchzill4: 🤦 https://github.com/arm/topo/pull/88/commits/c13ebc09f50fdc557a4d9d16509d371b51b941ed

Luke-Parkin: It feels a bit odd calling it hostDependencies for the host, and then targetStatus for the target?

hostDependencies is probably correct for the host, but targetStatus feels wrong for the target.

But this is a bit nitpicking and isn't what this pr is changing

Luke-Parkin (approved):
LGTM

PR #89, opened 2026-03-10
feat: support host-only `topo health`

muchzill4:
## Changes

⚠️ Builds on top of https://github.com/arm/topo/pull/88

- Removes the requirement to specify target for `topo health`

Screenshot 2026-03-10 at 09 37 55

### DIFF
--- a/cmd/topo/health.go
+++ b/cmd/topo/health.go
@@ -45,7 +45,11 @@ var healthCmd = &cobra.Command{
 		if err != nil {
 			return err
 		}
-		return printable.Print(templates.PrintableHealthReport(report), os.Stdout, outputFormat)
+		toPrint := templates.PrintableHealthReport{
+			Host:   report.Host,
+			Target: &report.Target,
+		}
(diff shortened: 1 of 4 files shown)


cmd/topo/health.go
```diff
@@ -38,14 +34,28 @@ var healthCmd = &cobra.Command{
 			spinner = term.StartSpinner(os.Stderr, "Checking health...")
```
Luke-Parkin, line 34: (not a this pr thing- just something i've noticed)
Is sending a spinner to standard err the normal way to do this? It does actually seem to work for piping stdout to less / nano etc. But might fail if you pipe stderr?

Luke-Parkin (approved):
LGTM

cmd/topo/health.go
muchzill4:
```
topo health 2>/tmp/health.log
cat /tmp/health.log  # should contain no spinner characters
```

But, we could prolly output to `os.Stdout`. 🤔

PR #90, opened 2026-03-11
refactor: improve `DockerComposeUp` interface

muchzill4:
## Changes

- Rename `DockerComposeRun` -> `DockerComposeUp` - where possible, operations are named after the command they actually invoke
- Stop representing mutually exclusive state with a struct - enum is more appropriate

### DIFF
--- a/cmd/topo/deploy.go
+++ b/cmd/topo/deploy.go
@@ -22,6 +22,8 @@ var (
 	noRegistry        bool
 	registryPort      string
 	skipProjectChecks bool
+	forceRecreate     bool
+	noRecreate        bool
 )
 
 var deployOpts docker.DeployOptions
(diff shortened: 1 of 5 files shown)


cmd/topo/deploy.go
```diff
@@ -98,6 +100,12 @@ Use --dry-run to see what commands would be executed without actually running th
 		goos := runtime.GOOS
 		deployOpts.WithRegistry = docker.SupportsRegistry(noRegistry, targetHost)
 		deployOpts.UseSSHControlSockets = docker.SupportsSSHControlSockets(goos)
+		switch {
```
tgonzalezorlandoarm, line 103: I don't understand the benefit of doing this? At least at first glance, it adds more layers of abstraction and checking that with no benefit?

muchzill4: You're commenting on a `switch` statement and mention "layers of abstraction". What abstraction?

tgonzalezorlandoarm: It's true that abstraction is not the right word... more complexity perhaps? More code?
We can just write straight into the structure like we were doing before and leverage on the muttually exclusive flag, I don't know why we'd need that extra switch statement

I guess `Stop representing mutually exclusive state with a struct - enum is more appropriate` is not obvious to me

muchzill4: I think being accurate leads to an API that's less confusing for the caller. In this case it's a flag that has 3 states, not 4. A struct is a poor representation of that.

Imagine:
```go
MakeDinner(Food{
  Devolay: false,
  Pizza: true,
  Spaghetti: false,
})
```

I think that's not very good. You want:

```go
MakeDinner(Pizza) // assume Pizza is a member of Food enum
```

A `switch` statement is a side effect of being accurate. I think it's a small price to pay.

tgonzalezorlandoarm: That makes sense to me, thanks for the explanation

tgonzalezorlandoarm (approved):

PR #91, opened 2026-03-11
Remove unsigned binary releases. Use Signed Artifact urls

Luke-Parkin:
Dependent on signer repo changes.

- Removes unsigned binaries from github release releases.
- Uses new signer action changes to allow artifact urls to be returned, to be added to the release page.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -17,10 +17,13 @@ on:
 permissions:
   contents: write
   packages: read
+  actions: read
 
 jobs:
   release:
     runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.version.outputs.version }}
     steps:
       - name: Checkout
         uses: actions/checkout@v6
@@ -53,6 +56,7 @@ jobs:
           echo "Next version: $new"
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
       - name: Set git config
         run: |
           git config --global user.email "topo-release@users.noreply.github.com"
@@ -73,7 +77,7 @@ jobs:
         with:
           distribution: goreleaser
           version: '~> v2'
-          args: release --clean
+          args: release --clean --skip=publish
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           GORELEASER_CURRENT_TAG: v${{ steps.version.outputs.version }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -114,83 +118,64 @@ jobs:
...
-        env:
-          GH_TOKEN: ${{ steps.app-token.outputs.token }}
-          OWNER: Arm-debug
-          REPO: signer
-          WORKFLOW_FILE: archive-and-upload-to-artifactory.yml
-          REF: main
-          UPSTREAM_RUN_ID: ${{ github.run_id }}
+  sign-macos:
+    needs: build
+    uses: Arm-debug/signer/.github/workflows/sign-and-post-macos.yml@main
```
muchzill4: Let's use a hash instead of `main`, you'll thank me later.

.github/workflows/release.yaml
```diff
@@ -114,83 +118,64 @@ jobs:
...
+      run-id: ${{ github.run_id }}
+      artifact: topo-v${{ needs.build.outputs.version }}-windows-unsigned
+      release-version: ${{ needs.build.outputs.version }}
+      repo-name: topo
+      org-name: arm
+      binary-name: topo.exe
+      signing-artifactory-repo: edge-ai-tooling.topo-cli-signing
+      artifactory-uploadables: topo_windows_*
+
+  sign-linux:
```
muchzill4: Don't need this anymore I guess?

.github/workflows/release.yaml
```diff
@@ -114,83 +118,64 @@ jobs:
...
+
+  sign-windows:
+    needs: build
+    uses: Arm-debug/signer/.github/workflows/sign-and-post-windows.yml@main
+    with:
+      run-id: ${{ github.run_id }}
+      artifact: topo-v${{ needs.build.outputs.version }}-windows-unsigned
+      release-version: ${{ needs.build.outputs.version }}
+      repo-name: topo
+      org-name: arm
```
muchzill4, line 166: Perhaps worth reordering as `org/repo`, not `repo/org`?

Also, OCD, but I'd group related arguments together.

Source: `org-name, repo-name, run-id, artifact, binary-name`
Artifactory details: `...`

muchzill4:
As an aside: My preference would be to remove `--publish` as one PR, then add signing/url output separately.

.github/workflows/release.yaml
yejseo01: Using provisional signer branch for now for testing. Will set it to a hash later

yejseo01:
not necessary anymore

PR #95, opened 2026-03-11
feat: check that docker is accessible by the user

tgonzalezorlandoarm:
Docker may be present in the system but the user may not be able to run it.

## Changes

Instead of checking for just docker being installed in the system, check if `docker info` is able to run. This guarantees that the user has docker access, which is what we really want to check.


tgonzalezorlandoarm:
If you think this is too generic and should not iterate, go to #96 and let me know.

PR #96, opened 2026-03-11
feat: check that docker is accessible by the user

tgonzalezorlandoarm:
## Changes

Docker may be present in the system but the user may not be able to run it.
Instead of checking for just docker being installed in the system, check if `docker info` is able to run. This guarantees that the user has docker access, which is what we really want to check.

### DIFF
--- a/internal/health/health.go
+++ b/internal/health/health.go
@@ -7,6 +7,8 @@ import (
 	"os"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


tgonzalezorlandoarm:
If you think this is not generic enough and we should iterate, go to #95 and let me know

PR #97, opened 2026-03-12
feat: add remoteproc as a Lightbulb-moment template feature

tgonzalezorlandoarm:
The templates catalog has the `features` argument.
Currently, this argument is presenting both features that are required to run a template on the target (like `NEON`) and also listing target features that can be optionally used to run in the target (like `SVE`, for performance comparisons).

We therefore only ask for at least one of the requirements to be met for the template to be considered compatible with the target hardware.

## Changes

- make template compatible with the target hardware if any of its feature requirements is met.
- Update the `Lightbulb-moment` template to add `remoteproc` as a required feature.


th3james:
@tgonzalezorlandoarm was this generated by running `go run ./scripts/update_templates`? These files shouldn't be edited by hand

tgonzalezorlandoarm:
> @tgonzalezorlandoarm was this generated by running `go run ./scripts/update_templates`? These files shouldn't be edited by hand

No it was not, I missed this, thanks for spotting it!

th3james (approved):

PR #101, opened 2026-03-12
feat: suggest a fix when warning about missing remoteproc runtime

muchzill4:
## Changes

- Turns missing `remoteproc-runtime` components into warnings instead of errors
- Adds facility to suggest fixes (right now only done for `remoteproc-runtime`)

Screenshot 2026-03-12 at 19 59 09

## Implementation details

- `Dependencies` can now have multiple `Checks` - this breaks the implicit "binary exists", where we had no control over the outcome of such check
   - Side effect: will facilitate doing `docker info` check for @tgonzalezorlandoarm 
- We now `PerformChecks` instead of `CheckInstalled`

### DIFF
--- a/internal/health/dependencies.go
+++ b/internal/health/dependencies.go
@@ -28,6 +28,7 @@ type Check struct {
 	Kind     CheckKind
 	Arg      string
 	Severity CheckSeverity
+	Fix      string
 }
 
 func BinaryExists() Check {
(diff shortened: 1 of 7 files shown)


internal/health/dependencies.go
```diff
@@ -7,6 +7,38 @@ import (
 	"github.com/arm/topo/internal/ssh"
 )
 
+type CheckKind int
+
+const (
+	CheckBinaryExists CheckKind = iota
```
muchzill4, line 13: A new check kind, could be `CheckCommandSucceeds` or `CommandOutputMatches` (for versions).

I'm still undecided if this is the right way to go, but we'll see when we add more checks. I thought about using a `Checker` interface such as:

```go
type Runner interface {
   BinaryExists(name string)
   Exec(command string)
}

type Checker interface {
   Check(r Runner)
}
```

This would let us have more flexibility in `Checks`, but sounded like a bit too much complexity for now.

tgonzalezorlandoarm (approved):
Very nice work! LGTM!

PR #104, opened 2026-03-13
feat!: make path optional in clone command

muchzill4:
## Changes

- Make path in `topo clone` optional
- This is a breaking change as it required reordering the arguments: `topo <path> <source>` becomes `topo <source> [path]`

### DIFF
--- a/cmd/topo/clone.go
+++ b/cmd/topo/clone.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"os"
+	"strings"
 
 	"github.com/arm/topo/internal/arguments"
 	"github.com/arm/topo/internal/output/console"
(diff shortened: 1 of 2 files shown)


cmd/topo/clone.go
```diff
@@ -12,51 +13,66 @@ import (
...
-  topo clone my-demo dir:/path/to/template/folder
-  topo clone my-demo dir:./relative/path
+  topo clone dir:/path/to/template/folder
+  topo clone dir:./relative/path
 
 Some projects require build arguments. Supply them on the command line or answer prompts:
 
-  # Will prompt for required args
-  topo clone my-demo template:Hello-World
+  # Will prompt for required args (path defaults to template name)
```
federicobozzini: Should we specifiy what the default is also for the other types?

federicobozzini (approved):

cmd/topo/clone.go
muchzill4: I have fixed the issue by not specifying it at all: https://github.com/arm/topo/pull/104/commits/f536bd26368198023b3a0d5e732f8d81a8c865ac. :trollface:

federicobozzini (approved):

PR #105, opened 2026-03-13
fix: make authentication errors comply with current schema

tgonzalezorlandoarm:
Currently, when authentication errors happen, they are shown to the user in a different way than the rest of the errors in the codebase.

## Changes

- Align `password-based authentication errors` to the current error schema.

### DIFF
--- a/internal/health/health.go
+++ b/internal/health/health.go
@@ -10,11 +10,6 @@ import (
 	"github.com/arm/topo/internal/target"
 )
 
-// #nosec G101 -- Does not contain hardcoded credentials
-const passwordAuthErrorMessage = `note: Topo does not support SSH password-based authentication. To connect, either:
-- create your own SSH keys for the target, or
-- run 'topo setup-keys --target %s' to let Topo generate keys and configure passwordless authentication`
-
 type CheckStatus string
 
 func NewCheckStatusFromError(err error) CheckStatus {
@@ -79,9 +74,6 @@ func CheckTarget(sshTarget string, acceptNewHostKeys bool) (TargetReport, error)
 	}
 	conn := target.NewConnection(sshTarget, opts)
 	targetStatus := ProbeHealthStatus(conn)
-	if errors.Is(targetStatus.ConnectionError, target.ErrPasswordAuthentication) {
-		return TargetReport{}, fmt.Errorf(passwordAuthErrorMessage, sshTarget)
-	}
 	return GenerateTargetReport(targetStatus), nil
 }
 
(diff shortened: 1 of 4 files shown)


internal/health/health_test.go
```diff
@@ -62,6 +62,8 @@ func TestGenerateTargetReport(t *testing.T) {
 		got := health.GenerateTargetReport(ts)
 
 		assert.Equal(t, health.CheckStatusError, got.Connectivity.Status)
+		assert.Equal(t, assert.AnError.Error(), got.Connectivity.Value)
+		assert.Empty(t, got.Connectivity.Fix)
```
muchzill4: I'm indifferent, though personally I'd skip the `Empty` assertions. It might be overly defensive. 🤷

internal/health/health.go
```diff
@@ -120,6 +108,22 @@ func GenerateTargetReport(targetStatus Status) TargetReport {
...
+		Name:   "Connectivity",
+		Status: NewCheckStatusFromError(targetStatus.ConnectionError),
+	}
+	if targetStatus.ConnectionError == nil {
+		return check
+	}
+
+	check.Value = targetStatus.ConnectionError.Error()
+	if errors.Is(targetStatus.ConnectionError, target.ErrPasswordAuthentication) {
+		check.Fix = fmt.Sprintf(`To connect, create your own SSH keys for the target, or run 'topo setup-keys --target %s' to let Topo generate keys and configure passwordless authentication`, targetStatus.SSHTarget)
```
muchzill4: I wonder if we could make this more succinct. Also, I might've made a mistake in the OG fix, as I didn't use a sentence. Whatever we go with, we should be consistent - so let's either update runtime fix to be a sentence, or make this not a sentence.

More succint (also using backticks for the command, which I've done for the other command - I'm precious about that 😉):
```
run `topo setup-keys` or manually setup SSH keys for the target
```

I'm a bit jealous that you're easily able to specify `--target %s` here, though does it handle the case if target is coming from `TOPO_ENV`, then `--target` is redundant, no? I wonder if we should just say `topo setup-keys`, though your version is likely copy-pasteable.

muchzill4 (commented):
LGTM.

The only change I'm requesting is making the Fix a bit more succinct.

muchzill4:
Nit: is it a `fix:`? 🤔

tgonzalezorlandoarm:
> Nit: is it a `fix:`? 🤔


done!

internal/health/health.go
tgonzalezorlandoarm: Done!

Yes, `--target` would be redundant in that case, but I guess that's the price we have to pay for covering all cases and to make it copy-pasteable, as you mentioned.

internal/health/health_test.go
tgonzalezorlandoarm: done!

muchzill4 (approved):

PR #107, opened 2026-03-13
ci: Make release use new signer changes

Luke-Parkin:
## Changes

- This allows use of the new signer workflow to release, and get urls back in the releases.

It is currently a bit messy with lots of args- this will be fixed in future changes to the CI.


Luke-Parkin:
https://github.com/arm/topo/actions/runs/23140718749

yejseo01 (approved):
LGTM

yejseo01:
Minor issue with parameter names, but we will get rid of most of it in next PR

PR #108, opened 2026-03-13
refactor: extract ssh error classification

muchzill4:
## Changes

- Introduce `ssh.ErrSSH`, `ssh.ErrAuthFailed`, and `ssh.ErrConnectionFailed` a small error hierarchy where the specific errors wrap the base, allowing callers to handle them individually or catch both with `errors.Is(err, ssh.ErrSSH)`
- Add `ssh.ClassifyStderr` to pattern-match SSH stderr output into typed errors
- Call `ClassifyStderr` in `conn.Run` so all callers automatically get structured errors without inspecting strings themselves
- Remove `ErrSSHAuthFailed` from `install` and replace the SSH pattern matching in `install()` with `errors.Is(err, ssh.ErrSSH)`
  - Keep `install.ErrPermissionDenied` and a local `classifyStderr` for install-command-specific patterns (`permission denied`, `cannot create`, `read-only`), since these are not SSH errors

### DIFF
--- a/internal/install/install.go
+++ b/internal/install/install.go
@@ -25,10 +25,17 @@ const (
 	downloadTimeout = 2 * time.Minute
 )
 
-var (
-	ErrPermissionDenied = errors.New("permission denied")
-	ErrSSHAuthFailed    = errors.New("SSH authentication failed")
-)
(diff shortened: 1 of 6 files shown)


awphi (approved):
Nice!

PR #109, opened 2026-03-13
feat: verify Docker is runnable during health dependency checks

tgonzalezorlandoarm:
## Changes

- Add a new runnable-binary dependency check alongside the existing binary presence check.
- Use it for the host Docker dependency by running docker info, so health checks fail when Docker is installed but not accessible to the current user.

### DIFF
--- a/internal/health/dependencies.go
+++ b/internal/health/dependencies.go
@@ -78,7 +78,7 @@ var HostRequiredDependencies = []Dependency{
 			Kind:     CheckCommandSuccessful,
 			Arg:      "docker info",
 			Severity: SeverityError,
-			Fix:      "Ensure docker can be executed successfully by the current user",
+			Fix:      "Ensure current user can run docker commands",
 		}},
 	},
 }
@@ -92,7 +92,7 @@ var TargetRequiredDependencies = []Dependency{
 			Kind:     CheckCommandSuccessful,
 			Arg:      "docker info",
 			Severity: SeverityError,
-			Fix:      "Ensure docker can be executed successfully by the current user",
+			Fix:      "Ensure current user can run docker commands",
 		}},
 	},
 	{
--- a/internal/health/dependencies.go
+++ b/internal/health/dependencies.go
@@ -3,6 +3,7 @@ package health
 import (
 	"fmt"
 	"os/exec"
+	"strings"
 
 	"github.com/arm/topo/internal/ssh"
 )
@@ -11,6 +12,7 @@ type CheckKind int
 
 const (
 	CheckBinaryExists CheckKind = iota
+	CheckCommandSuccessful
 )
 
 type CheckSeverity int
@@ -72,7 +74,12 @@ var HostRequiredDependencies = []Dependency{
 		Binary:         "docker",
 		Label:          "Container Engine",
 		SoftwareEnumID: Docker,
-		Checks:         []Check{BinaryExists()},
+		Checks: []Check{BinaryExists(), {
+			Kind:     CheckCommandSuccessful,
+			Arg:      "docker info",
+			Severity: SeverityError,
+			Fix:      "Ensure docker can be executed successfully by the current user",
+		}},
 	},
 }
 
@@ -81,7 +88,12 @@ var TargetRequiredDependencies = []Dependency{
 		Binary:         "docker",
 		Label:          "Container Engine",
 		SoftwareEnumID: Docker,
-		Checks:         []Check{BinaryExists()},
+		Checks: []Check{BinaryExists(), {
+			Kind:     CheckCommandSuccessful,
+			Arg:      "docker info",
+			Severity: SeverityError,
+			Fix:      "Ensure docker can be executed successfully by the current user",
+		}},
 	},
 	{
 		Binary:                "remoteproc-runtime",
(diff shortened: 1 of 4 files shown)


internal/health/dependencies.go
```diff
@@ -11,6 +12,7 @@ type CheckKind int
 
 const (
 	CheckBinaryExists CheckKind = iota
+	CheckBinaryCanBeRun
```
muchzill4: I think it should be something more generic, like `CheckCommandSuccessful` or similar.

internal/health/dependencies.go
```diff
@@ -72,7 +83,7 @@ var HostRequiredDependencies = []Dependency{
 		Binary:         "docker",
 		Label:          "Container Engine",
 		SoftwareEnumID: Docker,
-		Checks:         []Check{BinaryExists()},
+		Checks:         []Check{BinaryExists(), BinaryCanBeRun("info")},
```
muchzill4: I think we could skip `func BinaryCanBeRun()` and just inline the `Check{}` here. This also lets you have a more specific `Fix:`.

muchzill4: I also wonder if `Arg` shouldn't be the entire command to run. I was thinking of it as `Arg` to pass to the check function:
- In case of `BinaryExists` we could argue that we should duplicate `"docker"` and not rely on `Dependency.Binary`. In other words, update the `BinaryExists()` to `BinaryExists("docker")` - I went back and forth on this.
- In case of `CheckCommandSuccessful("info")` this is more jarring. I dislike the implicit coupling to "docker". Especially that it's possible that we might want to run some other command, unrelated to `Dependency.Binary`.

internal/health/dependencies.go
```diff
@@ -199,3 +215,16 @@ func BinaryExistsLocally(bin string) error {
 	}
 	return nil
 }
+
+func BinaryCanBeRunLocally(bin string, args string) error {
```
muchzill4: This could be `Execute` or `Run`, right? For remote target, we could leverage `Connection.Run` perhaps?

internal/health/dependencies.go
tgonzalezorlandoarm: done

internal/health/dependencies.go
tgonzalezorlandoarm: hopefull this is done now

internal/health/dependencies.go
tgonzalezorlandoarm: Well, not exactly because run output `string, error`, but did a small tweak to adapt that hopefully works for you?

internal/health/dependencies.go
```diff
@@ -72,7 +74,12 @@ var HostRequiredDependencies = []Dependency{
 		Binary:         "docker",
 		Label:          "Container Engine",
 		SoftwareEnumID: Docker,
-		Checks:         []Check{BinaryExists()},
+		Checks: []Check{BinaryExists(), {
+			Kind:     CheckCommandSuccessful,
+			Arg:      "docker info",
+			Severity: SeverityError,
+			Fix:      "Ensure docker can be executed successfully by the current user",
```
muchzill4: Is this the fix? 🤔 What does it mean to execute docker? One might think that `docker` is not chmodded correctly.

The only thing I can think of is: "Ensure current user can run docker commands", but I'm not sure if that catches that. WDYT?

internal/health/dependencies_test.go
```diff
@@ -190,14 +196,66 @@ func TestPerformChecks(t *testing.T) {
...
 			{Dependency: health.Dependency{Binary: "standalone", Label: "Tools", Checks: []health.Check{health.BinaryExists()}}, Error: nil},
 		}
 		assert.Equal(t, want, got)
 	})
+
+	t.Run("captures failure from a command successful check", func(t *testing.T) {
+		dep := health.Dependency{
+			Binary: "docker",
+			Label:  "Container Engine",
+			Checks: []health.Check{health.BinaryExists(), {
```
muchzill4, line 213:
```suggestion
			Checks: []health.Check{{
```

Don't need this here, it's not part of the test.

internal/health/dependencies_test.go
```diff
@@ -190,14 +196,66 @@ func TestPerformChecks(t *testing.T) {
...
+			{
+				Dependency: dep,
+				Error:      errors.New("permission denied"),
+				Fix:        "Ensure docker can be executed successfully by the current user",
+			},
+		}
+		assert.Equal(t, want, got)
+	})
+
+	t.Run("passes runnable check args through from the check definition", func(t *testing.T) {
```
muchzill4: Do you think it's worth merging this with above test? We could configure the mock so that it only errors when receiving `"docker --version"`, which gives us confidence that it was called correctly.

muchzill4 (approved):
LGTM

internal/health/dependencies.go
tgonzalezorlandoarm: This sounds reasonable! Done: https://github.com/arm/topo/pull/109/commits/b8ce6567e0df1dd81810fc1364ee9c3b286af49c

internal/health/dependencies_test.go
tgonzalezorlandoarm: Nice, this covers both cases.
Done: https://github.com/arm/topo/pull/109/commits/24fba58185154c862c88be7c4f76ec08751c4390

internal/health/dependencies_test.go
tgonzalezorlandoarm: done https://github.com/arm/topo/pull/109/commits/24fba58185154c862c88be7c4f76ec08751c4390

muchzill4 (approved):
🚢 🇮🇹

PR #115, opened 2026-03-17
refactor: ssh config resolution

muchzill4:
## Changes

- Introduced a `Config` type in the `ssh` package with a cleaner public API for reading SSH config values via `ssh -G`.
- Replaced the internal `resolveSSHConfigHost` helper (which returned three values) with a  focused `resolveHost` function that returns only the hostname. 
   - Moved the resolution logic closer to where it's used.

---

1. This is a foundation I'm building for myself to parse timeouts from ssh config, where the default timeout we'd set for ssh connections could be different from user's configuration, we should respect.
2. I'm starting to think that we could always call `ssh -G <destination>` as it seems to parse destinations really nicely and return `host`, `port` and `user`. Something for the future.


yejseo01 (approved):
Looks good to me

PR #116, opened 2026-03-17
refactor: Rename Host to SSHDetination

yejseo01:
## Changes

- Host is not actually a host. It's a string that includes user, host and port. Renaming the type to SSHDestination


yejseo01:
not necessary anymore

PR #118, opened 2026-03-17
ci: run e2e test packages in series

awphi:
## Background

- My suspicion is that most of the flaky e2e tests are race conditions stemming from the parallelization of go's test runner at the package level since our e2e tests are spread across different packages

## Changes

- This PR runs all test packages in series w/ `-p 1` (default is `-p <cores>`) as a simple bandaid that will slow the CI down but should improve its stability. We can revisit this in the future if speed becomes a problem.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -92,6 +92,10 @@ jobs:
       - name: Build
         run: |
           go build -v ./cmd/topo
+  
+      - name: Run unit tests
+        run: |
+          go test ./internal/... -v
 
       - name: Run e2e tests
         env: 
@@ -99,7 +103,3 @@ jobs:
         run: |
           docker build -t topo-e2e-target:latest ./internal/testutil/test-container
           go test ./e2e/... -p 1 -v
-
-      - name: Run unit tests
-        run: |
-          go test ./internal/... -v
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -93,9 +93,13 @@ jobs:
         run: |
           go build -v ./cmd/topo
 
-      - name: Run tests with coverage
+      - name: Run e2e tests
         env: 
           DOCKER_HOST: ${{ matrix.docker_host }}
         run: |
           docker build -t topo-e2e-target:latest ./internal/testutil/test-container
-          go test ./... -p 1 -v -covermode=atomic -coverprofile=coverage.out
+          go test ./e2e/... -p 1 -v
+
+      - name: Run unit tests
+        run: |
+          go test ./internal/... -v
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -98,4 +98,4 @@ jobs:
           DOCKER_HOST: ${{ matrix.docker_host }}
         run: |
           docker build -t topo-e2e-target:latest ./internal/testutil/test-container
-          go test ./... -v -covermode=atomic -coverprofile=coverage.out
+          go test ./... -p 1 -v -covermode=atomic -coverprofile=coverage.out


muchzill4:
Can you provide more detail about suspected race conditions?

.github/workflows/ci.yaml
```diff
@@ -98,4 +98,4 @@ jobs:
           DOCKER_HOST: ${{ matrix.docker_host }}
         run: |
           docker build -t topo-e2e-target:latest ./internal/testutil/test-container
-          go test ./... -v -covermode=atomic -coverprofile=coverage.out
+          go test ./... -p 1 -v -covermode=atomic -coverprofile=coverage.out
```
muchzill4: If we do this, we'd benefit from splitting internal and e2e test runs. We could still run `./internal` in parallel and use `-p 1` only for `go test ./e2e`.

I suspect we can drop coverage too, so no need to worry about merging coverages.

awphi: Ah nice, I hadn't realized all our e2e tests were under the e2e package now - that wasn't the case last time I checked

muchzill4: I certainly hope that's the case, but depends on what you define as e2e test. 👍 I think there's some docker spawning happening in `deploy_test.go`.

awphi: Split the e2e and unit test job steps https://github.com/arm/topo/pull/118/commits/b27f476c8675e11d627ad7f2f1f554e707631127 

Explicitly not running unit and e2e tests in parallel with eachother to avoid shared docker state weirdness, although we could split them into separate jobs

awphi:
> Can you provide more detail about suspected race conditions?

Answered in DMs but posting here for posterity.

The underlying problem for this e2e failure seems to me that docker push from the local registry will sometimes resolve localhost to ipv6 [::1] instead of ipv4 127.0.0.1 https://github.com/arm/topo/actions/runs/23134377867/job/67194561800

```
        	            	┌─ Transfer via registry ───────────────────────────────
        	            	Using default tag: latest
        	            	The push refers to repository [localhost:12737/project-hello-server]
        	            	351f952fa185: Preparing
        	            	482a2828fed2: Preparing
        	            	fa1d322086e6: Preparing
        	            	5184c31ad48f: Preparing
        	            	b0c20bcd44fe: Preparing
        	            	a257f20c716c: Preparing
        	            	b0c20bcd44fe: Waiting
        	            	a257f20c716c: Waiting
        	            	482a2828fed2: Pushed
        	            	fa1d322086e6: Pushed
        	            	351f952fa185: Pushed
        	            	b0c20bcd44fe: Retrying in 5 seconds
        	            	a257f20c716c: Retrying in 5 seconds
                                 ...
        	            	dial tcp [::1]:12737: connect: connection refused
        	            	Exit request sent.
        	            	ERROR: failed to execute docker push localhost:12737/project-hello-server: exit status 1
```

I've never seen this locally nor has anyone else afaik so my assumption is there's something about running the e2e in parallel that causes this to happen since otherwise the failures would be consistent. As I say though it's just a suspicion so I was hoping to the use the -p 1 change as a tool to narrow this down over the next week or so.

.github/workflows/ci.yaml
awphi: aaaand flipped the order :) https://github.com/arm/topo/pull/118/commits/d31946dda5f18ca83ed18c491e5fea682ad82396

muchzill4 (approved):

awphi:
Sorry @muchzill4, appreciate the review but after digging in response to your comments I've spotted that all our e2e tests (or rather tests that start a target container) are not in the e2e package. I'm going to close this PR and raise another one that moves these tests + corresponding utils into the correct package, rendering `-p 1` useless.

PR #119, opened 2026-03-17
feat: add ssh connection timeout

muchzill4:
## Changes

- Added a configurable connection timeout for SSH operations.
   - A 5-second default timeout is applied to all SSH commands, with a user-friendly error message displayed on a connection timeout.
   - If the user has ConnectTimeout set in their SSH config, that value takes precedence over the default.

Screenshot 2026-03-17 at 14 15 46

---

As a followup PR, I'd like to make this configurable via `--timeout` parameter. I think 5s is a sensible default to start with.

### DIFF
--- a/internal/ssh/config.go
+++ b/internal/ssh/config.go
@@ -4,13 +4,16 @@ import (
 	"bufio"
 	"bytes"
 	"os/exec"
+	"strconv"
 	"strings"
+	"time"
 )
 
 type Config struct {
-	host string
-	user string
-	port string
+	host           string
+	user           string
+	port           string
+	connectTimeout time.Duration
 }
 
 func NewConfig(destination string) Config {
@@ -36,7 +39,19 @@ func NewConfigFromBytes(data []byte) Config {
 			config.user = fields[1]
 		case "port":
 			config.port = fields[1]
+		case "connecttimeout":
+			if secs, err := strconv.Atoi(fields[1]); err == nil {
+				config.connectTimeout = time.Duration(secs) * time.Second
+			}
 		}
 	}
 	return config
+}
+
+// ConnectTimeout returns the user's configured ConnectTimeout if set, otherwise the fallback.
+func (c Config) ConnectTimeout(fallback time.Duration) time.Duration {
+	if c.connectTimeout > 0 {
+		return c.connectTimeout
+	}
+	return fallback
 }
--- a/internal/target/connection.go
+++ b/internal/target/connection.go
@@ -138,6 +138,17 @@ var (
 	ErrAuthenticationFailure = errors.New("ssh authentication failed")
 )
 
+type ConnectionTimeoutError struct {
+	Timeout time.Duration
+}
+
+func (e ConnectionTimeoutError) Error() string {
+	if e.Timeout > 0 {
+		return fmt.Sprintf("ssh connection timed out after %s", e.Timeout)
+	}
+	return "ssh connection timed out"
+}
+
 func (c *Connection) isPasswordAuthenticated() (bool, error) {
 	var extraArgs []string
 	if c.opts.AcceptNewHostKeys {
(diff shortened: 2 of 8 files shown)


internal/ssh/config.go
```diff
@@ -4,13 +4,16 @@ import (
...
 )
 
 type Config struct {
-	host string
-	user string
-	port string
+	host           string
+	user           string
+	port           string
+	connectTimeout time.Duration
```
yejseo01, line 16: `timeout` would be less confusing and accurate name for this attribute

muchzill4: Can you explain what's confusing or inaccurate about `connectTimeout`?

I think `timeout` on its own is odd, "timeout of what exactly?" Also, the current name follows ssh's naming convention.

yejseo01: There are multiple difference usage of the phrase connectTimeout/ConnectTimeout in this PR. `ConnectTimeout()` function which retrieves the `connectTimeout` attribute, `sshConnectTimeout` which is a constant, `connectTimeout` which is the time duration of the timeout and `ConnectTimeout` as a different attribute name of struct `target.ConnectionOptions`

It would be less confusing if either this `connectTimeout` is changed to `timeout`, or the function `ConnectTimeout()` becomes `GetConnectTimeout()`. `sshConnectTimeout` maybe can be changed to `sshConnectTimeoutValue` to represent that it's a constant?

yejseo01: `sshConnectTimeout` -> `timeoutDuration` ?

muchzill4: > There are multiple difference usage of the phrase connectTimeout/ConnectTimeout in this PR.

These are in different packages and contexts.
- `sshConnectTimeout` constant is in `cmd` package has `ssh` prefix, because we want to specifically say that it's a timeout for ssh connections
- `connectTimeout` and `ConnectTimeout()` are in `ssh` package, that's hopefully obvious why prefix isn't required
- the only valid argument I see is to potentially rename `ConnectionOptions.ConnectTimeout` in `target` package, we could call it `SSHConnectTimeout` 🤷 

> It would be less confusing if either this connectTimeout is changed to timeout, or the function ConnectTimeout() becomes GetConnectTimeout()
                                                                                                                           
`GetConnectTimeout()` is a Java-ism, Go avoids the Get prefix for accessors. [You can learn more in Effective Go](https://go.dev/doc/effective_go#Getters).

---

I'd rather keep `connectTimeout` since it intentionally mirrors ssh's own `ConnectTimeout` option.

`timeoutDuration` loses the "connect" specificity entirely, and `sshConnectTimeoutValue` is unusual (why `Value`?).

yejseo01: If we change to `sshConnectTimeout` -> `connectTimeoutDuration`, I'll be happy.

yejseo01 (approved):
Looks good to me besides from suggested constant name change

internal/ssh/config.go
muchzill4: I think I've already explained why `sshConnectTimeout` makes sense. If anything we'd want `sshConnectTimeoutDuration`, but I don't see a reason to use that name, all it does is encode the type in the variable name, which is an anti pattern[^1].

[^1]: Guilty as charged for using `host` everywhere with `ssh.Host`, though `target` was already taken, so I'm excused.

internal/target/connection.go
```diff
@@ -176,5 +197,8 @@ func (c *Connection) runSSHAuthenticationProbe(sshArgs []string) error {
 	if strings.Contains(output, "permission denied") || strings.Contains(output, "authentication failed") || strings.Contains(output, "password") {
 		return ErrAuthenticationFailure
 	}
+	if strings.Contains(output, "timed out") || strings.Contains(output, "connection timeout") || strings.Contains(output, "did not properly respond after a period of time") {
```
th3james, line 200: This seems a little fragile, and apparently there's a non-zero risk these strings might be localised on Linux and Windows.
I wonder if enforcing this in Go with `context.WithTimeout` might be the simpler, more reliable option?

muchzill4: Unsure if simpler, because context will need to be passed from cli down to `cmd.Exec` somehow, which right now is at the bottom of the call stack. (I have some ideas brewing on how to make this less bad, which are going to be required for proper catch-all timeout).

More reliable for sure, you raise a good point with localisation issues, but that ship has sailed with all the `string.Contains` we already have.

🤔

PR #121, opened 2026-03-17
refactor: Rename SSH connection structs accurately

yejseo01:
## Changes

- Config is renamed to SSHAlias to make it more specific
- Host is renamed to SSHDestination. Host actually includes user and port, so this was not an accurate description

### DIFF
--- a/internal/deploy/docker/operation/docker.go
+++ b/internal/deploy/docker/operation/docker.go
@@ -11,11 +11,11 @@ import (
 
 type Docker struct {
 	description string
-	host        ssh.SSHDestination
+	host        ssh.Destination
 	args        []string
 }
 
-func NewDocker(description string, h ssh.SSHDestination, args []string) *Docker {
+func NewDocker(description string, h ssh.Destination, args []string) *Docker {
 	return &Docker{
 		description: description,
 		host:        h,
@@ -44,17 +44,17 @@ func (d *Docker) buildCommand() *exec.Cmd {
 	return command.Docker(d.host, d.args...)
 }
 
-func NewDockerPull(host ssh.SSHDestination, image string) *Docker {
+func NewDockerPull(host ssh.Destination, image string) *Docker {
 	description := fmt.Sprintf("Pull image %s", image)
 	return NewDocker(description, host, []string{"pull", image})
 }
 
-func NewDockerStart(host ssh.SSHDestination, container string) *Docker {
+func NewDockerStart(host ssh.Destination, container string) *Docker {
 	description := fmt.Sprintf("Start container %s", container)
 	return NewDocker(description, host, []string{"start", container})
 }
 
-func NewDockerRun(host ssh.SSHDestination, image string, container string, dockerArgs []string) *Docker {
+func NewDockerRun(host ssh.Destination, image string, container string, dockerArgs []string) *Docker {
 	description := fmt.Sprintf("Run image %s as container %s", image, container)
 	allArgs := []string{"run"}
 	allArgs = append(allArgs, dockerArgs...)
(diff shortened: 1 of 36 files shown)


internal/ssh/dest.go
muchzill4, line 1: We don't need the `ssh_` prefix in the filename. It already lives in the `ssh` directory. Also, let's use the full name: `destination`. It's cheap to do so here.

internal/ssh/ssh_dest.go
```diff
@@ -7,29 +7,29 @@ import (
 	"unicode"
 )
 
-type Host string
+type SSHDestination string
```
muchzill4: We don't need to prefix `SSH`, it's already in `ssh` package. For external consumers it's redundant to say: `ssh.SSHDestination`, `ssh.Destination` is enough.

internal/ssh/ssh_dest.go
```diff
@@ -7,29 +7,29 @@ import (
...
-const PlainLocalhost = Host("localhost")
+const PlainLocalhost = SSHDestination("localhost")
 
-func (h Host) IsPlainLocalhost() bool {
+func (h SSHDestination) IsPlainLocalhost() bool {
 	return strings.EqualFold(string(h), "localhost") || h == "127.0.0.1"
 }
 
-func (h Host) IsLocalhost() bool {
+func (h SSHDestination) IsLocalhost() bool {
```
muchzill4: Can you rename `h` do `d` in every method?

internal/deploy/docker/operation/docker.go
```diff
@@ -11,11 +11,11 @@ import (
 
 type Docker struct {
 	description string
-	host        ssh.Host
+	host        ssh.SSHDestination
```
muchzill4: This essentially becomes `-H` in `docker -H` invocation, so calling it `host` makes sense. 

Though we discussed introducing a docker level type in a subsequent PR, which will make this more explicit and hold `docker-land` specific value (it needs to be a uri specifically, not any ssh destination).

muchzill4 (changes requested):
I think outside of docker-land, the variables that hold `ssh.Destination` should be called `dest`, `destination` or `remote` (pick your poison). Naming it `host` is the same lie we're trying to fix in this PR, by renaming the structs.

internal/ssh/ssh_dest.go
yejseo01: https://github.com/arm/topo/pull/121/commits/5ec5619c7659d822319f2744a66fd461e1242c28

internal/ssh/dest.go
yejseo01: https://github.com/arm/topo/pull/121/commits/5ec5619c7659d822319f2744a66fd461e1242c28

internal/ssh/ssh_dest.go
yejseo01: https://github.com/arm/topo/pull/121/commits/5ec5619c7659d822319f2744a66fd461e1242c28

muchzill4 (commented):
Still a bunch of `h` variable names where it should be `d` instead. 😉

yejseo01:
> Still a bunch of `h` variable names where it should be `d` instead. 😉

I cheekily left the docker host structs as `h`, as I would be changing it back on the next PR, but I guess it's not great to have dependency in style between PRs. Will change the other `h`s

yejseo01:
> Still a bunch of `h` variable names where it should be `d` instead. 😉

Updated the relevant variable names excluding anything docker as we discussed

muchzill4 (approved):
Nice!

PR #122, opened 2026-03-17
test: move all e2e tests into a single package

awphi:
## Background

#118 attempted to fix some of the e2e test failures by running the e2e in series. However, it was noted we'd like to still run the normal unit test packages in parallel. This is tricky since 2 of our e2e tests (that is, tests that start a container to serve as a topo target) don't live in the `e2e` test package.

Thus to avoid a confusing ci.yaml I've created this PR.

## Changes

- Moves the two tests which require a target D-in-D container into the e2e package
- Created a e2e/testutil package to hold e2e-specific test utils and re-export some global ones, similar to the existing docker/testutil package
- Moved the test-container Dockerfile into e2e testdata

Note there is no new code in this PR, just a bunch of code moving around. The net result is all tests that start a target container are in the `e2e` package and thus run in series as per the default behaviour of `go test`, which _might_ address the flakiness noted in #118 - we will need to monitor and see.

I'm a little bit unhappy with the fact these two tests I've moved aren't really e2e tests in that they test internal structs/methods of the application. We could convert them to e2e tests if this is not acceptable but I'd like to do that in a follow-up to separate moving old code from writing new code.


muchzill4 (commented):
I'm not a fan of this reorg. 

It feels like a docker issue in CI is forcing us to structure our folders this specific way. 

I'm starting to wonder if we shouldn't just `-p 1` everything and learn to deal.

Alternatively, I'd be in favour of leveraging go's build tags or mutexing the container creation if `os.GetEnv("CI")` is true.

WDYT?

awphi:
> It feels like a docker issue in CI is forcing us to structure our folders this specific way.

Agreed

> I'm starting to wonder if we shouldn't just -p 1 everything and learn to deal.

As discussed offline let's just go with this - https://github.com/arm/topo/pull/123

PR #124, opened 2026-03-18
refactor: nil-able registry config drives registry enablement

muchzill4:
## Changes

- Replaces the flat registry fields on `DeployOptions` with a single nullable `*RegistryConfig` 
   - A `nil` `Registry` now means "no registry" while a non-`nil` value improves the coherency of associated fields


awphi (approved):
Much cleaner!

PR #127, opened 2026-03-18
refactor: always resolve ssh destination

yejseo01:
## Changes

- Always resolve ssh destination as `ssh -G` command will take non alias as inputs as well. This will reduce some code

### DIFF
--- a/internal/ssh/ssh_tunnel.go
+++ b/internal/ssh/ssh_tunnel.go
@@ -254,16 +254,3 @@ func resolveHost(raw string) string {
 	config := NewConfig(raw)
 	return config.host
 }
-
-func isExplicitHost(raw string) bool {
-	if strings.HasPrefix(raw, "ssh://") {
-		return true
-	}
-	if strings.Contains(raw, "@") || strings.Contains(raw, ":") {
-		return true
-	}
-	if strings.HasPrefix(raw, "[") {
-		return true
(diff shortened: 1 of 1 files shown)


yejseo01:
this change is on top of this PR - https://github.com/arm/topo/pull/121
merge [this one](https://github.com/arm/topo/pull/121) first

internal/ssh/ssh_tunnel.go
```diff
@@ -251,24 +251,6 @@ func (s *SSHTunnelProcessStop) DryRun(w io.Writer) error {
 }
 
 func resolveHost(raw string) string {
-	if raw == "" || isExplicitHost(raw) {
-		_, host, _ := SplitUserHostPort(raw)
```
muchzill4, line 255: Is there a way we can get rid of `SplitUserHostPort`? I think we could use the same mechanism everywhere as `ssh -G` will yield `user`, `host` and `port` (don't take my word for it, validate!).

Can also be done as a separate PR. 🚀

muchzill4 (approved):
Nice!

internal/ssh/ssh_tunnel.go
yejseo01: I think if we use NewConfig, we can. Will make a separate PR for that

PR #128, opened 2026-03-19
refactor: remove useless `GenerateTargetDescription` proxy

muchzill4:
## Changes

- Removes a one-liner wrapping `conn.ProbeHardware()` with no additional logic. Call sites now invoke `ProbeHardware` directly.


yejseo01 (approved):
Great find!

PR #130, opened 2026-03-19
refactor: move binary name validation to `command` package

muchzill4:
## Changes

- Moved binary validation into `internal/command` (it's not an `ssh` concern)
- Moved and updated tests to validate through `ValidateBinaryName`, not through exported `BinaryRegex`
    - Made regex unexported (`binaryRegex`)

### DIFF
--- a/internal/command/validate.go
+++ b/internal/command/validate.go
@@ -1,0 +1,15 @@
+package command
+
+import (
+	"fmt"
+	"regexp"
+)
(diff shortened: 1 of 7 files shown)


federicobozzini (commented):
I actually insisted for this change when this code was added. Happy to see if happening.

federicobozzini (approved):

PR #132, opened 2026-03-19
refactor: Remove `SplitUserHostPort()`

yejseo01:
## Changes

- Remove usage of `SplitUserHostPort()` function
- Use `Config`'s Host, `User`, `Port` instead after making them public
- fix tests so that it uses `ssh://` prefix when specifying the port to comply with ssh target input from `man ssh`

### DIFF
--- a/internal/setupkeys/sshconfig/ssh_config.go
+++ b/internal/setupkeys/sshconfig/ssh_config.go
@@ -100,22 +100,19 @@ func hasIncludeLine(data []byte, includeLine string) bool {
 }
 
 func buildSSHConfigFragment(targetHost string, privKeyPath string) string {
-	user, host, port := ssh.SplitUserHostPort(targetHost)
-	hostAlias := host
-	if hostAlias == "" {
-		hostAlias = targetHost
-	}
+	config := ssh.NewConfig(targetHost)
+	hostAlias := config.Host
 
 	var b strings.Builder
 	fmt.Fprintf(&b, "Host %s\n", hostAlias)
-	if host != "" && (strings.Contains(targetHost, "@") || strings.Contains(targetHost, ":")) {
-		fmt.Fprintf(&b, "  HostName %s\n", host)
+	if config.Host != "" && (strings.Contains(targetHost, "@") || strings.Contains(targetHost, ":")) {
+		fmt.Fprintf(&b, "  HostName %s\n", config.Host)
 	}
-	if user != "" {
-		fmt.Fprintf(&b, "  User %s\n", user)
+	if config.User != "" {
+		fmt.Fprintf(&b, "  User %s\n", config.User)
 	}
-	if port != "" {
-		fmt.Fprintf(&b, "  Port %s\n", port)
+	if config.Port != "" {
+		fmt.Fprintf(&b, "  Port %s\n", config.Port)
 	}
 
 	// needs to be this way even on Windows to work with ssh config parsing, which generally accepts forward slashes
--- a/internal/ssh/config_test.go
+++ b/internal/ssh/config_test.go
@@ -16,7 +16,7 @@ port 2222
 
 		got := NewConfigFromBytes(input)
 
-		want := Config{host: "springfield.nuclear.gov", user: "homer", port: "2222"}
+		want := Config{HostName: "springfield.nuclear.gov", User: "homer", Port: "2222"}
 		assert.Equal(t, want, got)
 	})
 
@@ -28,7 +28,7 @@ user homer
 
 		got := NewConfigFromBytes(input)
 
-		want := Config{host: "springfield.nuclear.gov", user: "homer"}
+		want := Config{HostName: "springfield.nuclear.gov", User: "homer"}
 		assert.Equal(t, want, got)
 	})
 
@@ -47,7 +47,7 @@ Port 22
 
 		got := NewConfigFromBytes(input)
 
-		want := Config{host: "kwik.e.mart", user: "apu", port: "22"}
+		want := Config{HostName: "kwik.e.mart", User: "apu", Port: "22"}
 		assert.Equal(t, want, got)
 	})
 
--- a/internal/ssh/ssh_tunnel.go
+++ b/internal/ssh/ssh_tunnel.go
@@ -66,11 +66,11 @@ func NewSSHTunnelStart(targetDest Destination, port string, useControlSockets bo
 
 func (s *SSHTunnelStart) Command() *exec.Cmd {
 	rawHost := string(s.TargetDest)
-	user, host, port := SplitUserHostPort(rawHost)
-	hostArg := formatSSHHost(rawHost, user, host)
+	config := NewConfig(rawHost)
+	hostArg := config.ConnectionString()
 	args := []string{"ssh", "-N", "-o", "ExitOnForwardFailure=yes"}
-	if port != "" {
-		args = append(args, "-p", port)
+	if s.Port != "" && config.Port != "22" {
+		args = append(args, "-p", s.Port)
 	}
 	if s.UseControlSockets {
 		args = append(args,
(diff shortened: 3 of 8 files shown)


internal/ssh/config_test.go
```diff
@@ -16,7 +16,7 @@ port 2222
 
 		got := NewConfigFromBytes(input)
 
-		want := Config{host: "springfield.nuclear.gov", user: "homer", port: "2222"}
+		want := Config{HostName: "springfield.nuclear.gov", User: "homer", Port: "2222"}
```
muchzill4, line 19: Welp, I just realised I was setting `hostname` as `host` originally, so my bad.

I think we should add `host` entry to the test input as well, because we're missing a field right now.

internal/ssh/ssh_tunnel.go
```diff
@@ -66,11 +52,11 @@ func NewSSHTunnelStart(targetDest Destination, port string, useControlSockets bo
...
 	rawHost := string(s.TargetDest)
-	user, host, port := SplitUserHostPort(rawHost)
-	hostArg := formatSSHHost(rawHost, user, host)
+	config := NewConfig(rawHost)
+	hostArg := config.ConnectionString()
 	args := []string{"ssh", "-N", "-o", "ExitOnForwardFailure=yes"}
-	if port != "" {
-		args = append(args, "-p", port)
+	if config.Port != "" && config.Port != "22" {
+		args = append(args, "-p", config.Port)
```
muchzill4, line 59: Can we not use the `ssh://<host>:<port>` format here? That'd mean we can drop the `-p` mangling.

I think `config.ConnectionString()` should be renamed to `config.Destination()` and return a valid ssh destination.

internal/ssh/ssh_tunnel.go
```diff
@@ -66,11 +52,11 @@ func NewSSHTunnelStart(targetDest Destination, port string, useControlSockets bo
 
 func (s *SSHTunnelStart) Command() *exec.Cmd {
 	rawHost := string(s.TargetDest)
-	user, host, port := SplitUserHostPort(rawHost)
-	hostArg := formatSSHHost(rawHost, user, host)
+	config := NewConfig(rawHost)
```
muchzill4, line 55: This means we're calling `ssh -G` in tests, which is an undesirable side-effect.

We should either:
1. Inject `ssh.Config` to `SSHTunnelStart` 
2. My preference: ensure `ssh.Destination` is indeed a valid destination that doesn't require further parsing. Gradually establishing that pattern would benefit the codebase greatly - it'd mean that wherever we have `ssh.Destination`, we can _just_ give it to any `ssh` cli command and it will work. In this particular case, I think the way to do it would be to move `config := NewConfig(resolvedTarget)` to the cli layer, where we have the raw value. Then, we could introduce `func (c ssh.Config) Destination() ssh.Destination`.
Let's reap the benefits of static typing! Right now `ssh.Destination` is only a string wrapper that has a bunch of methods. We could make it "a validated string value", which would be oh so good.

internal/ssh/ssh_tunnel.go
```diff
@@ -123,7 +109,7 @@ func NewCheckSSHTunnelSecurity(targetDest Destination, port string) *CheckSSHTun
 
 func (ct *CheckSSHTunnelSecurity) Command() *exec.Cmd {
 	if !ct.TargetDest.IsLocalhost() {
-		host := resolveHost(string(ct.TargetDest))
+		host := resolveHostName(string(ct.TargetDest))
```
muchzill4, line 112: This suggestion assumes `ct.TargetDest` is a validated destination as described in https://github.com/arm/topo/pull/132/changes#r2964562908. We could then _just_ do `ct.TargetDest.Host()`. 

Or even better, make `Destination` a proper structure:

```go
type Destination struct {
  User string  
  Host string
  Port string
}

func NewDestination(raw string) (Destination, error) {
   user, host, port, err := parseRawDestination(raw)
   if err != nil {
      return Destination{}, err
   }
   return Destination{user, host, port}, nil
}

// Useful for tests, where we do `ssh.Destination("yolo-swag.com:123")` and don't want to bother handling an error
func MustNewDestination(raw string) Destination {
    dest, err := NewDestination(raw)
    if err != nil {
        panic()
    }
    return dest
}

// Implement Stringer interface, so that we can still `fmt.Sprintf("%s", destination)`
func (d Destination) String() string {
   return fmt.Sprintf("ssh://...") // return safest destination form, which can include all elements of the destination, including port
}
```

Then this becomes:

```
host := ct.TargetDest.HostName
```

And no resolving needs to happen at this point.

internal/setupkeys/sshconfig/ssh_config.go
```diff
@@ -100,22 +100,19 @@ func hasIncludeLine(data []byte, includeLine string) bool {
...
 	var b strings.Builder
 	fmt.Fprintf(&b, "Host %s\n", hostAlias)
-	if host != "" && (strings.Contains(targetHost, "@") || strings.Contains(targetHost, ":")) {
-		fmt.Fprintf(&b, "  HostName %s\n", host)
+	if config.Host != "" && (strings.Contains(targetHost, "@") || strings.Contains(targetHost, ":")) {
+		fmt.Fprintf(&b, "  HostName %s\n", config.Host)
 	}
-	if user != "" {
-		fmt.Fprintf(&b, "  User %s\n", user)
+	if config.User != "" {
```
muchzill4, line 111: I'm not sure if this matters, but `ssh.NewConfig` always populates `config.User`. If it hasn't been specified explicitly in `--target foo` call, it will be likely set to current user's username, as returned by `ssh -G`.

Thus, this will never be empty. I think we should still gate for emptiness. Hopefully this doesn't break anything.

internal/setupkeys/sshconfig/ssh_config.go
```diff
@@ -100,22 +100,19 @@ func hasIncludeLine(data []byte, includeLine string) bool {
...
+		fmt.Fprintf(&b, "  User %s\n", config.User)
 	}
-	if port != "" {
-		fmt.Fprintf(&b, "  Port %s\n", port)
+	if config.Port != "" {
```
muchzill4, line 114: Similarly to `config.User`, this will just be the default SSH port if not specified explicitly.

muchzill4 (changes requested):

internal/ssh/ssh_tunnel.go
yejseo01: for ipv6 format, we can't use `:<port>` suffix, so that is why we are going this direction. I'll change the name of the ConnectionString()

internal/ssh/ssh_tunnel.go
yejseo01: As far as I understand, this is what we concluded is going to be the second PR coming after this one yesterday isn't it?

internal/setupkeys/sshconfig/ssh_config.go
yejseo01: okay. So do I leave it like this?

internal/setupkeys/sshconfig/ssh_config.go
yejseo01: I'm up for getting rid of the if statements around it.

internal/ssh/ssh_tunnel.go
awphi: You can surround the address part of v6 with `[]` to be able to use the port suffix e.g. `[::1]:8080`

internal/ssh/ssh_tunnel.go
yejseo01: this change will be on the following PR, where I would implment calling of the `NewConfig` in highest stack possible

internal/ssh/ssh_tunnel.go
muchzill4: ~Good point. One must wonder if `docker -H` even supports ipv6 strings. I'd rather drop the "partial" ipv6 support right now unless we have it tested e2e. Then we can figure out how to do this properly across the codebase.~

~Actually, I think it's one of the reasons we saw e2e failures - we don't do ipv6 that well.~

Edit: we should listen to @awphi, which also means we can do my original suggestion. 😉

yejseo01:
Not taking this approach anymore

PR #133, opened 2026-03-20
refactor: tighten `Destination` type

muchzill4:
## Background

`Destination` was previously a plain string type alias, which allowed it to be constructed anywhere by casting. This PR converts it to a struct with explicit `User`, `Host`, and `Port` fields, making the type self-contained and eliminating the need to parse the string in multiple places.

## Changes

- `Destination` is now a struct; construction goes through `MustNewDestination(raw string)` (for tests only!) or direct field assignment
- `String()` returns the canonical `ssh://user@host:port` URI form (previously done by `AsURI()`, which now delegates to `String()`)
- `Config` embeds `Destination` directly, replacing its private `host`/`user`/`port` fields
- `NewConfig(target)` is used at command boundaries to parse the raw string; inner packages receive a typed `Destination`
- All call sites updated: `ssh.Destination("...")` casts replaced with `ssh.MustNewDestination("...")`
- `target.NewConnection` now accepts `ssh.Destination` instead of `string`

PR #134, opened 2026-03-23
refactor: less expensive `Destination`

muchzill4:
## Background

As part of https://github.com/arm/topo/pull/133, we've made a poor decision of always leveraging `ssh.Config` obtained by `ssh -G` to calculate a valid `ssh.Destination` from `--target`. 
This PR rolls back that change and leverages `SplitUserHost` to do that instead. Only in place where a legit hostname (not alias) is required, we call to `ssh.Config`.

## Changes

- Introduce `ssh.NewDestination()` and leverage it everywhere a destination is built
   - Deprecate `MustNewDestination()` (poorly named, as it didn't panic)
- Deprecate unused fields from `ssh.Config`


yejseo01:
Have you tried `topo deploy` with this change? I think the curl in the security check will fail at this state

internal/command/command_test.go
```diff
@@ -4,13 +4,13 @@ import (
...
 	"github.com/arm/topo/internal/command"
-	"github.com/arm/topo/internal/testutil"
+	"github.com/arm/topo/internal/ssh"
 	"github.com/stretchr/testify/assert"
 )
 
 func TestString(t *testing.T) {
 	t.Run("converts docker command to string", func(t *testing.T) {
-		h := testutil.MustNewDestination("user@remote")
+		h := ssh.NewDestination("user@remote")
```
yejseo01, line 13:
```suggestion
		dest := ssh.NewDestination("user@remote")
```

internal/command/command_test.go
```diff
@@ -20,7 +20,7 @@ func TestString(t *testing.T) {
 	})
 
 	t.Run("converts docker compose command to string", func(t *testing.T) {
-		h := testutil.MustNewDestination("user@remote")
+		h := ssh.NewDestination("user@remote")
```
yejseo01, line 23: https://github.com/arm/topo/pull/134/changes#r2974537824

internal/ssh/ssh_tunnel.go
yejseo01, line 1: ah I see how it would work now.

internal/ssh/destination_test.go
```diff
@@ -8,6 +8,86 @@ import (
...
+				raw:  "ssh://user@example.com:2222",
+				want: ssh.Destination{User: "user", Host: "example.com", Port: "2222"},
+			},
+			{
+				raw:  "ssh://[2001:db8::1]",
+				want: ssh.Destination{Host: "2001:db8::1"},
+			},
+			{
+				raw:  "ssh://[2001:db8::1]:2222",
+				want: ssh.Destination{Host: "2001:db8::1", Port: "2222"},
```
yejseo01, line 47: nit: more fun examples?

internal/ssh/destination_test.go
```diff
@@ -8,6 +8,86 @@ import (
...
+				want: ssh.Destination{User: "user", Host: "example.com", Port: "2222"},
+			},
+			{
+				raw:  "[2001:db8::1]",
+				want: ssh.Destination{Host: "2001:db8::1"},
+			},
+			{
+				raw:  "user@[2001:db8::1]:2222",
+				want: ssh.Destination{User: "user", Host: "2001:db8::1", Port: "2222"},
+			},
```
yejseo01, line 79: https://github.com/arm/topo/pull/134/changes#r2974563259

internal/ssh/ssh_tunnel_test.go
```diff
@@ -116,7 +116,7 @@ func TestSSHTunnelStart(t *testing.T) {
 
 	t.Run("Description", func(t *testing.T) {
 		t.Run("it returns expected string", func(t *testing.T) {
-			st := ssh.NewSSHTunnelStart(testutil.MustNewDestination("user@remote"), operation.DefaultRegistryPort, true)
+			st := ssh.NewSSHTunnelStart(ssh.NewDestination("user@remote"), operation.DefaultRegistryPort, true)
```
yejseo01, line 119: nit: do you know why it's called `st`?

If it's trying to represent TunnelStart like other lines of code, it should be `ts`?

yejseo01: Can we just name it start?

internal/ssh/ssh_tunnel_test.go
```diff
@@ -165,7 +165,7 @@ func TestCheckSSHTunnelSecurity(t *testing.T) {
 
 	t.Run("Description", func(t *testing.T) {
 		t.Run("it returns the expected string", func(t *testing.T) {
-			cs := ssh.NewCheckSSHTunnelSecurity(testutil.MustNewDestination("user@remote"), operation.DefaultRegistryPort)
+			cs := ssh.NewCheckSSHTunnelSecurity(ssh.NewDestination("user@remote"), operation.DefaultRegistryPort)
```
yejseo01, line 168: and `checkSecurity` here?

internal/ssh/ssh_tunnel_test.go
```diff
@@ -205,7 +205,7 @@ func TestSSHTunnelStop(t *testing.T) {
 
 	t.Run("Description", func(t *testing.T) {
 		t.Run("it returns expected string", func(t *testing.T) {
-			st := ssh.NewSSHTunnelStop(testutil.MustNewDestination("user@remote"))
+			st := ssh.NewSSHTunnelStop(ssh.NewDestination("user@remote"))
```
yejseo01, line 208: `stop` would be good here. We are using `st` for start tunnel as well.

yejseo01 (commented):
Looks good in general, minot nits

internal/ssh/ssh_tunnel_test.go
muchzill4: I'd rather not fix this as part of this PR. Seems like a separate issue.

internal/ssh/destination_test.go
muchzill4: What do you mean?

yejseo01: It was a suggestion on whether we want something like:

`ssh://ironman@avengers.com`

instead of `user@examples.com` everywhere
but it doesn't matter if we don't

muchzill4: Ah. I see. Well, because ipv6 I think we'll keep this less _fun_. 😉

yejseo01 (approved):

federicobozzini (approved):

PR #135, opened 2026-03-23
feat!: setup-ssh command

awphi:
## Background 

https://github.com/Arm-Debug/vscode-topo/pull/236 plans to change the storage mechanism for topo targets in the vscode extension to ssh config files. By default we will only display the files we take ownership over in `~/.ssh/topo_config/*`.

Since the extension needs the ability to 'create' new targets in a way that is compatible with setup-keys and any other ssh config modification we may do in the future it makes sense to do this in the CLI. As long as our ssh config manipulation is done in one place and is consistent (even if consistently wrong) it should be easy to change and understand.

## Changes

- Introduces a new, hidden, command; `setup-ssh` - this performs pretty much the same thing as setup-keys without support for either dry-running or provisioning/transferring keys
- Small refactor of the `sshconfig` package to support arbitrary directives or none at all to support this work
  - Also moves it from `setupkeys/sshconfig` to `ssh/sshconfig`


cmd/topo/setup_ssh.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/spf13/cobra"
+)
+
+var setupSSHCommand = &cobra.Command{
+	Use:   "setup-ssh",
+	Short: "Create a topo-managed SSH config entry for the target",
+	Long: `Create a topo-managed SSH config entry for the target in ~/.ssh/topo_config.
+	
+This will also update the main SSH config (~/.ssh/config) to include the topo-managed configs, if not already included.`,
+	Hidden: true,
```
awphi: hidden since it's not very useful outside of usage by vscode

there's a case to be made we should unit setup-keys and setup-ssh under one command but should be done in a separate refactoring follow up imo

internal/ssh/sshconfig/ssh_config.go
awphi: this dry run log runs the risk of being inaccurate when used for non-setup-keys callers of this function since directives can be arbitrary.

worth solving here or considering deleting dry run as we've previously discussed?

cmd/topo/setup_ssh.go
muchzill4: For commands like this - hidden, vscode specific, we have `cmd/topo/vscode.go`, though it'll break the "one file = one command" pattern. 🤷

cmd/topo/setup_ssh.go
```diff
@@ -0,0 +1,40 @@
...
+	Args:   cobra.ExactArgs(0),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		cmd.SilenceUsage = true
+
+		targetArg, err := requireTarget(cmd)
+		if err != nil {
+			return err
+		}
+
+		targetSlug := ssh.NewConfig(targetArg).Slugify()
```
muchzill4: ~Wait for https://github.com/arm/topo/pull/134 🙏 it'll make this `Destination` dance less bad.~

Merged.

internal/ssh/sshconfig/ssh_config.go
muchzill4: I think we should aim to deprecate `--dry-run`. Perhaps it's ok to just yolo remove it from all commands and then worry about cleaning up internals.

cmd/topo/setup_keys.go
```diff
@@ -64,7 +65,12 @@ Use --dry-run to see what commands would be executed without actually running th
 			return err
 		}
 
-		return sshconfig.ModifySSHConfig(targetArg, privateKeyPath, targetSlug, dryRun, os.Stdout)
+		directives := []sshconfig.SSHConfigDirective{
+			{Key: "IdentityFile", Value: filepath.ToSlash(privateKeyPath)},
```
muchzill4: `filepath.ToSlash` is too much detail for this layer. Perhaps we could have:

```go
sshconfig.NewDirectiveIdentityFile(path string) SSHConfigDirective // <- normalises path internally, so that caller doesn't have to care
```

cmd/topo/setup_ssh.go
```diff
@@ -0,0 +1,40 @@
...
+
+		if err != nil {
+			return err
+		}
+
+		return sshconfig.ModifySSHConfig(targetArg, targetSlug, false, os.Stdout, nil)
```
muchzill4: I think passing `nil` here is a bit of a design smell. It repurposes the `Modify*` to just do half the work.

I'd much rather have an explicit call, such as: `sshconfig.CreateTopoConfig(targetArg, targetSlug string)`. Heck, we could call `ssh.ModifySSHConfig()` inside of that function and clean up internals afterwards if we deem necessary.

internal/ssh/sshconfig/ssh_config.go
```diff
@@ -125,29 +134,40 @@ func buildSSHConfigFragment(targetHost string, privKeyPath string) []byte {
...
+	}
 	return []byte(b.String())
 }
 
-func mergeOwnedSSHConfigDirectives(existing []byte, privKeyPath string) []byte {
-	identityLine := []byte(fmt.Sprintf("  IdentityFile %s", filepath.ToSlash(privKeyPath)))
-	identitiesOnlyLine := []byte("  IdentitiesOnly yes")
+func mergeOwnedSSHConfigDirectives(existing []byte, directives []SSHConfigDirective) []byte {
 	var merged [][]byte
+	directiveKeys := make(map[string]bool)
```
muchzill4, line 141:
```suggestion
	keysToWrite := make(map[string]bool)
```

Makes it more explicit why we need this.

muchzill4 (changes requested):

internal/ssh/sshconfig/ssh_config.go
awphi: I agree but not sure it belongs in this PR - WDYT?

muchzill4: Definitely out of scope here. 👍

internal/ssh/sshconfig/ssh_config.go
muchzill4: Having re-read my suggestion, I'm not sure if this is better. 🤷

cmd/topo/setup_ssh.go
awphi: Agreed, this function is a little bit of a monolith atm. Initially tried to peel it apart a bit but felt like more effort than it's worth but I'll revisit!

awphi: So I've been playing with this and the fiddly bit is that `Modify*` is more of `CreateOrModify*` or upsert-like operation so I've updated the name accordingly.

> Heck, we could call ssh.ModifySSHConfig() inside of that function and clean up internals afterwards if we deem necessary.

I think this is the most pragmatic way forward for now and it can be cleaned up after the fact. I'm keen to get this in to unblock the vscode work depending on it.

I also gutted the dry-run from setup-keys since it was really irking me that it was now incorrect - lmk if you think I've overstepped.

https://github.com/arm/topo/pull/135/commits/41f708fc023c6a855b8a456b17d3ee87e3d6ea38

cmd/topo/setup_ssh.go
awphi: https://github.com/arm/topo/pull/135/commits/4b23ced1ff0861c498dd4b3e3c522ab96cd636c6

cmd/topo/setup_keys.go
awphi: https://github.com/arm/topo/pull/135/commits/714821aa5caf75b91f921f5c3a7c6dfd76f27b3c

internal/ssh/sshconfig/ssh_config.go
awphi: Tend to agree it's not noticeably clearer to me, have left as-is but please re-open if you get another idea

cmd/topo/setup_ssh.go
awphi: https://github.com/arm/topo/pull/135/commits/55d0b9c98fd3fa90dfe8782ebb159c1a4aee740e

muchzill4 (approved):
LGTM, thanks for addressing comments.

muchzill4:
We should probably make it `feat!:` as it's breaking (`--dry-run` deprecation).

PR #136, opened 2026-03-24
refactor: extract auth probe from `Connection`

muchzill4:
## Changes

- Introduces `SSHAuthenticationProbe` that encapsulates the logic for detecting whether a host uses public-key vs. password authentication, and whether host key verification passes                                                                                                                                                                           
    - Slims down `Connection`'s responsibilities significantly by delegating auth probing to the new type
- Removes `connection_auth_test.go` - its tests are migrated to `ssh_authentication_probe_test.go`


tgonzalezorlandoarm (approved):

internal/health/status_test.go
```diff
@@ -49,16 +47,13 @@ func TestProbeHealthStatus(t *testing.T) {
 
 	t.Run("probe succeeds when no remoteproc support", func(t *testing.T) {
 		mockExec := func(_ ssh.Destination, command string, _ []byte, _ ...string) *exec.Cmd {
-			switch command {
-			case "true":
+			if command == "true" {
```
Luke-Parkin, line 50: _chef's kiss_

internal/target/connection.go
```diff
@@ -93,6 +68,20 @@ func (c *Connection) Run(cmdStr string) (string, error) {
...
 
+func (c *Connection) RunWithArgs(cmdStr string, sshArgs ...string) (string, error) {
+	allArgs := append(c.connectTimeoutArgs(), sshArgs...)
+	cmd := c.exec(c.SSHTarget, cmdStr, nil, allArgs...)
+	out, err := cmd.CombinedOutput()
+	if err == nil {
+		return string(out), nil
+	}
+	output := strings.ToLower(string(out))
+	if strings.Contains(output, "timed out") || strings.Contains(output, "connection timeout") || strings.Contains(output, "did not properly respond after a period of time") {
```
Luke-Parkin, line 79: I have no idea if this is something we even support/want to consider- but.. localisation?

Also if we use different SSH clients this'll blow up. I'm less concerned about this, I think your 'did not properly' and the other two cover windows and macos/linux? But is someones ssh client being an in alternative language an issue? Can we check exit codes instead?

internal/target/ssh_authentication_probe.go
```diff
@@ -0,0 +1,111 @@
...
+	return nil
+}
+
+func (p SSHAuthenticationProbe) isPasswordAuthenticated() (bool, error) {
+	var extraArgs []string
+	if p.opts.AcceptNewHostKeys {
+		extraArgs = acceptNewHostKeyArgs
+	}
+
+	// If public key auth succeeds, the target doesn't require password auth.
```
Luke-Parkin: This feels possibly a bit crowded, control flow is a bit hidden.

Is it better to have a 

`isPublicKeyAuthenticated()`
then 
`isPasswordAuthenticated()` (which just does password)
at the top level, rather than hiding the pubkey auth in here?

internal/target/connection_auth_test.go
```diff
@@ -1,179 +1 @@
 package target_test
```
Luke-Parkin: empty file?

internal/target/connection.go
muchzill4: I 100% agree about this being an issue. Having said that, it's outside of the scope here - this PR's remit is lift and shift.

There's more places where what you're describing is an issue.

Luke-Parkin (changes requested):

internal/target/connection.go
Luke-Parkin: Agreed

Luke-Parkin:
Good tests and a really good refactor!!

internal/target/connection_auth_test.go
muchzill4: https://github.com/arm/topo/pull/136/commits/d50fa1da6e1ebe908947b4ba47007733c86f3777

internal/target/ssh_authentication_probe.go
muchzill4: https://github.com/arm/topo/pull/136/commits/7145e9b5d0adfe2a685c0ec95843a6018209ac25

Luke-Parkin (approved):
Nice! TY

PR #139, opened 2026-03-25
ci: update Linux runners to use ubuntu-24.04-arm

federicobozzini:
## Changes

- update Linux runners in CI to use ubuntu-24.04-arm
- Windows runners cannot be updated as easily as the WSL action seems to fail when run on ARM (https://github.com/arm/topo/actions/runs/23535780339/job/68510476002#step:2:42)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -37,7 +37,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix: 
-        os: [ ubuntu-24.04-arm, windows-latest ]
+        os: [ ubuntu-latest, ubuntu-24.04-arm, windows-latest ]
         include:
           - os: windows-latest
             docker_host: ssh://test@localhost:2222
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   lint:
-    runs-on: [ ubuntu-latest ]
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -37,7 +37,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix: 
-        os: [ ubuntu-latest, windows-latest ]
+        os: [ ubuntu-24.04-arm, windows-latest ]
```
awphi: I think we should test on linux on arm and linux on x86 no?

awphi (commented):
One question.

On the WoA note encountered this way back when - https://github.com/actions/partner-runner-images/issues/109

.github/workflows/ci.yaml
federicobozzini: Good point. I think it makes sense if we expect some potential differences between the two architectures. Do we?

awphi: I would given our history of things like compose automatically selecting `platform` based on host architecture if you don't specify `linux/arm` etc.

federicobozzini: Re-added https://github.com/arm/topo/pull/139/commits/215d95957c0097bbc984734f6771e8ec4c2c7411  :+1:

awphi (approved):
Looks good! Worth updating the branch protection rules to include the new test job as a merge blocker

PR #140, opened 2026-03-25
fix: scope template build args per service

muchzill4:
## Changes

- Filter resolved template arguments against each service's declared `build.args` during `extend` so unrelated args are not duplicated across all generated service stubs.

### DIFF
--- a/internal/compose/compose.go
+++ b/internal/compose/compose.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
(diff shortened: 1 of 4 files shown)


tgonzalezorlandoarm (approved):

tgonzalezorlandoarm (changes requested):
testing this... removed Approval

tgonzalezorlandoarm (approved):
tested!

PR #142, opened 2026-03-26
feat!: remove `--dry-run` support

awphi:
## Changes

- Removes the `--dry-run` flag from stop/deploy + updates relevant help strings
- Does not clean up any internals, that will come in a later refactor PR

### DIFF
--- a/cmd/topo/stop.go
+++ b/cmd/topo/stop.go
@@ -16,9 +16,7 @@ var topoStopCmd = &cobra.Command{
 
 Executing this command does not remove the containers.
 
-The compose file (compose.yaml) must be in the current working directory, as this is used to select the containers to be stopped.
-
-Use --dry-run to see what commands would be executed without actually running them.`,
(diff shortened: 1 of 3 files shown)


cmd/topo/stop.go
```diff
@@ -42,16 +34,12 @@ Use --dry-run to see what commands would be executed without actually running th
 		dest := ssh.NewDestination(targetArg)
 
 		stop := docker.NewDeploymentStop(composeFile, dest)
-		if dryRun {
-			return stop.DryRun(os.Stdout)
```
tgonzalezorlandoarm, line 46: ~~Perhaps from another PR, but should we remove the DryRun implementations as well? (and the required interface)~~
scratch that, already clarified

tgonzalezorlandoarm (approved):

muchzill4 (approved):
If we want to start thinking about PR titles == changelog entries, we could perhaps say: "feat!: remove `--dry-run` support" or something along those lines.

PR #143, opened 2026-03-26
feat: Add topo ps command

Luke-Parkin:
## Changes

- Adds a `topo ps` command using the existing docker compose framework, to allow users to view running services of the current compose file. This command works in a way congruent with the stop and deploy.

- vscode already does a similar thing, this moves the functionality into the CLI so CLI users can benefit from it

```
~/demos/topo-lightbulb-moment main *1 ❯ topo ps --target root@10.2.2.39

┌─ List running services ───────────────────────────────
NAME                             IMAGE                          COMMAND                  SERVICE   CREATED       STATUS       PORTS
topo-lightbulb-moment-ollama-1   topo-lightbulb-moment-ollama   "/usr/bin/ollama ser…"   ollama    2 hours ago   Up 2 hours   0.0.0.0:11434->11434/tcp, [::]:11434->11434/tcp
topo-lightbulb-moment-webapp-1   topo-lightbulb-moment-webapp   "uv run python app.py"   webapp    2 hours ago   Up 2 hours   0.0.0.0:3000->3000/tcp, [::]:3000->3000/tcp
topo-lightbulb-moment-zephyr-1   topo-lightbulb-moment-zephyr   "/zephyr.elf"            zephyr    2 hours ago   Up 2 hours
```


tgonzalezorlandoarm:
Didn't we say in planning that this was unmet?

Luke-Parkin:
> Didn't we say in planning that this was unmet?

Well yes. But... it was right there!!

I don't really see any reasons to not do it; due to our existing compose structure it's low hanging fruit.

internal/deploy/docker/ps.go
```diff
@@ -0,0 +1,11 @@
+package docker
+
+import (
+	"github.com/arm/topo/internal/deploy/docker/operation"
+	goperation "github.com/arm/topo/internal/operation"
+	"github.com/arm/topo/internal/ssh"
+)
+
+func NewDeploymentPs(composeFile string, dest ssh.Destination) goperation.Sequence {
```
muchzill4: Is this `DeploymentPs` or just `Ps`? Should we call this `NewListRunningServices` instead? Unsure, just throwing it out there.

Luke-Parkin: Mmm. Yeah. I guess newListRunningServices is a bit more generic, and it's not a child of deployment.

muchzill4:
Is this useful for other consumers w/o `--json`? 🤔

federicobozzini:
> Is this useful for other consumers w/o `--json`? 🤔

No, but I personally don't think this is even particularly useful for the users since it's just a `docker ps`. One of the main points `topo ps` should show is on which subsystems the containers run.

Luke-Parkin:
> No, but I personally don't think this is even particularly useful for the users since it's just a docker ps.

This could be said for nearly all of our commands- And it's not a docker ps, it's a docker compose ps.

This command allows you to use the existing topo infrastructure, target, ssh, etc, to see the containers you've just run. 

So you can do
`topo deploy --target hostname`
And then
`Topo ps --target hostname`

As you did say, you could run the same commands on the remote.. but then the user has to create their own docker compose ps with remote host. I think this PR fixes a relatively big hole in Topo. Without this, Topo is *nearly* all you need to deploy and get started. This pushes it to all.

Luke-Parkin:
> One of the main points topo ps should show is on which subsystems the containers run.

I agree this would be useful, and that we should do it in the future. But should we not put a feature out because we will improve it in the future?

th3james:
@Luke-Parkin could we close this until we're ready to revisit it? There is a problem here that needs solving, but I don't think this is yet the right answer.

PR #144, opened 2026-03-26
refactor: remove unused dry-run implementations

awphi:
## Changes

- Removes `DryRun` from operation interface + corresponding implementations


tgonzalezorlandoarm (approved):
LGTM after formatter is applied

internal/deploy/docker/operation/docker_compose_test.go
```diff
@@ -64,16 +46,6 @@ func TestNewDockerComposeBuild(t *testing.T) {
...
-
-		err := op.DryRun(&buf)
-
-		require.NoError(t, err)
-		want := fmt.Sprintf("docker -H %s compose -f %s build\n", remoteHost.AsURI(), composeFilePath)
-		assert.Equal(t, want, buf.String())
-	})
 }
 
 func TestNewDockerComposePull(t *testing.T) {
```
muchzill4, line 51: Thoughts about the value of these tests now?

`DryRun` was _kind of testing_ the command itself. I wonder if we can just drop this test file altogether, as testing `Descriptions` is probably not that interesting? 🤔 Unless we can come up with a strategy of testing those command wrappers.

muchzill4 (commented):
The LOC change I crave.

internal/deploy/docker/operation/docker_compose_test.go
awphi: `TestDockerCompose` further up in this file does test `Run` so I'm keen to keep the file. I think with the DinD targets it would be quite easy/low-cost to run docker commands to test these wrappers so it's something I'm open to doing.

Not for this PR though imo - this is pure removal. Similar story for `ssh_tunnel_test.go` and `docker_test.go`

muchzill4 (approved):

PR #145, opened 2026-03-27
fix: check password authentication is necessary on deploy and error if so

yejseo01:
## Changes

- check password authentication is necessary on deploy and error if so


yejseo01:
closing this, because I like this resolution wayy better - https://github.com/arm/topo/pull/148

PR #147, opened 2026-03-27
refactor: move stdin from connection to command execution

muchzill4:
## Changes

- Removed `WithStdin` from `target.ConnectionOptions`
- Added per-command stdin support via `Connection.RunWithStdin(...)`

## Rationale

`Connection` is reusable transport/configuration. Stdin is command-specific input.

Keeping stdin on the connection made the API surprising and easy to misuse, because one command’s input could implicitly affect other commands executed through the same connection. This change makes the data flow explicit at the call site and narrows the responsibility of `Connection`.


federicobozzini (approved):
LGTM!

PR #149, opened 2026-03-27
chore: build config fragments from parsed destinations

tgonzalezorlandoarm:
Pass ssh.Destination through SSH config creation instead of reparsing the raw target string inside sshconfig.
This reuses functionality used elsewhere and avoids code duplication.

### DIFF
--- a/internal/ssh/sshconfig/ssh_config.go
+++ b/internal/ssh/sshconfig/ssh_config.go
@@ -3,6 +3,7 @@ package sshconfig
 import (
 	"bytes"
 	"fmt"
+	"net"
 	"os"
 	"path/filepath"
 	"slices"
@@ -34,11 +35,11 @@ func NewDirective(key, value string) SSHConfigDirective {
 	}
 }
 
-func CreateSSHConfig(targetHost string, targetSlug string) error {
-	return CreateOrModifySSHConfig(targetHost, targetSlug, nil)
+func CreateSSHConfig(dest ssh.Destination, targetSlug string) error {
+	return CreateOrModifySSHConfig(dest, targetSlug, nil)
 }
 
-func CreateOrModifySSHConfig(targetHost string, targetSlug string, directives []SSHConfigDirective) error {
+func CreateOrModifySSHConfig(dest ssh.Destination, targetSlug string, directives []SSHConfigDirective) error {
 	home, err := os.UserHomeDir()
 	if err != nil {
 		return fmt.Errorf("failed to determine home directory for SSH config: %w", err)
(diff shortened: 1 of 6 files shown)


internal/ssh/sshconfig/ssh_config.go
```diff
@@ -111,23 +112,17 @@ func hasIncludeLine(data []byte, includeLine string) bool {
...
-		hostAlias = targetHost
-	}
-
+func buildSSHConfigFragment(dest ssh.Destination, directives []SSHConfigDirective) []byte {
 	var b strings.Builder
-	fmt.Fprintf(&b, "Host %s\n", hostAlias)
-	if host != "" && (strings.Contains(targetHost, "@") || strings.Contains(targetHost, ":")) {
-		fmt.Fprintf(&b, "  HostName %s\n", host)
+	fmt.Fprintf(&b, "Host %s\n", dest.Host)
+	if dest.Host != "" && (dest.User != "" || net.ParseIP(dest.Host) != nil) {
```
muchzill4, line 118: Is this right? There's a double `dest.Host != ""`.

muchzill4: Also, `net.ParseIP()` is a different logic to what we had here before I think.

Previously if input was `foo@bar:123`, we'd output `HostName bar`? Unless I'm missing some implicit internals of `SplutUserHostPort`.

muchzill4 (commented):
Nice, thanks for doing this!

Can we now make `SplitUserHostPort` private?

internal/ssh/sshconfig/ssh_config.go
tgonzalezorlandoarm: > Is this right? There's a double dest.Host != ""

I'm missing it :( could you please point me to where I'm duplicating this?

tgonzalezorlandoarm: Regarding the logic, we are still achieving the same behaviour as far as I'm aware?

```
$ cat ~/.ssh/topo_config/topo_tomorl01_imx93.conf 
Host imx93
  HostName imx93
  User tomorl01
  IdentityFile /home/tomgon01/.ssh/id_ed25519_topo_tomorl01_imx93
  IdentitiesOnly yes
```

tgonzalezorlandoarm: and:
```
$ cat ~/.ssh/topo_config/topo_tomorl01_10.2.2.21.conf 
Host 10.2.2.21
  HostName 10.2.2.21
  User tomorl01
  IdentityFile /home/tomgon01/.ssh/id_ed25519_topo_tomorl01_10.2.2.21
  IdentitiesOnly yes
```

tgonzalezorlandoarm: Which case do you think works differently?

muchzill4: > I'm missing it :( could you please point me to where I'm duplicating this?

I can't read. It's `dest.Host` and `dest.User`. Ignore me. 😅

muchzill4 (approved):
🚢 🇮🇹

PR #150, opened 2026-03-27
fix: avoid overwriting HostName for SSH aliases in setup-keys

tgonzalezorlandoarm:
Currently, setup-keys writes a HostName line on the Topo SSH config fragment it generates. This line is written only when @ or : are present in the target definition.

If the user's SSH config is, for example:

```
Host some-board
  HostName <ip>
  User root
```

and we call `topo setup-keys --target otheruser@some-board`, this will create an SSH config fragment like the following:

```
Host some-board
  HostName some-board
  User otheruser
  IdentityFile ...
  IdentitiesOnly yes
```

Which creates a HostName conflict, and we get the following error:

$ ssh some-board
ssh: Could not resolve hostname some-board: nodename nor servname provided, or not known

Fix:

 * Don't write HostName in the fragment in those cases where we are passing down a target thath uses an alias already defined in our effective SSH config.

### DIFF
--- a/internal/ssh/sshconfig/ssh_config.go
+++ b/internal/ssh/sshconfig/ssh_config.go
@@ -3,7 +3,6 @@ package sshconfig
 import (
 	"bytes"
 	"fmt"
-	"net"
 	"os"
 	"path/filepath"
 	"slices"
@@ -115,7 +114,8 @@ func hasIncludeLine(data []byte, includeLine string) bool {
 func buildSSHConfigFragment(dest ssh.Destination, directives []SSHConfigDirective) []byte {
 	var b strings.Builder
 	fmt.Fprintf(&b, "Host %s\n", dest.Host)
-	if dest.Host != "" && (dest.User != "" || net.ParseIP(dest.Host) != nil) {
+	sshConfig := ssh.NewConfig(dest)
+	if dest.Host != "" && (sshConfig.HostName == "" || strings.EqualFold(sshConfig.HostName, dest.Host)) {
 		fmt.Fprintf(&b, "  HostName %s\n", dest.Host)
 	}
 	if dest.User != "" {


tgonzalezorlandoarm:
There's a separate issue that I want to deal with in a different PR:

If I do:
```
./topo setup-keys --target imx93
./topo setup-keys --target tomorl01@imx93
```

I will get:

```
$ cat ~/.ssh/topo_config/topo_tomorl01_adam-imx93.conf
Host imx93
  User tomorl01
  IdentityFile /home/tomgon01/.ssh/id_ed25519_topo_tomorl01_imx93
  IdentitiesOnly yes
```

...and:

```
$ cat ~/.ssh/topo_config/topo_imx93.conf
Host imx93
  IdentityFile /home/tomgon01/.ssh/id_ed25519_topo_imx93
  IdentitiesOnly yes
```

1. The only effective user now is `tomorl01` (as it's the last one), so if I do `ssh imx93` I'll log in to `tomorl01`. This is a problem.
2. Perhaps we want the same conf file to cover all users ?

awphi:
> There's a separate issue that I want to deal with in a different PR:
> 
> If I do:
> 
> ```
> ./topo setup-keys --target imx93
> ./topo setup-keys --target tomorl01@imx93
> ```
> 
> I will get:
> 
> ```
> $ cat ~/.ssh/topo_config/topo_tomorl01_adam-imx93.conf
> Host imx93
>   User tomorl01
>   IdentityFile /home/tomgon01/.ssh/id_ed25519_topo_tomorl01_imx93
>   IdentitiesOnly yes
> ```
> 
> ...and:
> 
> ```
> $ cat ~/.ssh/topo_config/topo_imx93.conf
> Host imx93
>   IdentityFile /home/tomgon01/.ssh/id_ed25519_topo_imx93
>   IdentitiesOnly yes
> ```
> 
> 1. The only effective user now is `tomorl01` (as it's the last one), so if I do `ssh imx93` I'll log in to `tomorl01`. This is a problem.
> 2. Perhaps we want the same conf file to cover all users ?

Also feeling the pain of this in VSCode w/ `topo setup-ssh` which follows mostly the same codepath. I don't think we can support multiple users on the same machine if we re-use the `Host` like we do now because only one `User` directive can ever win out.

I wonder if we can include the user in the `Host` i.e. `Host imx93` and `Host tomorl01@imx93` would be two separate entries

awphi (commented):
Could you add a test for this behaviour?

muchzill4:
I wonder if the correct behaviour here would be to inform the user about the problem and not proceed. I.e.:
"We've detected an existing entry for host FOO in ~/.path/to/config. Remove entry from ssh config before proceeding."

I worry fixing this is a slippery slope. This would be a non issue if we isolated ourselves from ssh config with `-F` of course, but carries its own set of problems.

awphi:
> I wonder if the correct behaviour here would be to inform the user about the problem and not proceed. I.e.: "We've detected an existing entry for host FOO in ~/.path/to/config. Remove entry from ssh config before proceeding."

I'm also amenable to that but it does carry implications for vscode where I wouldn't be able to have adam@foo and root@foo as two separate targets. Not sure why you'd ever do that though so probably okay and definitely better than silently breaking things.

tgonzalezorlandoarm:
> I wonder if we can include the user in the Host i.e. Host imx93 and Host tomorl01@imx93 would be two separate entries

This is an option that I was considering. The problem however would be that the user would not be made aware that the entry is going to actually be `tomorl01-imx93` (or any applicable variants) instead of `imx93`

tgonzalezorlandoarm:
> > I wonder if the correct behaviour here would be to inform the user about the problem and not proceed. I.e.: "We've detected an existing entry for host FOO in ~/.path/to/config. Remove entry from ssh config before proceeding."
> 
> I'm also amenable to that but it does carry implications for vscode where I wouldn't be able to have adam@foo and root@foo as two separate targets. Not sure why you'd ever do that though so probably okay and definitely better than silently breaking things.

Will do!

tgonzalezorlandoarm:
Created #154

PR #153, opened 2026-03-30
chore: Use https git urls for templates

Luke-Parkin:
Use https urls for cloning templates to increase reliability

### DIFF
--- a/e2e/templates_test.go
+++ b/e2e/templates_test.go
@@ -22,7 +22,7 @@ func TestTemplates(t *testing.T) {
 		output := string(out)
 
 		assert.Contains(t, output, "topo-welcome")
-		assert.Contains(t, output, "git@github.com:")
+		assert.Contains(t, output, "https://github.com/")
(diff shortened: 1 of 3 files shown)


federicobozzini (commented):
Can you please also update the URLs used in the `clone` and `extend` docs?

Luke-Parkin:
> Can you please also update the URLs used in the clone and extend docs?

@federicobozzini WDYM?

Do you mean the help commands?
```
  topo clone git@github.com:user/repo.git
  topo clone https://github.com/user/repo.git#develop
  topo clone git:git@github.com:user/repo.git
  topo clone git:https://github.com/user/repo.git#main
  topo clone git:ubuntu@example.com:repo.git
  topo clone git:builder@host:tools/platform.git#v2
```
I don't think this needs changing?

federicobozzini (approved):
Good point, we already cover quite well all the cases. LGTM!

PR #154, opened 2026-03-30
feat: Only allow one user per Host/alias when setting up keys

tgonzalezorlandoarm:
When we use `topo setup-keys --target <target>`, we parse the user's effective SSH config, and create an SSH config fragment under `~/.ssh/topo_config` that sets up the key path (Identity File) for that target.

However, if we later on do `topo setup-keys --target newuser@<target>`, this will create a HostName collision. SSH configs should not have multiple users defind for the same target Host/alias as this would effectively overwrite the user with the one defined last.

 * Error out when the requested Host/alias is already defined in the effective SSH config but for a different user.

### DIFF
--- a/internal/ssh/config.go
+++ b/internal/ssh/config.go
@@ -10,12 +10,14 @@ import (
 )
 
 type Config struct {
-	HostName       string
-	connectTimeout time.Duration
+	HostName            string
+	User                string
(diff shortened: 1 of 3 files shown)


tgonzalezorlandoarm:
Closing in favour of #156

PR #155, opened 2026-03-30
refactor: decouple `ProbeHealthStatus` from `Connection` 

muchzill4:
## Changes

- `ProbeHealthStatus` now accepts a `runner` interface instead of `target.Connection`, returning `HealthStatus` (hardware + dependencies only) — it owns what it discovers, nothing more               
- Auth probe moves into `CheckTarget`, gated on `!dest.IsPlainLocalhost()`, laying the groundwork for localhost targets skipping SSH auth entirely                                                     
- `ConnectionStatus` groups destination and connection error; `CheckTarget` assembles the full `Status` from ConnectionStatus` + `HealthStatus`                                                       
- Tests switch from `WithMockExec` to `testify/mock`

### DIFF
--- a/internal/health/health.go
+++ b/internal/health/health.go
@@ -65,6 +65,17 @@ func CheckHost() HostReport {
 	return GenerateHostReport(dependencyStatuses)
 }
 
+type ConnectionStatus struct {
+	Destination ssh.Destination
+	Error       error
+}
+
+type Status struct {
+	Connection   ConnectionStatus
+	Dependencies []DependencyStatus
+	Hardware     HardwareProfile
+}
+
 func CheckTarget(dest ssh.Destination, probeOpts target.SSHAuthenticationProbeOptions, connectTimeout time.Duration) (TargetReport, error) {
 	opts := target.ConnectionOptions{
 		Multiplex:      true,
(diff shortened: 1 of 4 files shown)


internal/health/health.go
```diff
@@ -84,7 +109,7 @@ func GenerateHostReport(statuses []DependencyStatus) HostReport {
 
 func GenerateTargetReport(targetStatus Status) TargetReport {
 	report := TargetReport{}
-	report.IsLocalhost = targetStatus.SSHTarget.IsPlainLocalhost()
+	report.IsLocalhost = targetStatus.Connection.Destination.IsPlainLocalhost()
```
yejseo01: Isn't this quite coupling to the shape of the `Connection` and `Status`? or is that acceptable within the same file?

muchzill4: It's not ideal (too many dots), though I prefer the fact that `Connection` information is now grouped together.

Do you have a suggestion I could implement instead?

yejseo01: Would getters be javaism even in this case?

muchzill4: I have to guess what you mean, hopefully I got it right: https://github.com/arm/topo/pull/155/commits/28268eeb9601f4fc6d6cbe1716fc16fad9f5dac6

yejseo01 (approved):
LGTM

PR #156, opened 2026-03-30
feat: only allow one user per Host/alias when setting up keys

tgonzalezorlandoarm:
When we use `topo setup-keys --target <target>`, we parse the user's effective SSH config, and create an SSH config fragment under `~/.ssh/topo_config` that sets up the key path (Identity File) for that target.

However, if we later on do `topo setup-keys --target newuser@<target>`, this will create a HostName collision. SSH configs should not have multiple users defined for the same target Host/alias as this would effectively overwrite the user with the one defined last.

 * Error out when the requested Host/alias is already defined in the effective SSH config but for a different user.

### DIFF
--- a/internal/ssh/config.go
+++ b/internal/ssh/config.go
@@ -3,6 +3,7 @@ package ssh
 import (
 	"bufio"
 	"bytes"
+	"fmt"
 	"os/exec"
 	"strconv"
 	"strings"
@@ -11,11 +12,16 @@ import (
 
 type Config struct {
 	HostName       string
+	User           string
 	connectTimeout time.Duration
 }
 
+func readConfig(dest Destination) ([]byte, error) {
+	return exec.Command("ssh", "-v", "-G", dest.String()).CombinedOutput()
+}
+
 func NewConfig(dest Destination) Config {
-	output, err := exec.Command("ssh", "-G", dest.String()).Output()
+	output, err := readConfig(dest)
 	if err != nil {
 		return Config{}
 	}
@@ -26,13 +32,17 @@ func NewConfigFromBytes(data []byte) Config {
 	var config Config
 	scanner := bufio.NewScanner(bytes.NewReader(data))
 	for scanner.Scan() {
-		fields := strings.Fields(scanner.Text())
+		line := scanner.Text()
+
+		fields := strings.Fields(line)
 		if len(fields) < 2 {
 			continue
 		}
 		switch strings.ToLower(fields[0]) {
 		case "hostname":
 			config.HostName = fields[1]
+		case "user":
+			config.User = fields[1]
 		case "connecttimeout":
 			if secs, err := strconv.Atoi(fields[1]); err == nil {
 				config.connectTimeout = time.Duration(secs) * time.Second
@@ -48,4 +58,59 @@ func (c Config) ConnectTimeout(fallback time.Duration) time.Duration {
 		return c.connectTimeout
 	}
 	return fallback
+}
+
+func IsDestinationAlreadyConfiguredWithAnotherUser(dest Destination) error {
+	hostConfig, err := LookupExplicitHostConfig(dest.Host, dest.Port)
+	if err != nil {
+		return err
+	}
+
+	if dest.User != "" && hostConfig.User != dest.User {
+		return fmt.Errorf("ssh host/alias %q is already associated with user %q", dest.Host, hostConfig.User)
+	}
+	return nil
+}
+
+// LookupExplicitHostConfig returns configuration, only if there's an explicit entry for the given host/port
+func LookupExplicitHostConfig(host, port string) (Config, error) {
+	dest := Destination{Host: host, Port: port}
+	output, err := readConfig(dest)
+	if err != nil {
+		return Config{}, err
+	}
+
+	if !isExplicitHostConfig(host, output) {
+		return Config{}, fmt.Errorf("no explicit host config found for %s", host)
+	}
+
+	return NewConfigFromBytes(output), nil
+}
+
+func isExplicitHostConfig(host string, config []byte) bool {
+	const marker = ": Applying options for "
+
+	scanner := bufio.NewScanner(bytes.NewReader(config))
+	for scanner.Scan() {
+		line := scanner.Text()
+
+		if !strings.Contains(line, marker) {
+			continue
+		}
+
+		hostCandidates := strings.FieldsFunc(strings.TrimSpace(line[strings.Index(line, marker)+len(marker):]), func(r rune) bool {
+			return r == ',' || r == ' ' || r == '\t'
+		})
+
+		for _, hostCandidate := range hostCandidates {
+			if hostCandidate == "" || strings.HasPrefix(hostCandidate, "!") || strings.ContainsAny(hostCandidate, "*?") {
+				continue
+			}
+			if strings.EqualFold(hostCandidate, host) {
+				return true
+			}
+		}
+	}
+
+	return false
 }
--- a/internal/ssh/config_test.go
+++ b/internal/ssh/config_test.go
@@ -8,14 +8,16 @@ import (
 )
 
 func TestNewConfigFromBytes(t *testing.T) {
-	t.Run("parses hostname", func(t *testing.T) {
+	t.Run("parses basic config fields", func(t *testing.T) {
 		input := []byte(`hostname springfield.nuclear.gov
+user homer
 `)
 
 		got := NewConfigFromBytes(input)
 
 		want := Config{
 			HostName: "springfield.nuclear.gov",
+			User:     "homer",
 		}
 		assert.Equal(t, want, got)
 	})
@@ -30,6 +32,7 @@ user homer
 
 		want := Config{
 			HostName: "springfield.nuclear.gov",
+			User:     "homer",
 		}
 		assert.Equal(t, want, got)
 	})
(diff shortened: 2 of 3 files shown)


internal/ssh/config_test.go
```diff
@@ -20,6 +20,18 @@ func TestNewConfigFromBytes(t *testing.T) {
 		assert.Equal(t, want, got)
 	})
 
+	t.Run("parses user", func(t *testing.T) {
```
muchzill4: NIT: could merge this with above test and say "parses basic config fields" instead. 🤷

internal/ssh/config.go
```diff
@@ -49,3 +58,28 @@ func (c Config) ConnectTimeout(fallback time.Duration) time.Duration {
 	}
 	return fallback
 }
+
+func (c Config) HasExactHostMatch(host string) bool {
+	for _, pattern := range c.matchedHostPatterns {
```
muchzill4: The patterns stored on the `Config` should not require any further parsing, they should be parsed **completely** when config is first created.

I'd expect the logic of `HasExactHostMatch` to leverage `strings.EqualFold` and `slices.Contains` at most.

internal/ssh/config_test.go
```diff
@@ -71,6 +84,30 @@ connecttimeout none
 
 		assert.Equal(t, time.Duration(0), got.connectTimeout)
 	})
+
+	t.Run("records exact host matches from verbose ssh output", func(t *testing.T) {
+		input := []byte(`debug1: /tmp/config line 1: Applying options for board
```
muchzill4: I assume this is going to work in a non en locale? I appreciate we already have a bunch of string matching in the codebase, but...

muchzill4 (commented):
Can you explain the mechanism we're using here? I find this somewhat hard to understand.

tgonzalezorlandoarm:
> Can you explain the mechanism we're using here? I find this somewhat hard to understand.

We are parsing the effective SSH config as output by `ssh -v -G <target>`
This way, we can check the User that is already set for that target, and if it does not match the user that is being requested, we error with:

As for when we are actually doing this check, we should do this before we actually create any keys, otherwise we are left with some left over keys that we haven't actually used for our config fragment. In #154 I was actually doing the check after the keys were being created.

tgonzalezorlandoarm:
I appreciate I may not be responding to specifically what was confusing you, so please let me know if so @muchzill4

muchzill4:
> We are parsing the effective SSH config as output by ssh -v -G <target>
> This way, we can check the User that is already set for that target

So, in other words, given `--target user@host`, we check if `ssh -G host` yields the same user? I'm trying to understand the information you're using to figure out that a "host" is already taken by a different user.

tgonzalezorlandoarm:
> > We are parsing the effective SSH config as output by ssh -v -G 
> > This way, we can check the User that is already set for that target
> 
> So, in other words, given `--target user@host`, we check if `ssh -G host` yields the same user? I'm trying to understand the information you're using to figure out that a "host" is already taken by a different user.

Yes, we are checking `ssh -G host` for whether the same user (as indicated by `--target user@host`) appears in the effective SSH configuration.
We are adding the `-v` option though so that `ssh -v -G host` will also provide information on where the effective user is coming from.
`applying...` that string will give us the `Host/alias` the effective user comes from

tgonzalezorlandoarm:
With the `-v` option, we get, for example:

```
OpenSSH_8.9p1 Ubuntu-3ubuntu0.14, OpenSSL 3.0.2 15 Mar 2022
debug1: Reading configuration data /home/tomgon01/.ssh/config
debug1: /home/tomgon01/.ssh/config line 1: include /home/tomgon01/.ssh/topo_config/*.conf matched no files
debug1: /home/tomgon01/.ssh/config line 25: Applying options for myimx
debug1: Reading configuration data /etc/ssh/ssh_config
debug1: /etc/ssh/ssh_config line 3: Applying options for *
user root
```

So the host `myimx` is appearing there, and the `user` as well

internal/ssh/config.go
tgonzalezorlandoarm: Done

tgonzalezorlandoarm: Good point

internal/ssh/config_test.go
tgonzalezorlandoarm: Done

internal/ssh/config_test.go
tgonzalezorlandoarm: https://github.com/openssh/openssh-portable/blob/master/readconf.c#L1860

This is hardcoded as far as I'm aware.

muchzill4:
@tgonzalezorlandoarm thanks for clarifying.

I'm still a bit puzzled by the logic here. 😮‍💨 As far as I can tell you can't really gather from the output alone where the `user` comes from. As far as I'm aware `ssh -G` will always provide a value for `user`, which can either be the default (=`whoami`) or set explicitly in matching `Host` block. If I understand correctly, what we want is: "is there an explicit user set for this host that does not match the user we've been given". I don't think it's possible to figure out with `ssh -G` alone, though I hope I'm wrong.

---

edit: I've dropped the code feedback section, as I realised I wasn't making much sense. I'll revisit this once we agree if `ssh -G` works for this.

tgonzalezorlandoarm:
> @tgonzalezorlandoarm thanks for clarifying.
> 
> I'm still a bit puzzled by the logic here. 😮‍💨 As far as I can tell you can't really gather from the output alone where the `user` comes from. As far as I'm aware `ssh -G` will always provide a value for `user`, which can either be the default (=`whoami`) or set explicitly in matching `Host` block. If I understand correctly, what we want is: "is there an explicit user set for this host that does not match the user we've been given". I don't think it's possible to figure out with `ssh -G` alone, though I hope I'm wrong.
> 
> edit: I've dropped the code feedback section, as I realised I wasn't making much sense. I'll revisit this once we agree if `ssh -G` works for this.

As discussed offline, even if the user that is being decided upon is SSH's default user, we should still error out. This is because we should have a per-user alias when setting up keys (identity files should be mapped to specific users). 
Some rewording of the error message needs to happen.

internal/ssh/config.go
```diff
@@ -11,11 +11,16 @@ import (
 
 type Config struct {
 	HostName       string
+	User           string
 	connectTimeout time.Duration
 }
 
+func getSSHEffectiveConfig(dest Destination) ([]byte, error) {
```
muchzill4: Why `Effective`? 🤔 It really is `readConfig()`.

We can also drop `SSH`, it's obvious, we are in `ssh` package.

internal/ssh/config.go
```diff
@@ -49,3 +58,40 @@ func (c Config) ConnectTimeout(fallback time.Duration) time.Duration {
...
+	return IsHostInEffectiveHostListFromBytes(dest, effectiveConfig), nil
+}
+
+func IsHostInEffectiveHostListFromBytes(dest Destination, effectiveConfig []byte) bool {
+	const marker = ": Applying options for "
+
+	scanner := bufio.NewScanner(bytes.NewReader(effectiveConfig))
+	for scanner.Scan() {
+		line := scanner.Text()
+		idx := strings.Index(line, marker)
```
muchzill4: I think `strings.Contains` would read better here.

internal/ssh/sshconfig/ssh_config.go
```diff
@@ -35,6 +35,25 @@ func NewDirective(key, value string) SSHConfigDirective {
...
+		return nil
+	}
+
+	destinationNoUserSpecified := dest
+	destinationNoUserSpecified.User = ""
+	configForHost := ssh.NewConfig(destinationNoUserSpecified)
+
+	isInEffectiveHostList, err := ssh.IsHostInEffectiveHostList(destinationNoUserSpecified)
+	if err != nil {
+		return fmt.Errorf("failed to check if host is in effective host list: %w", err)
```
muchzill4: - This error is surfaced to the user, right? If so, I think it'd be nice to think of a better name for "effective host list" or skip that detail altogether. 

- I tried to come up with an improvements to naming strategy, which was not straightforward.. Named host, declared host, configured host? Anyway, while doing that, I massaged the code a bit and this is what I ended up with. WDYT?

```go
package ssh

func IsDestinationAlreadyConfiguredWithAnotherUser(dest Destination) bool {
   hostConfig, err := LookupExplicitHostConfig(dest.Host, dest.Port)
   if err != nil {
      ...
   }
   return hostConfig.User != dest.User
}

// LookupExplicitHostConfig returns configuration, only if there's an explicit entry for the given host/port
// I'm debating if this should take Destination or not. I like the fact this is now explicit about ignoring user.
func LookupExplicitHostConfig(host, port string) (Config, error) {
   dest := Destination{Host: host, Port: port}
   output, err := readConfig(dest.String())
   if err != nil {
      ...
   }

   if !isExplicitHostConfig(host, output) {
      return fmt.Errorf(...)
   }

   return NewConfigFromBytes(output), nil
}

func isExplicitHostConfig(host string, config []byte) bool {
    // you have this logic already
}
```

I like it not only because it's the code I've written (obviously! 😉), but because it keeps everything contained. I noticed that your proposal still does "de-usering" of the `Destination` twice in seemingly unrelated locations. Also, we call `ssh -G` once, not twice! 💃

muchzill4 (commented):
Thanks for addressing feedback thus far @tgonzalezorlandoarm . 

Your changes improved the legibility and helped the cogs in my head to start spinning and hopefully suggest something meaningful. 👍

internal/ssh/sshconfig/ssh_config.go
muchzill4: BTW: I hope this works, I did not test it. The snippet is a result of coding inside the comment window on GH like a caveman. 🙈

tgonzalezorlandoarm: > BTW: I hope this works, I did not test it.

That's completely fine!
When it's submitting suggestions like this I assume they will need some touching up. When using the actual suggestion format in Github (```suggestion```, etc) I assume that it will work

muchzill4: > When using the actual suggestion format in Github (suggestion, etc) I assume that it will work

And then `gofmt` yells at you for not removing the import. 😉

tgonzalezorlandoarm: Done!

internal/ssh/config_test.go
```diff
@@ -88,3 +91,43 @@ func TestConfigConnectTimeout(t *testing.T) {
...
 }
+
+func TestIsExplicitHostConfig(t *testing.T) {
+	t.Run("returns true for exact host matches in verbose ssh output", func(t *testing.T) {
+		config := []byte(`debug1: /tmp/config line 1: Applying options for Board,board-alt
+debug1: /tmp/config line 5: Applying options for *
+hostname springfield.nuclear.gov
+`)
+
+		got := isExplicitHostConfig("board", config)
```
muchzill4, line 102: Argh. 

Only now I've realised that this test file allows white box testing (it is part of `ssh` package). Everywhere else we try to do black box testing (`package <name>_test`).

Not the end of the world I guess, but this makes things inconsistent. I don't have a good answer here. Perhaps learn to deal with having `IsExplicitHostConfig` exported and update `config_test.go` to be `package ssh_test`?

muchzill4 (approved):

internal/ssh/config_test.go
tgonzalezorlandoarm: Created https://github.com/arm/topo/pull/165

PR #158, opened 2026-03-30
bug: Warn if args are resolved but not found in both extend and clone

Luke-Parkin:
## Changes

`topo clone` reports when an arg is resolved but not found. topo extend, before this change, does not.

This change will mean when `topo extend`ing, a arg that is in x-topo but not found in the service args, a warn will be sent out; same as it does in clone.


yejseo01 (approved):
LGTM. Sounds like a suitable approach for now.

awphi (approved):

PR #159, opened 2026-03-31
fix!: lack of remoteproc endpoints is an info, not a warning

muchzill4:
## Changes

- Updated lack of remoteproc devices on a target to an info level instead of a warning
- ⚠️ Breaking change: this affects consumers of `health --output json`, as we now have a `status: "info"`

### DIFF
--- a/internal/output/templates/health_test.go
+++ b/internal/output/templates/health_test.go
@@ -80,6 +80,29 @@ func TestPrintHealthReport(t *testing.T) {
 			assert.Contains(t, out.String(), "no remoteproc devices found")
 		})
 
+		t.Run("it renders an info icon for info checks", func(t *testing.T) {
+			toPrint := templates.PrintableHealthReport{
+				Target: &health.TargetReport{
+					Connectivity: health.HealthCheck{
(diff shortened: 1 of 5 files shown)


internal/output/templates/health_test.go
```diff
@@ -80,6 +80,29 @@ func TestPrintHealthReport(t *testing.T) {
 			assert.Contains(t, out.String(), "no remoteproc devices found")
 		})
 
+		t.Run("it renders an info icon for info checks", func(t *testing.T) {
+			toPrint := templates.PrintableHealthReport{
+				Target: &health.TargetReport{
+					Connectivity: health.HealthCheck{
+						Name:   "Connected",
+						Status: health.CheckStatusOK,
+					},
```
yejseo01, line 89: It seems like we are not checking anything about this in line 101-103. Do we even need this part?

muchzill4: Sadly we do, because template won't render remoteproc details w/o a passing connectivity check.

yejseo01 (approved):
Looks good!

PR #161, opened 2026-03-31
feat: standardised structured logging

awphi:
## Changes

- Stops returning log entries from inner functions and instead logs in-place. While this is side-effecty, logging is a special class of side effect that is easy to mock or stub and doesn't need testing as the only state it mutates is stderr.
- Removes our custom `ConsoleLogger` implementation and replaces with a package-level singleton `logger` based on `log/slog` for out of the box JSON support
  - For 'plain' log entries I've used a [3rd party slog handler](https://github.com/lmittmann/tint) that makes them a bit prettier. I'm happy to ditch this if it's not well-liked.
  - As for the question of why not use slog directly and abstract it behind this little `logger` package? I want to make it easy to replace the underlying implementation that actually writes to stderr in the future;
    - There's also a question here that if we're using tint should we just bite the bullet use [charmbracelet/log](https://github.com/charmbracelet/log)

Deploy error (plain):

Screenshot 2026-03-31 at 13 12 44

Deploy error (JSON):

Screenshot 2026-03-31 at 13 05 07

Info message (plain):

Screenshot 2026-03-31 at 13 08 37


internal/output/logger/logger_test.go
```diff
@@ -0,0 +1,76 @@
...
+	"encoding/json"
+	"os"
+	"testing"
+
+	"github.com/arm/topo/internal/output/logger"
+	"github.com/arm/topo/internal/output/term"
+	"github.com/stretchr/testify/assert"
+)
+
+func setup(t *testing.T) *bytes.Buffer {
```
muchzill4: Thoughts about calling this something like `setupTestLogOutputBuffer` or such?

internal/output/logger/logger.go
```diff
@@ -1,35 +1,53 @@
...
 )
 
-const (
-	Info    Level = "INFO"
-	Warning Level = "WARN"
-	Err     Level = "ERROR"
-)
+func newPlainLogger() *slog.Logger {
+	return slog.New(tint.NewHandler(output, &tint.Options{
+		TimeFormat: time.TimeOnly,
```
awphi, line 20: Not in love with including timestamps atm because we have so few logs messages it just seems weird. Tint doesn't provide an easy means of disabling it though so perhaps one vote in favour of charmbracelet/log?

I also think this would be solved if we moved more of our console output to logger calls e.g. it would be useful to see a chronological record of all the deploy steps, especially for us to debug timeout issues or slow deploys etc.

cmd/topo/clone.go
```diff
@@ -50,7 +50,7 @@ Some projects require build arguments. Supply them on the command line or answer
 		if err != nil {
 			return err
 		}
-		c := console.NewLogger(os.Stderr, outputFormat)
+		logger.SetOutputFormat(outputFormat)
```
muchzill4: WDYT about putting it somewhere in `main`? Will make it easier not to forget about setting up logging when we add new commands.

internal/output/logger/logger_test.go
```diff
@@ -0,0 +1,76 @@
...
+	t.Helper()
+	var buf bytes.Buffer
+	logger.SetOutput(&buf)
+	logger.SetOutputFormat(term.JSON)
```
muchzill4: NIT: I wonder if we should take format as argument to setup. This is now implicit.

muchzill4 (approved):
Great stuff! 💯

cmd/topo/clone.go
awphi: Agreed - required changing the behaviour of `resolveOutput` a bit to not panic but I think it's a net positive - we can freely add an `--output` flag to any command we want and `logger.X` calls will just work https://github.com/arm/topo/pull/161/commits/6dffea1806b0fb873b4043c47beeba44fced9b6b

internal/output/logger/logger_test.go
awphi: https://github.com/arm/topo/pull/161/commits/44d0486beec4f98087673fe2bae4a09146389775

internal/output/logger/logger_test.go
awphi: https://github.com/arm/topo/pull/161/commits/44d0486beec4f98087673fe2bae4a09146389775

tgonzalezorlandoarm (approved):

PR #162, opened 2026-03-31
refactor: remove `WithMockExec` and `ExecSSH`

muchzill4:
## Changes

- Removed `ExecSSH` along with related fields on `Connection` and `ConnectionOptions`. We now call `ssh.ExecCmd` directly.
- `ConnectionOptions.SSHArgs() []string` — new method that converts options to SSH arguments. Pure function, replaces the scattered arg-building logic.

### DIFF
--- a/internal/target/connection.go
+++ b/internal/target/connection.go
@@ -3,7 +3,6 @@ package target
 import (
 	"bytes"
 	"fmt"
-	"os/exec"
 	"runtime"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


internal/target/connection.go
```diff
@@ -47,12 +37,7 @@ func (c *Connection) RunWithStdin(cmdStr string, stdin []byte) (string, error) {
 }
 
 func (c *Connection) run(cmdStr string, stdin []byte) (string, error) {
-	sshArgs := c.connectTimeoutArgs()
-	if c.opts.Multiplex && runtime.GOOS != "windows" {
-		sshArgs = append(sshArgs, "-o", "ControlMaster=auto", "-o", "ControlPersist=10s", "-o", "ControlPath=~/.ssh/topo-cm-%r@%h:%p")
-	}
-
-	cmd := c.exec(c.SSHTarget, cmdStr, stdin, sshArgs...)
+	cmd := ssh.ExecCmd(c.SSHTarget, cmdStr, stdin, c.opts.SSHArgs()...)
```
tgonzalezorlandoarm, line 40: Do we understand correctly that ExecCmd can now be a method of Connection instead of a separate function defined in SSH?

tgonzalezorlandoarm:
Thanks for doing this change! The mock injection was something that needed to go, I like this testing pattern

internal/target/connection.go
awphi: Had the same observation but I do appreciate the purity of `ssh.ExecCmd` so would vote to leave as-is.

awphi (approved):
Nice! 🚀

tgonzalezorlandoarm (approved):

internal/target/connection.go
muchzill4: Astutely observed. If you're both ok with it I'll leave it as is to keep this (already bloated) change contained.

The plan is to eventually remove `ssh.ExecCmd` and have two command runners we'll construct at the top of the stack and inject instead - one for running commands on the host (when we're explicitly or implicitly targeting localhost) and one for running commands on remote target.

PR #163, opened 2026-03-31
refactor: logger options + fix JSON output

awphi:
## Changes

- `--output json` was broken by a refactor made after review in the slog PR (#161) without manual re-testing (this is my bad, will hold my hands up) - I'll make a follow-up to add a proper test for this as it's a pretty big hole in our e2e tests atm.
- Refactors logger options to be a bit neater and improve extensibility (`--log-level` coming soon tm)


muchzill4 (approved):
:shipit:

PR #164, opened 2026-03-31
test: add e2e tests for JSON output

awphi:
## Changes

- Adds a couple of small e2e tests for health and templates to ensure `--output json` produces valid JSON in the expected form
  - This introduces slight coupling between the tests and the JSON output format but I think this is a positive as it will force us to detect breaking changes in this format.
- Add an e2e test to templates to ensure errors are also written in JSON form when specified w/ the expected fields

### DIFF
--- a/e2e/health_test.go
+++ b/e2e/health_test.go
@@ -1,11 +1,9 @@
 package e2e
 
 import (
-	"encoding/json"
 	"os/exec"
 	"testing"
 
-	"github.com/arm/topo/internal/health"
 	"github.com/arm/topo/internal/testutil"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -44,9 +42,7 @@ func TestHealthCheck(t *testing.T) {
 		out, err := runCheckHealth(topo, target, "--output", "json")
 
 		assert.NoError(t, err)
-		var entry health.TargetReport
-		err = json.Unmarshal([]byte(out), &entry)
-		assert.NoError(t, err)
+		testutil.AssertGoldenFile(t, out, "testdata/TestHealthCheckJson.golden")
 	})
 }
 
--- a/e2e/health_test.go
+++ b/e2e/health_test.go
@@ -1,9 +1,11 @@
 package e2e
 
 import (
+	"encoding/json"
 	"os/exec"
 	"testing"
 
+	"github.com/arm/topo/internal/health"
 	"github.com/arm/topo/internal/testutil"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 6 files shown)


e2e/health_test.go
```diff
@@ -37,10 +39,19 @@ func TestHealthCheck(t *testing.T) {
 		assert.NoError(t, err)
 		assert.Contains(t, out, "Connectivity: ❌")
 	})
+
+	t.Run("outputs JSON when specified", func(t *testing.T) {
+		out, err := runCheckHealth(topo, target, "--output", "json")
+
+		assert.NoError(t, err)
+		var entry health.TargetReport
```
muchzill4: Reaching into `internal` in `e2e` tests smells. It is _acceptable_ for test utils, but for production code I would rather the e2e tests had no access to the source of topo. The idea is that they should see the world as a user would.

How about:
```go
want := `{
  "host": {
    "dependencies": [
      { "name": "SSH", "status": "ok", "value": "ssh" },
      { "name": "Container Engine", "status": "ok", "value": "docker" }
    ]
  },
  ....
  }
}`
assert.JSONEq(t, want, got)
```

Though this is a bit verbose (note I dotted some of the JSON). 
Another approach would be to use golden files (TIL!).

WDYT?

muchzill4 (commented):
Nice addition, but I think we should try to avoid reaching into internals from e2e if feasible.

e2e/health_test.go
awphi: Golden files sounds like a good approach to me! Was just trying to avoid being overly verbose but agree the coupling is a bit icky

awphi: https://github.com/arm/topo/pull/164/commits/0123bb215813a8f91806c1c630f0deab60857863 - didn't create a golden file for the error case since it includes a timestamp

muchzill4: I guess a lighter alternative would be to some sort of `assertValidJSON` and `assertJSONContains`, where the latter would check for one of dependencies perhaps.

awphi: The thing I like about the golden file approach is it keeps us honest about breaking changes to the JSON schema in a way that `assertValidJSON` or `assertJSONContains` does not

internal/testutil/testutil.go
```diff
@@ -103,3 +103,13 @@ func AssertFileContents(t *testing.T, wantContents string, path string) {
 	require.NoError(t, err)
 	require.Equal(t, wantContents, string(got))
 }
+
+func AssertGoldenFile(t *testing.T, got string, goldenPath string) {
+	t.Helper()
+
```
muchzill4, line 109:
```go
if os.Getenv("UPDATE_GOLDEN") == "1" {
    writeGoldenFile(got)
}
```

And a snippet in the `DEVELOPMENT.md` explaining how to update golden files?

e2e/templates_test.go
```diff
@@ -60,6 +61,28 @@ totalmemory_kb: 4194304
...
+	t.Run("outputs JSON when specified", func(t *testing.T) {
+		cmd := exec.Command(bin, "templates", "--output", "json")
+		out, err := cmd.CombinedOutput()
+		require.NoError(t, err)
+
+		testutil.AssertGoldenFile(t, string(out), "testdata/TestTemplatesJson.golden")
+	})
+
+	t.Run("outputs errors as JSON when specified", func(t *testing.T) {
+		cmd := exec.Command(bin, "templates", "--output", "json", "--target", "invalid-target")
```
muchzill4, line 74: Should this also be a golden error file? 🤔

awphi: This one is stickier because it contains a timestamp and I wanted to avoid the hassle of wiring in a fake `time.Now` implementation. If we used a goldenfile library that supported some sort of matching pattern syntax we could do this but seems overkill imo

internal/testutil/testutil.go
awphi: https://github.com/arm/topo/pull/164/commits/2346f1ac8c1b71928e917bd4ca045f03bef85bc6

muchzill4 (approved):

PR #166, opened 2026-03-31
refactor: merge config and sshconfig

tgonzalezorlandoarm:
## Background

Currently, we have two different files for manipulating SSH configs: config.go and sshconfig.go
There's no need to split the functionality between reading the config and writing it into two separate files

## Changes

- Merge the two files and tests together
- Refactor naming to remove unnecessary "SSH" (as the functions belong to the `ssh` package)
- Reorder public APIs to be shown first, private APIs second.

No behavioral changes intended.


internal/ssh/config.go
```diff
@@ -114,3 +119,153 @@ func IsExplicitHostConfig(host string, config []byte) bool {
...
+		Key:   key,
+		Value: value,
+	}
+}
+
+func (d ConfigDirective) String() string {
+	return fmt.Sprintf("%s %s", d.Key, d.Value)
+}
+
+func CreateConfig(dest Destination, targetSlug string) error {
```
muchzill4: I wonder if we should somehow differentiate between `Config` struct and config file writing. Right now we have `NewConfig()` but also `CreateConfig()`. Perhaps we should call the latter `CreateConfigFile`? If so, then all config file-y things should follow suit.

internal/ssh/config_test.go
```diff
@@ -135,3 +140,167 @@ debug1: /tmp/config line 5: Applying options for *
 		assert.False(t, got)
 	})
 }
+
+func TestModifySSHConfig(t *testing.T) {
```
muchzill4:
```suggestion
func TestCreateOrModifyConfig(t *testing.T) {
```

muchzill4: Nit, `CreateOrModifyConfig` is defined after `CreateConfig` in `config.go`, but here the test order is reversed.

internal/ssh/config.go
tgonzalezorlandoarm: Ah, I forgot about this. Hopefully it's done?

internal/ssh/config_test.go
tgonzalezorlandoarm: done!

internal/ssh/config.go
```diff
@@ -114,3 +119,153 @@ func IsExplicitHostConfig(host string, config []byte) bool {
 
 	return false
 }
+
+func NewDirectiveIdentityFile(path string) ConfigDirective {
```
muchzill4: Should this be `NewConfigDirectiveIdentityFile`? 🤔

muchzill4 (approved):

internal/ssh/config.go
tgonzalezorlandoarm: done

PR #169, opened 2026-04-01
fix: prevent SSH host key prompt from overlapping with spinner

muchzill4:
## Changes

Use `StrictHostKeyChecking=yes` in the known host probe so SSH fails immediately when the host key is unknown, rather than prompting interactively and racing with the spinner output. We already provide an escape hatch with `--accept-new-host-keys` arg.

### DIFF
--- a/internal/health/health.go
+++ b/internal/health/health.go
@@ -146,8 +146,11 @@ func connectivityCheck(status ConnectionStatus) HealthCheck {
 	}
 
 	check.Value = status.Error.Error()
-	if errors.Is(status.Error, target.ErrPasswordAuthentication) {
-		check.Fix = fmt.Sprintf("run `topo setup-keys --target %s` or manually setup SSH keys for the target", status.Destination)
+	switch {
+	case errors.Is(status.Error, target.ErrPasswordAuthentication):
+		check.Fix = fmt.Sprintf("run `topo setup-keys --target %s` to configure SSH keys", status.Destination)
+	case errors.Is(status.Error, target.ErrHostKeyVerification):
+		check.Fix = fmt.Sprintf("run `topo health --target %s --accept-new-host-keys` to trust the target's identity", status.Destination)
 	}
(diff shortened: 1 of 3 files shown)


awphi (commented):
Nice, definite improvement. I wonder if we could hand-hold this a bit more and add a hint to the health check failure - something like;
```
Host
----
SSH: ✅ (ssh)
Container Engine: ✅ (docker)

Target
------
Connectivity: ❌ (ssh host key verification failed)
  → run `topo health --target <target> --accept-new-host-keys` if you trust the target system
```

Potentially too verbose but you get the idea

yejseo01 (approved):
Looks good to me!

awphi (approved):

internal/health/health.go
```diff
@@ -146,8 +146,11 @@ func connectivityCheck(status ConnectionStatus) HealthCheck {
 	}
 
 	check.Value = status.Error.Error()
-	if errors.Is(status.Error, target.ErrPasswordAuthentication) {
-		check.Fix = fmt.Sprintf("run `topo setup-keys --target %s` or manually setup SSH keys for the target", status.Destination)
+	switch {
+	case errors.Is(status.Error, target.ErrPasswordAuthentication):
+		check.Fix = fmt.Sprintf("run `topo setup-keys --target %s` to configure SSH keys", status.Destination)
+	case errors.Is(status.Error, target.ErrHostKeyVerification):
```
federicobozzini, line 152: Am I wrong in thinking we are not handling one case properly here?

This error comes from https://github.com/muchzill4/topo/blob/main/internal/target/ssh_authentication_probe.go#L112, but we can `ErrHostKeyVerification` even if the host key has changed / mismatches an existing
  known_hosts entry. In other words we collapse the two errors into one and we offer a fix that would only work for one case.

muchzill4: Good shout!

PR #170, opened 2026-04-01
chore: check the pr title to be conventional commits

yejseo01:
## Changes

- add `pr.yaml` in workflow, which will check the PR title to be conventional


.github/workflows/pr.yaml
```diff
@@ -0,0 +1,21 @@
+name: PR
+
+on:
+  pull_request_target:
```
awphi: I've heard some truly terrible things about `pull_request_target` and leaking secrets - this pipelines looks safe as-is but perhaps a comment warning to avoid executing arbitrary code in here in the future?

tgonzalezorlandoarm: If there is a risk on that front, I am of the opinion that it's not worth taking it just so that we can enforce conventional commits

yejseo01: I can use other things to enforce conventional commits

yejseo01: maybe this one -https://github.com/semantic-release/commit-analyzer#options

tgonzalezorlandoarm (approved):

.github/workflows/pr.yaml
tgonzalezorlandoarm: I gave this an ok - @awphi could you provide more information of what can happen and how?

PR #171, opened 2026-04-01
feat: linter - enable funcorder and reorder Go methods

tgonzalezorlandoarm:
Enable the golangci-lint `funcorder` checker and configure it to enforce constructor and struct-method ordering.

No functional behavior change intended.


awphi (approved):
🚀

PR #172, opened 2026-04-01
feat: use release please

yejseo01:
## Changes

- use release-please(https://github.com/googleapis/release-please) to release

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,27 +1,27 @@
 name: Release
 
 on:
+  push:
+    branches:
+      - main
   workflow_dispatch:
-    inputs:
-      release-type:
-        description: One of 'Major', 'Minor', 'Patch', or 'Pre-release'.
-        required: true
-        type: choice
-        default: Pre-release
-        options:
-          - Major
-          - Minor
-          - Patch
-          - Pre-release
 
 permissions:
   contents: write
+  issues: write
+  pull-requests: write
   packages: read
 
 jobs:
   release:
     runs-on: ubuntu-24.04-arm
     steps:
+      - name: Release Please
+        id: release
+        uses: googleapis/release-please-action@v4
+        with:
+          release-type: go
+
       - name: Checkout
         uses: actions/checkout@v6
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -1,27 +1,27 @@
...
-        default: Pre-release
-        options:
-          - Major
-          - Minor
-          - Patch
-          - Pre-release
 
 permissions:
   contents: write
+  issues: write
```
awphi, line 11: Does release-please create issues and if so do we need it?

.github/workflows/release.yaml
```diff
@@ -117,22 +92,12 @@ jobs:
...
-      - name: Create draft GitHub Release
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          VERSION: ${{ steps.version.outputs.version }}
-        run: |
-          gh release create "v${VERSION}" \
-            --draft \
-            --title "v${VERSION}" \
-            --notes "Release v${VERSION} - signing in progress..."
-
```
awphi, line 135: Is this not valuable anymore or is it handled by release-please? Tbh if we can get away with not having it then I'm all for it just want to understand a bit better

.github/workflows/release.yaml
```diff
@@ -1,27 +1,27 @@
...
+  pull-requests: write
   packages: read
 
 jobs:
   release:
     runs-on: ubuntu-24.04-arm
     steps:
+      - name: Release Please
+        id: release
+        uses: googleapis/release-please-action@v4
```
awphi, line 21: Should pin to a commit hash

.github/workflows/release.yaml
```diff
@@ -1,27 +1,27 @@
...
   contents: write
+  issues: write
+  pull-requests: write
   packages: read
 
 jobs:
   release:
     runs-on: ubuntu-24.04-arm
     steps:
+      - name: Release Please
```
awphi, line 19: More descriptive name would be helpful here e.g. 'Update release PR' or similar

awphi (changes requested):
Handful of nits and some questions about how/when we run this

.github/workflows/release.yaml
```diff
@@ -1,27 +1,27 @@
 name: Release
 
 on:
+  push:
+    branches:
+      - main
```
awphi, line 6: Not sure we should be running on every push to main right? Do we want to release signed artifacts/bump version number for every little tweak?

.github/workflows/release.yaml
awphi, line 19: Can we not skip everything after this step if the version number is unchanged i.e. for `refactor: xyz` or `test: xyz`

muchzill4:
Can you explain what `release-please` gains us over `goreleaser`?

yejseo01:
abandoning this approach as we have different approach -  https://github.com/arm/topo/pull/174

PR #173, opened 2026-04-01
refactor: introduce `runner.SSH` and `runner.Local` in place of `Connection`

muchzill4:
## Rationale

Replaces conflated `Connection` type with `runner.Runner` interface backed by an `SSH` and `Local` concrete types.

## Changes

- Introduce `runner.SSH` and `runner.Local` as concrete runner types, replacing the old `Connection` abstraction
- Add `runner.For` to wire the correct runner based on destination (SSH vs localhost)
- Add `runner.Mock` for use in tests, replacing hand-rolled local mock types
- Replace locally-defined runner interfaces across packages with `runner.Runner` interface
- Centralise and extend error classification with timeout detection

⚠️ This is bigger than I'd like, but a lot of this PR is noise. The bits worth reviewing are:
 - `runner` package
 - `runner.For` usage
 - `CheckTarget`, where we explicitly construct `Local` or `SSH` runner
 - extension to `ClassifyStderr`


yejseo01 (approved):
As we have discussed, one of the main purpose of this PR is to surface the execution logic difference on localhost/remote target to the highest stack. Before this change, such difference was treated in ExecCmd(which is not RunCommand). Now, the logic is moved to `prepareRunner` in `health.go`

I can't see any more issues on this PR, and it seems good to go! 🚀

PR #175, opened 2026-04-02
feat: delete setup-ssh

awphi:
## Background

Given we're pursuing a slightly different approach to target management in the vscode extension this command no longer has any utility

## Changes

- Deletes the `setup-ssh` command and corresponding internals/tests

### DIFF
--- a/cmd/topo/vscode.go
+++ b/cmd/topo/vscode.go
@@ -3,7 +3,6 @@ package main
 import (
 	"os"
 
-	"github.com/arm/topo/internal/ssh"
 	"github.com/arm/topo/internal/vscode"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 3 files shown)


muchzill4 (approved):

federicobozzini (commented):
Should the title be `feat!` in order to follow the proper conventional commit specs?

awphi:
> Should the title be `feat!` in order to follow the proper conventional commit specs?

I considered this but thought that commands defined in `vscode.go` probably don't constitute the public API and therefore this isn't breaking to end users. I guess the question is do we consider vscode a customer of the CLI that consumes semver?

federicobozzini (approved):
Good point. Since the command was hidden I think it's OK not consider this a breaking change (yet). Let's keep it as it is :+1:

PR #177, opened 2026-04-02
chore: generate changelog

yejseo01:
## Changes

- make better changelog

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -122,7 +122,7 @@ jobs:
         run: |
           gh workflow run sign-and-release.yml \
             --repo Arm-Debug/topo-signer \
-            --ref append-not-rewrite-release-note \
+            --ref v4 \
             --field run-id=${{ github.run_id }} \
             --field release-version=${VERSION} \
             --field release-tag=${VERSION} \
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -56,7 +56,7 @@ jobs:
         with:
           distribution: goreleaser
           version: '~> v2'
-          args: release --clean --skip=publish
+          args: release --clean --draft
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           GORELEASER_CURRENT_TAG: ${{ steps.calculate_next_version.outputs.next_version }}
@@ -106,16 +106,6 @@ jobs:
             dist/topo_linux_arm64
           if-no-files-found: error
 
-      - name: Create draft GitHub Release
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          VERSION: ${{ steps.calculate_next_version.outputs.next_version }}
-        run: |
-          gh release create "${VERSION}" \
-            --draft \
-            --title "${VERSION}" \
-            --notes "Release ${VERSION} - signing in progress..."
-
       - name: Create GitHub App token for signer
         id: signer-token
         uses: actions/create-github-app-token@v3
@@ -132,7 +122,7 @@ jobs:
         run: |
           gh workflow run sign-and-release.yml \
             --repo Arm-Debug/topo-signer \
-            --ref v3 \
+            --ref append-not-rewrite-release-note \
             --field run-id=${{ github.run_id }} \
             --field release-version=${VERSION} \
             --field release-tag=${VERSION} \
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -48,9 +48,9 @@ changelog:
   abbrev: -1
   filters:
     include:
-      - '^(feat|fix|perf|security|breaking|build)(\\([^)]+\\))?(!)?: .*$'
+      - '^(feat|fix|perf|security|breaking|build)(\\([^)]+\\))?: .*$'
+      # Scan for breaking changes
       - '^.*!:.*$'
-      - '^.*BREAKING CHANGE:.*$'
   groups:
     - title: Breaking Changes
       # Examples: "feat!: drop legacy flags", "fix(auth)!: change token format", "BREAKING CHANGE: drop v1 API"
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -39,7 +39,7 @@ checksum:
   name_template: "checksums.txt"
 
 changelog:
-  use: git
+  use: github
   sort: asc
   abbrev: -1
   filters:
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -43,6 +43,7 @@ checksum:
 
 changelog:
   use: github
+  format: "{{ .Message }}{{ if .Logins }} ({{ .Logins | englishJoin }}){{ end }}"
   sort: asc
   abbrev: -1
   filters:
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -53,11 +53,14 @@ changelog:
       - '^.*BREAKING CHANGE:.*$'
   groups:
     - title: Breaking Changes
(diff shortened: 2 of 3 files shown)


.goreleaser.yaml
```diff
@@ -39,4 +39,20 @@ checksum:
...
 changelog:
-  disable: true
\ No newline at end of file
+  use: git
+  sort: asc
+  abbrev: -1
+  filters:
+    exclude:
+      - "^docs:"
+      - "^chore:"
```
muchzill4: I wonder if we should use `include:` instead. 🤔 

We probably want to exclude `refactor:` as well. Unsure about other categories - best compare the list of semantic categories with what we actually want in the changelog.

.goreleaser.yaml
```diff
@@ -39,4 +39,20 @@ checksum:
   name_template: "checksums.txt"
 
 changelog:
-  disable: true
\ No newline at end of file
+  use: git
```
Luke-Parkin: did you try
`github`
or
`github-native` ?  (actually don't do this one, it disables groups)

https://goreleaser.com/customization/publish/changelog/

yejseo01: Will try.

.goreleaser.yaml
```diff
@@ -39,4 +39,20 @@ checksum:
...
+  sort: asc
+  abbrev: -1
+  filters:
+    exclude:
+      - "^docs:"
+      - "^chore:"
+  groups:
+    - title: Breaking Changes
+      regexp: '^.*?!:.*$|^.*BREAKING CHANGE:.*$'
```
Luke-Parkin, line 57: We're expecting here to match on BREAKING CHANGE:

But I thought our conventional commits meant we ALSO matched on a 
FEAT**!** ?

See:
https://www.conventionalcommits.org/en/v1.0.0/
`BREAKING CHANGE: a commit that has a footer BREAKING CHANGE:, or appends a ! after the type/scope, introduces`

And...
https://github.com/arm/topo/commit/887f1a1cf14ba1628b0070e0189fa046b741e960

Luke-Parkin (dismissed):

.goreleaser.yaml
```diff
@@ -39,4 +39,20 @@ checksum:
...
 changelog:
-  disable: true
\ No newline at end of file
+  use: github
+  sort: asc
+  abbrev: -1
+  filters:
+    exclude:
+      - "^docs:"
+      - "^chore:"
```
awphi: Should also hide `refactor`, `test` and `style`?

Also not sure if we should exclude docs - has some user value

.goreleaser.yaml
```diff
@@ -39,4 +39,20 @@ checksum:
...
+  groups:
+    - title: Breaking Changes
+      regexp: '^.*?!:.*$|^.*BREAKING CHANGE:.*$'
+      order: 0
+    - title: Features / Enhancements
+      regexp: '^.*?feat(\([[:word:]]+\))??:.*$'
+      order: 1
+    - title: Bug Fixes
+      regexp: '^.*?fix(\([[:word:]]+\))??:.*$'
```
awphi, line 65: Is this regexp just used for matching or for categorising commits by scope? If it's just the former couldn't we use a simpler expression like `^fix:`

awphi (commented):
Nothing blocking from me but a couple small observations

muchzill4:
Isn't this PR "generate a changelog", I agree that it's better than nonexistence of changelog, but it's weird to call it "generate better changelog". 😄

.goreleaser.yaml
federicobozzini: +1 on this. Let's be restrictive.

.goreleaser.yaml
yejseo01: I'm not sure what you mean? the regex matches both feat! header and BREAKING CHANGE footer. What is the issue?

Luke-Parkin: Oh yeah. Sorry Missed that. Although probably a good note on here about leaving a comment explaining what regex does? It's not easy to read

yejseo01: I will gently disagree with leaving a comment. I don't think it's a good practice to explain regexes in comment.

.goreleaser.yaml
yejseo01: done - [13ec441](https://github.com/arm/topo/pull/177/commits/13ec441809b28da3c6ec6895481baa470c1e1b85)

.goreleaser.yaml
Luke-Parkin: > I don't think it's a good practice to explain regexes in comment.

Why not?

yejseo01: I feel like it’s wayy too verbose. The most precise way of describing such string is a regex, and I think having examples in the comment might make it more confusing in details than just reading the regex itself.

Luke-Parkin: > The most precise way of describing such string is a regex
For code yes, but not to humans!

Explaining the intent behind the code is important when it's not clear. If it were clear, I'd agree, but this isn't clear! Proved by Adam questioning the intent of a regex below. Adding a comment just describing briefly what it does for the engineers who can't sight-read regex is useful and (imo) one of the few widely accepted uses of comments

.goreleaser.yaml
yejseo01: conventional commit format allows optional scopes such as `feat(api):...` and simplifying the regex won't support it. Simplifying it would also skip any merge commits, which I'm not sure is possible to happen right now, but better to be safe than sorry?

.goreleaser.yaml
yejseo01: done - [e9e4721](https://github.com/arm/topo/pull/177/commits/e9e4721f583ccac15ab05569db496d1969a4c872)

.goreleaser.yaml
```diff
@@ -39,4 +42,25 @@ checksum:
   name_template: "checksums.txt"
 
 changelog:
-  disable: true
\ No newline at end of file
+  use: github
+  format: "{{ .Message }}{{ if .Logins }} ({{ .Logins | englishJoin }}){{ end }}"
```
muchzill4, line 46: Do we need to explicitly set `format:`?

.goreleaser.yaml
```diff
@@ -39,4 +42,25 @@ checksum:
...
+  sort: asc
+  abbrev: -1
+  filters:
+    include:
+      - '^(feat|fix|perf|security|breaking|build)(\\([^)]+\\))?(!)?: .*$'
+      - '^.*!:.*$'
+      - '^.*BREAKING CHANGE:.*$'
```
muchzill4: - I think we can drop "BREAKING CHANGE", because we mark them with `!`
- I don't think we need `build`, but would benefit fro `chore` which I believe is dependabot bumps
- I'm not sure why we need first and second entry - the regex is hard to parse, _can't we just_:

```
include:
   - "^(feat|fix|perf|security|chore)!?:"
```

yejseo01: addressed at [181f748](https://github.com/arm/topo/pull/177/commits/181f7488eeb6816c0ce230417e3d53f7732be4e5)

.goreleaser.yaml
yejseo01: went with include instead of exclude

muchzill4 (approved):

.github/workflows/release.yaml
```diff
@@ -132,7 +122,7 @@ jobs:
         run: |
           gh workflow run sign-and-release.yml \
             --repo Arm-Debug/topo-signer \
-            --ref v3 \
+            --ref append-not-rewrite-release-note \
```
yejseo01: will merge after fixing this

PR #178, opened 2026-04-02
fix: use password-based auth for Public Key transfer operation

tgonzalezorlandoarm:
If not explicitly specified, the public key transfer operation will
attempt to SSH by using default SSH keys.
    
```
(...)
debug1: Will attempt key: /home/tomgon01/.ssh/id_rsa
debug1: Will attempt key: /home/tomgon01/.ssh/id_ecdsa
debug1: Will attempt key: /home/tomgon01/.ssh/id_ecdsa_sk
debug1: Will attempt key: /home/tomgon01/.ssh/id_ed25519_sk
debug1: Will attempt key: /home/tomgon01/.ssh/id_xmss
debug1: Will attempt key: /home/tomgon01/.ssh/id_dsa
```

It will iterate over multiple identity files and then fail as it
has tried too many options.

## Changes
 - Use password-based authentication when transferring the newly created SSH key onto the target.

### DIFF
--- a/internal/runner/mock.go
+++ b/internal/runner/mock.go
@@ -17,3 +17,12 @@ func (m *Mock) RunWithStdin(command string, stdin []byte) (string, error) {
 	args := m.Called(command, stdin)
 	return args.String(0), args.Error(1)
 }
+
+func (m *Mock) RunWithStdinAndArgs(command string, stdin []byte, sshArgs ...string) (string, error) {
+	callArgs := []any{command, stdin}
+	for _, arg := range sshArgs {
+		callArgs = append(callArgs, arg)
+	}
+	args := m.Called(callArgs...)
+	return args.String(0), args.Error(1)
+}
--- a/internal/runner/ssh.go
+++ b/internal/runner/ssh.go
@@ -47,6 +47,11 @@ func (r *SSH) RunWithArgs(cmdStr string, args ...string) (string, error) {
 	return ssh.RunCommand(r.dest, cmdStr, nil, args...)
 }
 
+func (r *SSH) RunWithStdinAndArgs(cmdStr string, stdin []byte, args ...string) (string, error) {
+	args = append(args, r.opts.SSHArgs()...)
+	return ssh.RunCommand(r.dest, cmdStr, stdin, args...)
+}
+
 func (r *SSH) exec(cmdStr string, stdin []byte, extraSSHArgs []string) (string, error) {
 	return ssh.RunCommand(r.dest, cmdStr, stdin, extraSSHArgs...)
 }
--- a/internal/setupkeys/pubkeytransfer/pub_key_transfer.go
+++ b/internal/setupkeys/pubkeytransfer/pub_key_transfer.go
@@ -6,17 +6,24 @@ import (
 	"os"
 
 	"github.com/arm/topo/internal/command"
-	"github.com/arm/topo/internal/runner"
 )
 
 const remoteAuthorizedKeysCommand = "mkdir -p ~/.ssh && chmod 700 ~/.ssh && cat >> ~/.ssh/authorized_keys && chmod 600 ~/.ssh/authorized_keys"
 
-type PubKeyTransfer struct {
-	pubKeyPath string
-	r          runner.Runner
+var passwordAuthArgs = []string{
+	"-o", "PreferredAuthentications=password",
 }
 
-func NewPubKeyTransfer(privKeyPath string, r runner.Runner) *PubKeyTransfer {
+type sshRunnerWithExtraArgs interface {
+	RunWithStdinAndArgs(command string, stdin []byte, sshArgs ...string) (string, error)
+}
+
+type PubKeyTransfer struct {
+	pubKeyPath string
+	r          sshRunnerWithExtraArgs
+}
+
+func NewPubKeyTransfer(privKeyPath string, r sshRunnerWithExtraArgs) *PubKeyTransfer {
 	return &PubKeyTransfer{
 		pubKeyPath: privKeyPath + ".pub",
 		r:          r,
@@ -33,7 +40,7 @@ func (kt *PubKeyTransfer) Run(outputWriter io.Writer) error {
 		return fmt.Errorf("failed to read public key %s: %w", kt.pubKeyPath, err)
 	}
 
-	cmdOutput, err := kt.r.RunWithStdin(command.WrapInLoginShell(remoteAuthorizedKeysCommand), pubKey)
+	cmdOutput, err := kt.r.RunWithStdinAndArgs(command.WrapInLoginShell(remoteAuthorizedKeysCommand), pubKey, passwordAuthArgs...)
 	if err != nil {
 		return fmt.Errorf("failed to transfer public key to target: %w", err)
 	}
--- a/internal/setupkeys/pubkeytransfer/pub_key_transfer_test.go
+++ b/internal/setupkeys/pubkeytransfer/pub_key_transfer_test.go
@@ -23,11 +23,13 @@ func TestPubKeyTransfer(t *testing.T) {
 
 			r := &runner.Mock{}
 			r.On(
-				"RunWithStdin",
+				"RunWithStdinAndArgs",
 				mock.MatchedBy(func(cmd string) bool {
 					return strings.Contains(cmd, "mkdir -p ~/.ssh && chmod 700 ~/.ssh && cat >> ~/.ssh/authorized_keys && chmod 600 ~/.ssh/authorized_keys")
 				}),
 				pubKeyContent,
+				"-o",
+				"PreferredAuthentications=password",
 			).Return("ssh invoked", nil)
 
 			op := pubkeytransfer.NewPubKeyTransfer(privKeyPath, r)


muchzill4:
I'm struggling to understand what this addresses. Can you help me understand it? 

As far as I understand it essentially flips the order of operations from:
1. Transfer keys
2. Write config

To:
1. Write config
2. Transfer keys

How does writing config entry that has no key available on the filesystem improve things? 🤔

internal/setupkeys/identityfilefragment/identity_file_fragment.go
```diff
@@ -0,0 +1,34 @@
+package identityfilefragment
+
+import (
+	"io"
+
+	"github.com/arm/topo/internal/ssh"
+)
+
+type IdentityFileFragmentWrite struct {
```
yejseo01: I would call this `WriteToIdentityFile`. what does the `fragment` bit of the type name embeds and why is it necessary?

internal/setupkeys/setup_keys_test.go
```diff
@@ -13,40 +14,48 @@ import (
 )
 
 func TestNewKeySetup(t *testing.T) {
-	t.Run("NewKeySetup returns SSHKeyGen then PubKeyTransfer for supported key types", func(t *testing.T) {
+	t.Run("NewKeySetup returns SSHKeyGen then IdentityFileFragmentWrite then PubKeyTransfer for supported key types", func(t *testing.T) {
```
yejseo01:
```suggestion
	t.Run("NewKeySetup returns SSHKeyGen, IdentityFileFragmentWrite then PubKeyTransfer for supported key types", func(t *testing.T) {
```

tgonzalezorlandoarm:
> I'm struggling to understand what this addresses. Can you help me understand it?
> 
> As far as I understand it essentially flips the order of operations from:

Correct, that's the gist of what this PR is doing.

> How does writing config entry that has no key available on the filesystem improve things? 🤔

The key thing to understand (I only wrote that for the pun) is that the `key transfer` operation uses an SSH operation by itself.

If the order remains the same as it currently is:

The key transfer operation will first try to ssh into the target by using  key-based authentication with the typical defaults:

```
$ ssh -v dxg
OpenSSH_8.9p1 Ubuntu-3ubuntu0.14, OpenSSL 3.0.2 15 Mar 2022
debug1: Reading configuration data /home/tomgon01/.ssh/config
debug1: Reading configuration data /home/tomgon01/.ssh/topo_config/topo_imx93.conf
debug1: Reading configuration data /home/tomgon01/.ssh/topo_config/topo_myimx.conf
debug1: /home/tomgon01/.ssh/config line 21: Applying options for dxg
debug1: Reading configuration data /etc/ssh/ssh_config
debug1: /etc/ssh/ssh_config line 3: Applying options for *
debug1: Connecting to <ip> [ip] port 22.
debug1: Connection established.
(...)
debug1: Will attempt key: /home/tomgon01/.ssh/id_rsa 
debug1: Will attempt key: /home/tomgon01/.ssh/id_ecdsa 
debug1: Will attempt key: /home/tomgon01/.ssh/id_ecdsa_sk 
debug1: Will attempt key: /home/tomgon01/.ssh/id_ed25519_sk 
debug1: Will attempt key: /home/tomgon01/.ssh/id_xmss 
debug1: Will attempt key: /home/tomgon01/.ssh/id_dsa 
```

It will iterate over multiple identity files and then fail as it has tried too many options.

If the config file is written first, then it will only try the indicated identity file and if that one fails, it will use password-based authentication, which is what we need at this point.

cmd/topo/setup_keys.go
```diff
@@ -45,22 +45,12 @@ var setupKeysCmd = &cobra.Command{
 			return fmt.Errorf("%w; note: a per user SSH config entry should be created  when setting up keys", err)
 		}
 
-		seq, err := setupkeys.NewKeySetup(dest, privateKeyPath, parsedKeyType)
+		seq, err := setupkeys.NewKeySetup(dest, privateKeyPath, parsedKeyType, targetSlug)
```
yejseo01: `targetSlug` does not need to be passed down when `dest`(type `Destination`) is passed. the slugified name can be generated at the bottom of the stack by calling `dest.Slugify()`

tgonzalezorlandoarm:
It is also worth noting that by the point in time when we write the config file, the key has already been generated and is available in the filesystem of the build machine, it's just not available in the target's filesystem yet

internal/setupkeys/identityfilefragment/identity_file_fragment.go
```diff
@@ -0,0 +1,34 @@
...
+	privateKeyPath string
+	targetSlug     string
```
yejseo01:
```suggestion
```
`targetSlug` is not necessary when `destination` is passed. You can just call `destination.Slugify()`

internal/setupkeys/identityfilefragment/identity_file_fragment.go
```diff
@@ -0,0 +1,34 @@
...
+	destination    ssh.Destination
+}
+
+func NewIdentityFileFragmentWrite(privKeyPath string, targetSlug string, destination ssh.Destination) *IdentityFileFragmentWrite {
+	return &IdentityFileFragmentWrite{
+		privateKeyPath: privKeyPath,
+		targetSlug:     targetSlug,
```
yejseo01:
```suggestion
```

internal/setupkeys/identityfilefragment/identity_file_fragment.go
```diff
@@ -0,0 +1,34 @@
...
+	return "Write IdentityFile fragment to SSH config for target"
+}
+
+func (ifw *IdentityFileFragmentWrite) Run(outputWriter io.Writer) error {
+	directives := []ssh.ConfigDirective{
+		ssh.NewConfigDirectiveIdentityFile(ifw.privateKeyPath),
+		ssh.NewDirective("IdentitiesOnly", "yes"),
+	}
+
+	return ssh.CreateOrModifyConfigFile(ifw.destination, ifw.targetSlug, directives)
```
yejseo01: Maybe it's out of scope for this PR, but `CreateOrModifyConfigFile` shouldn't need slugified version of destination as a separate parameters when it's already embedded inside the destination parameter. For now, at least it can be done like this in this PR. It would be great if there is another PR that fixes `CreateOrModifyConfigFile` to not take slugified destination


```suggestion
	return ssh.CreateOrModifyConfigFile(ifw.destination, ifw.destination.Slugify(), directives)
```

internal/setupkeys/identityfilefragment/identity_file_fragment.go
muchzill4, line 1: If we want namespacing for the sake of a neat public api of `setupkeys` (which is a nice thing to aim for!), I think we should put all of the "leaf" operation files in `operations` directory instead of creating new namespaces which result in repetitive naming. In other words: `identityfilefragment.NewIdentityFileFramentWrite` -> `operations.NewIdentityFileFramentWrite`.

internal/setupkeys/identityfilefragment/identity_file_fragment_test.go
```diff
@@ -0,0 +1,46 @@
...
+			fragmentPath := filepath.Join(tmp, ".ssh", "topo_config", fmt.Sprintf("topo_%s.conf", targetSlug))
+			wantFragmentContents := fmt.Sprintf(`Host example.com
+  HostName example.com
+  User user
+  Port 2222
+  IdentityFile %s
+  IdentitiesOnly yes
+`, filepath.ToSlash(privKeyPath))
+			testutil.AssertFileContents(t, wantFragmentContents, fragmentPath)
+		})
```
yejseo01:
```suggestion
			tmp := t.TempDir()
			testutil.SetHomeDir(t, tmp)
			dest := ssh.NewDestination("user@example.com:2222")
			targetSlug := dest.Slugify()
			privKeyPath := filepath.Join(tmp, ".ssh", fmt.Sprintf("id_ed25519_topo_%s", targetSlug))
			
			op := identityfilefragment.NewIdentityFileFragmentWrite(privKeyPath, targetSlug, dest)
			var buf bytes.Buffer
			require.NoError(t, op.Run(&buf))
			
			require.Empty(t, buf.String())
			mainConfigPath := filepath.Join(tmp, ".ssh", "config")
			wantIncludedFragmentPath := filepath.ToSlash(filepath.Join(tmp, ".ssh", "topo_config", "*.conf"))
			testutil.AssertFileContents(t, fmt.Sprintf("Include %s\n\n", wantIncludedFragmentPath), mainConfigPath)
			fragmentPath := filepath.Join(tmp, ".ssh", "topo_config", fmt.Sprintf("topo_%s.conf", targetSlug))
			wantFragmentContents := fmt.Sprintf(`Host example.com
  HostName example.com
  User user
  Port 2222
  IdentityFile %s
  IdentitiesOnly yes
`, filepath.ToSlash(privKeyPath))
			testutil.AssertFileContents(t, wantFragmentContents, fragmentPath)
		})
```

Not sure how the lines are broken, but this is how I think it should be broken based on AAA(https://semaphore.io/blog/aaa-pattern-test-automation) pattern

muchzill4:
@tgonzalezorlandoarm thanks for explaining!

> If the config file is written first, then it will only try the indicated identity file and if that one fails, it will use password-based authentication, which is what we need at this point.

If that's the case, shouldn't we be explicit about requesting password authentication for the key transfer functionality? The order dependant execution makes it implicit, by creating a _hidden_ failure state that's required for the operation to succeed. Seems hacky.

I have a feeling you could use `runner.SSHOptions` for that.

internal/setupkeys/identityfilefragment/identity_file_fragment.go
tgonzalezorlandoarm: The "fragment" part refers to the SSH config fragment, which is essentially what we are writing to.
We are not writing to an Identity File but rather we are writing an identity file directive to the SSH config fragment.

In the spirit of your comment though, I could use `WriteIdentityFileToFragment` ?

tgonzalezorlandoarm:
> It will iterate over multiple identity files and then fail as it has tried too many options.



> @tgonzalezorlandoarm thanks for explaining!
> 
> > If the config file is written first, then it will only try the indicated identity file and if that one fails, it will use password-based authentication, which is what we need at this point.
> 
> If that's the case, shouldn't we be explicit about requesting password authentication for the key transfer functionality? The order dependant execution makes it implicit, by creating a _hidden_ failure state that's required for the operation to succeed. Seems hacky.
> 
> I have a feeling you could use `runner.SSHOptions` for that.

Good point, done!

internal/setupkeys/identityfilefragment/identity_file_fragment_test.go
tgonzalezorlandoarm: Resolving as I'm not modifying these tests any more.

internal/setupkeys/identityfilefragment/identity_file_fragment.go
tgonzalezorlandoarm: Not applicable anymore

internal/setupkeys/identityfilefragment/identity_file_fragment.go
tgonzalezorlandoarm: resolving as repeated comment: https://github.com/arm/topo/pull/178#discussion_r3045086333

internal/setupkeys/identityfilefragment/identity_file_fragment.go
tgonzalezorlandoarm: resolving https://github.com/arm/topo/pull/178#discussion_r3045086333

cmd/topo/setup_keys.go
tgonzalezorlandoarm: resolving as this is the same comment as  https://github.com/arm/topo/pull/178#discussion_r3045086333

internal/setupkeys/setup_keys_test.go
tgonzalezorlandoarm: Not applicable anymore

internal/setupkeys/identityfilefragment/identity_file_fragment.go
tgonzalezorlandoarm: done in a separate PR #185

internal/runner/mock.go
```diff
@@ -17,3 +17,12 @@ func (m *Mock) RunWithStdin(command string, stdin []byte) (string, error) {
 	args := m.Called(command, stdin)
 	return args.String(0), args.Error(1)
 }
+
+func (m *Mock) RunWithStdinAndArgs(command string, stdin []byte, sshArgs ...string) (string, error) {
+	callArgs := []interface{}{command, stdin}
```
muchzill4:
```suggestion
	callArgs := []any{command, stdin}
```

muchzill4 (approved):

PR #179, opened 2026-04-02
feat!: use a single topo-managed ssh config 

awphi:
## Changes

- Pulls sshconfig-related stuff out of `ssh/config.go` into `ssh/configfile.go` to better separate in-memory config stuff obtained via `ssh -G` from actual file manipulation
- Rewrites out ssh config file manipulation to:
  - Deal with a single ssh config file at `~/.ssh/topo_config` instead of multiple small fragment files. This provides a single source of truth that's easy to edit or delete.
  - Make use of https://github.com/kevinburke/ssh_config (MIT) since single-file manipulation is bit fiddlier so benefits from a library
- Adds a check to `topo setup-keys` that checks if `~/.ssh/topo_config` is a dir and bails out if so with guidance on how to migrate to the new style config


internal/ssh/config_file.go
muchzill4, line 1: s/configfile/config_file`?

internal/ssh/configfile_test.go
```diff
@@ -0,0 +1,140 @@
...
+		tmp := t.TempDir()
+		testutil.SetHomeDir(t, tmp)
+		require.NoError(t, os.MkdirAll(filepath.Join(tmp, ".ssh"), 0o700))
+		dest := ssh.Destination{Host: "board1"}
+		directives := []ssh.ConfigDirective{
+			ssh.NewConfigDirective("IdentityFile", "~/.ssh/id_ed25519"),
+		}
+
+		err := ssh.CreateOrModifyConfigFile(dest, directives)
+		require.NoError(t, err)
```
muchzill4:
```suggestion
		err := ssh.CreateOrModifyConfigFile(dest, directives)
		require.NoError(t, err)

```

Is this part of arrange? 🤔

internal/ssh/configfile_test.go
```diff
@@ -0,0 +1,140 @@
...
+		err = ssh.CreateOrModifyConfigFile(dest, directives)
+		require.NoError(t, err)
```
muchzill4: I think this is part of assert? 🤔

internal/ssh/configfile_test.go
```diff
@@ -0,0 +1,140 @@
...
+
+		err = ssh.CreateOrModifyConfigFile(dest, []ssh.ConfigDirective{
+			ssh.NewConfigDirective("IdentityFile", "~/.ssh/key_new"),
+		})
+		require.NoError(t, err)
+
+		topoConfigPath := filepath.Join(tmp, ".ssh", "topo_config")
+		testutil.AssertFileContents(t,
+			"Host board1\nIdentityFile ~/.ssh/key_new\nUser homer\n",
```
muchzill4: NIT: Raw strings instead of `\n` read better.

internal/ssh/configfile_test.go
```diff
@@ -0,0 +1,140 @@
...
+		)
+	})
+}
+
+func TestCheckForLegacyConfigEntries(t *testing.T) {
+	t.Run("returns nil if no legacy config file exists", func(t *testing.T) {
+		tmp := t.TempDir()
+		testutil.SetHomeDir(t, tmp)
+
+		err := ssh.CheckForLegacyConfigEntries()
```
muchzill4: Does it make sense to call this `CheckForLegacyTopoConfigEntries`? I've added `Topo`.

muchzill4 (approved):
LGTM. I only have some small nits.

One thing we're missing here - not sure if worth addressing - is migration path for existing users. I wonder if `setup_keys` should (for a couple of versions) try to merge our old configs.

tgonzalezorlandoarm (approved):

internal/ssh/configfile_test.go
awphi: https://github.com/arm/topo/pull/179/commits/ddce98f771ec0f8e8dcb328603e5a57c5b646ef3

internal/ssh/configfile_test.go
awphi: https://github.com/arm/topo/pull/179/commits/485f00d679a1a6970110b134142223becb64b0e5

internal/ssh/configfile_test.go
awphi: https://github.com/arm/topo/pull/179/commits/485f00d679a1a6970110b134142223becb64b0e5

internal/ssh/configfile_test.go
awphi: https://github.com/arm/topo/pull/179/commits/485f00d679a1a6970110b134142223becb64b0e5

internal/ssh/config_file.go
awphi: https://github.com/arm/topo/pull/179/commits/95b23923d0d07fcea5003a8b1a4a74b1bb10960d

awphi:
> One thing we're missing here - not sure if worth addressing - is migration path for existing users. I wonder if setup_keys should (for a couple of versions) try to merge our old configs.

I went back and forth on this before landing on the simple check + bail out approach I've got here. The thing I wanted to avoid was deleting things from `~/.ssh/config` - even if it's just the Include line we added it feels like we're overstepping our bounds.

That said we're probably going to need some sort of migration utility for vscode regardless so implementing it in the CLI might make more sense.

WDYT about something as simple as a hidden command `topo migrate-ssh` that uses the `CheckForLegacyTopoConfigEntries` check and if it passes simply reads all the `topo_config/*.conf` files and concats them into a new unified `topo_config` file and updates the include statement in the user's default ssh config? I'll do this in a separate PR if that's okay

muchzill4:
> That said we're probably going to need some sort of migration utility for vscode regardless so implementing it in the CLI might make more sense.

Sounds good to me.

muchzill4 (approved):

PR #182, opened 2026-04-07
feat: add `--timeout` flag to `health`, `describe` and `templates`

muchzill4:
## Changes

- Thread `context.Context` through `Runner` interface, `ssh.RunCommand`, probes, and health checks
- Replace SSH `ConnectTimeout` (`-o ConnectTimeout=N`) with context-based timeout via `exec.CommandContext`
- Add `--timeout` flag to relevant commands (default 5s)
- Return `runner.ErrTimeout` on context cancellation or deadline
- Call sites not yet wired to a real context use `context.TODO()` at the `Run()` call (`install`, `pubkeytransfer`)

### How it looks

```
❱ topo health --target <target> --timeout 1ms
Host
----
SSH: ✅ (ssh)
Container Engine: ✅ (docker)

Target
------
Connectivity: ❌ (timed out)
```

### DIFF
--- a/cmd/topo/root.go
+++ b/cmd/topo/root.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"strings"
@@ -35,7 +36,6 @@ func init() {
 
 const targetEnvVar = "TOPO_TARGET"
 
-const sshConnectTimeout = 5 * time.Second
 
 func addTargetFlag(cmd *cobra.Command) {
 	cmd.Flags().StringP(
@@ -68,6 +68,21 @@ func requireTarget(cmd *cobra.Command) (string, error) {
 		return "", fmt.Errorf("target not specified: provide --target or set TOPO_TARGET env var")
 	}
 	return t, nil
+}
+
+func addTimeoutFlag(cmd *cobra.Command, defaultTimeout time.Duration) {
+	cmd.Flags().Duration("timeout", defaultTimeout, "Maximum time to wait for the command to complete (0 to disable)")
+}
+
+func contextWithTimeout(cmd *cobra.Command) (context.Context, context.CancelFunc) {
+	timeout, err := cmd.Flags().GetDuration("timeout")
+	if err != nil {
+		panic(fmt.Sprintf("internal error: timeout flag not registered: %v", err))
+	}
+	if timeout == 0 {
+		return context.Background(), func() {}
+	}
+	return context.WithTimeout(context.Background(), timeout)
 }
 
 func resolveOutput(cmd *cobra.Command) term.Format {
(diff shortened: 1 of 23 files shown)


cmd/topo/root.go
```diff
@@ -70,6 +69,23 @@ func requireTarget(cmd *cobra.Command) (string, error) {
 	return t, nil
 }
 
+const defaultTimeout = 5 * time.Second
+
+func addTimeoutFlag(cmd *cobra.Command, defaultTimeout time.Duration) {
+	cmd.Flags().Duration("timeout", defaultTimeout, "Maximum time to wait for the command to complete (0 to disable)")
```
Luke-Parkin:
```suggestion
	cmd.Flags().Duration("timeout", defaultTimeout, "Maximum time to wait for the command to complete (0 to disable timeout)")
```
very minor nitpick but seems a little clearer

internal/setupkeys/pubkeytransfer/pub_key_transfer.go
```diff
@@ -40,7 +41,7 @@ func (kt *PubKeyTransfer) Run(outputWriter io.Writer) error {
 		return fmt.Errorf("failed to read public key %s: %w", kt.pubKeyPath, err)
 	}
 
-	cmdOutput, err := kt.r.RunWithStdinAndArgs(command.WrapInLoginShell(remoteAuthorizedKeysCommand), pubKey, passwordAuthArgs...)
+	cmdOutput, err := kt.r.RunWithStdinAndArgs(context.TODO(), command.WrapInLoginShell(remoteAuthorizedKeysCommand), pubKey, passwordAuthArgs...)
```
Luke-Parkin, line 44: There may be a risk here with `cat >> ~/.ssh/authorized_keys` bailing mid write and leaving a malformed entry?

Luke-Parkin: (i guess since it's todo.. not atm?)

internal/runner/ssh.go
```diff
@@ -30,28 +25,39 @@ type SSH struct {
 }
 
 func NewSSH(dest ssh.Destination, opts SSHOptions) *SSH {
-	opts.ConnectTimeout = ssh.NewConfig(dest).ConnectTimeout(opts.ConnectTimeout)
```
Luke-Parkin, line 33: This means we now completely ignore the user's SSH config.

If nothing else, this should probably make it a breaking change. But also we should consider if we want this.

Luke-Parkin (commented):
I think this LGTM for now.

I think you've correctly avoided adding this for deploy at the moment. I wouldn't want to kill transfer commands like this as it could drastically mess up the state. 

Some q's before I approve though!

internal/runner/ssh.go
muchzill4: We don't ignore - it's implicitly respected (as it was before we've added `opts.ConnectTimeout`). 

If ssh's `ConnectionTimeout` is less than `--timeout` it will stop the `ssh` handshake before our `--timeout` hits.

internal/setupkeys/pubkeytransfer/pub_key_transfer.go
muchzill4: Yeah, right now it won't do anything.

Luke-Parkin (approved):

PR #184, opened 2026-04-07
fix: show binary name in health check failure instead of raw runner error

muchzill4:
## Changes

- Add `BinaryExists(bin string) error` to the `Runner` interface, with implementation-specific error messages: `Local` uses `exec.LookPath` and reports "not found in $PATH", `SSH` reports "not found on remote target's $PATH"
- Replace inline `command -v` binary lookups in `health/status.go` and `target/hardware_probe.go` with `Runner.BinaryExists`
- Migrate `CheckHost` to use the runner instead of standalone `BinaryExistsLocally`/`CommandSuccessfulLocally` helpers (now deleted)

### Before

```
❱ topo health --target localhost
...
Target
------
Container Engine: ❌ (local command failed: exit status 127 | stderr: )
Hardware Info: ❌ (local command failed: exit status 127 | stderr: )
...
```

### After

```
❱ topo health --target localhost
...
Target
------
Container Engine: ❌ ("docker" not found in $PATH)
Hardware Info: ❌ ("lscpu" not found in $PATH)
...
```

### DIFF
--- a/internal/health/dependencies.go
+++ b/internal/health/dependencies.go
@@ -1,5 +1,7 @@
 package health
 
+import "context"
+
 type CheckKind int
 
 const (
@@ -147,11 +149,11 @@ func hardwareCapabilityMatches(required []HardwareCapability, available map[Hard
 }
 
 type (
-	BinaryExistsFn      = func(bin string) error
+	BinaryExistsFn      = func(ctx context.Context, bin string) error
 	CommandSuccessfulFn = func(fullCmd string) error
 )
 
-func PerformChecks(dependencies []Dependency, binaryExists BinaryExistsFn, commandSuccessful CommandSuccessfulFn) []DependencyStatus {
+func PerformChecks(ctx context.Context, dependencies []Dependency, binaryExists BinaryExistsFn, commandSuccessful CommandSuccessfulFn) []DependencyStatus {
 	installed := make(map[SoftwareDependency]struct{})
 	result := make([]DependencyStatus, 0, len(dependencies))
 
@@ -165,7 +167,7 @@ func PerformChecks(dependencies []Dependency, binaryExists BinaryExistsFn, comma
 		for _, check := range dep.Checks {
 			switch check.Kind {
 			case CheckBinaryExists:
-				err = binaryExists(dep.Binary)
+				err = binaryExists(ctx, dep.Binary)
 				if err == nil && dep.SoftwareEnumID != UnsetSoftwareDependency {
 					installed[dep.SoftwareEnumID] = struct{}{}
 				}
--- a/internal/health/dependencies.go
+++ b/internal/health/dependencies.go
@@ -1,12 +1,5 @@
 package health
 
-import (
-	"fmt"
-	"os/exec"
-	"strings"
-
-	"github.com/arm/topo/internal/command"
-)
 
 type CheckKind int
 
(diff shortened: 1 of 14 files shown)


muchzill4:
The CI failure is actually nice, it revealed that local runner always does `sh -c`, which is not ideal! Working on a fix separately to this PR.

muchzill4:
~Waits for https://github.com/arm/topo/pull/192.~

internal/health/dependencies.go
```diff
@@ -155,11 +149,11 @@ func hardwareCapabilityMatches(required []HardwareCapability, available map[Hard
 }
 
 type (
-	BinaryExistsFn      = func(bin string) error
+	BinaryExistsFn      = func(ctx context.Context, bin string) error
 	CommandSuccessfulFn = func(fullCmd string) error
```
muchzill4, line 153: `CommandSuccessfulFn` probably also needs `ctx`, but that's something for another day.

internal/health/status.go
```diff
@@ -33,18 +33,11 @@ func ProbeHealthStatus(ctx context.Context, r runner.Runner) HealthStatus {
 	hs.Hardware.RemoteCPU = remoteprocs
 
 	dependenciesToCheck := FilterByHardware(TargetRequiredDependencies, hs.Hardware.Capabilities())
-	binaryExists := func(bin string) error {
-		// We can use `UnsafeBinaryLookupCommand`, because the dependencies we're checking are hardcoded in the codebase
-		if _, err := r.Run(ctx, command.UnsafeBinaryLookupCommand(bin)); err != nil {
```
muchzill4, line 38: The crux of this change:
- for localhost, `command -v yolo` returns status code 127
- for remote, `ssh command -v yolo` returns status code 1

Leveraging `runner.BinaryExists` wraps the error nicely.

internal/health/health.go
```diff
@@ -63,7 +63,12 @@ func (r TargetReport) MarshalJSON() ([]byte, error) {
 }
 
 func CheckHost() HostReport {
-	dependencyStatuses := PerformChecks(HostRequiredDependencies, BinaryExistsLocally, CommandSuccessfulLocally)
+	r := runner.NewLocal()
+	commandSuccessful := func(fullCmd string) error {
+		_, err := r.Run(context.Background(), fullCmd)
+		return err
+	}
```
awphi, line 70: Thoughts on exposing this as a method on the runner? I noted you create a similar anonymous function in `status.go` (albeit with login shell wrapping) - or do you think they're small/different enough where it's not worth it?

internal/runner/ssh.go
```diff
@@ -54,6 +56,13 @@ func (r *SSH) RunWithStdinAndArgs(ctx context.Context, cmdStr string, stdin []by
 	return out, err
 }
 
+func (r *SSH) BinaryExists(ctx context.Context, bin string) error {
+	if _, err := r.Run(ctx, command.UnsafeBinaryLookupCommand(bin)); err != nil {
```
awphi: We previously had a comment about how we could safely use `UnsafeBinaryLookupCommand` since its operands were always hardcoded that we don't seem to have anymore. 

The truth of the comment hasn't changed but maybe we should rename to `UnsafeBinaryExists` or similar to self-document this?

awphi (approved):
Nice! Couple comments but nothing blocking

internal/runner/ssh.go
muchzill4: https://github.com/arm/topo/pull/184/commits/5786eccf209a5bdc7028d2f22b8aeb28d866411f I've chosen the path of least resistance and removed `UnsafeBinaryLookupCommand`. Otherwise, we'd need to dance around interfaces, which I want to avoid.

PR #186, opened 2026-04-07
feat: migrate-ssh command

awphi:
## Changes

- Addresses a suggestion from @muchzill4 [here](https://github.com/arm/topo/pull/179#pullrequestreview-4067572306) to provide a user-friendly path to migrating from the legacy per-target config files to a single config file. 
  - Such an approach can be utilised by vscode to delegate as much responsibility to the CLI as possible
- Only surfaces this command to the user if they encounter the related error when running `setup-keys`, otherwise you can safely continue to use the older ssh config as normal
- Slightly refactors `config_file.go` to provide 'modifier' structs as we now need to delete statements as well as just add/replace (now called 'ensure')

I'm open to auto-migrating when running `setup-keys` instead of requiring manual running a different command or indeed adding a new `--migrate` flag to `setup-keys` to do this. My thinking was if we're going to eventually remove this command it makes sense to decouple it from functionality that will remain as much as possible but lmk!

### DIFF
--- a/internal/ssh/config_file.go
+++ b/internal/ssh/config_file.go
@@ -11,8 +11,6 @@ import (
 	sshconfig "github.com/kevinburke/ssh_config"
 )
 
-var errLegacyConfigEntries = fmt.Errorf("legacy topo ssh config entries found; run 'topo migrate-ssh' to migrate to the new single-file format")
-
 const (
 	defaultConfigFileName = "config"
 	topoConfigFileName    = "topo_config"
@@ -193,25 +191,25 @@ func CreateOrModifyConfigFile(dest Destination, modifiers []ConfigDirectiveModif
 	})
 }
 
-func CheckForLegacyTopoConfigEntries() error {
+func LegacyTopoConfigDirectoryExists() (bool, error) {
 	topoConfigPath, err := getConfigFilePath(topoConfigFileName)
 	if err != nil {
-		return err
+		return false, err
 	}
 
 	info, err := os.Stat(topoConfigPath)
 	if err != nil {
 		if os.IsNotExist(err) {
-			return nil
+			return false, nil
 		}
-		return fmt.Errorf("failed to check for legacy topo ssh config file: %w", err)
+		return false, fmt.Errorf("failed to check for legacy topo ssh config file: %w", err)
 	}
 
 	if info.IsDir() {
-		return errLegacyConfigEntries
+		return true, nil
 	}
 
-	return nil
+	return false, nil
 }
 
 func MigrateLegacyTopoConfig() error {
@@ -220,11 +218,10 @@ func MigrateLegacyTopoConfig() error {
 		return err
 	}
 
-	if err := CheckForLegacyTopoConfigEntries(); err != errLegacyConfigEntries {
-		if err == nil {
-			return fmt.Errorf("legacy topo ssh config directory not found at %s; nothing to migrate", legacyDir)
-		}
+	if exists, err := LegacyTopoConfigDirectoryExists(); err != nil {
 		return err
+	} else if !exists {
+		return fmt.Errorf("legacy topo ssh config directory not found at %s; nothing to migrate", legacyDir)
 	}
 
 	legacyGlob := filepath.Join(legacyDir, "*.conf")
--- a/internal/ssh/config_file.go
+++ b/internal/ssh/config_file.go
@@ -125,7 +125,7 @@ func findOrCreateHostBlock(cfg *sshconfig.Config, alias string) (*sshconfig.Host
 
 func directiveMatches(node sshconfig.Node, directive sshconfig.KV) bool {
 	if kv, ok := node.(*sshconfig.KV); ok {
-		return kv.Key == directive.Key && kv.Value == directive.Value
+		return kv.Key == directive.Key
 	}
 
 	if include, ok := node.(*sshconfig.Include); ok {
--- a/internal/ssh/config_file.go
+++ b/internal/ssh/config_file.go
@@ -11,19 +11,64 @@ import (
 	sshconfig "github.com/kevinburke/ssh_config"
 )
 
-type ConfigDirective = sshconfig.KV
+var errLegacyConfigEntries = fmt.Errorf("legacy topo ssh config entries found; run 'topo migrate-ssh' to migrate to the new single-file format")
 
-func NewConfigDirectivePath(key, path string) ConfigDirective {
-	return ConfigDirective{
-		Key:   key,
-		Value: filepath.ToSlash(path),
+type ConfigDirectiveModifier interface {
+	Apply(host *sshconfig.Host)
+}
+
+type EnsureConfigDirective struct {
+	sshconfig.KV
+}
+
+func NewEnsureConfigDirective(key, value string) EnsureConfigDirective {
+	return EnsureConfigDirective{
+		KV: sshconfig.KV{
+			Key:   key,
+			Value: value,
+		},
 	}
 }
 
-func NewConfigDirective(key, value string) ConfigDirective {
-	return ConfigDirective{
-		Key:   key,
-		Value: value,
+func NewEnsureConfigDirectivePath(key, path string) EnsureConfigDirective {
+	return EnsureConfigDirective{
+		KV: sshconfig.KV{
+			Key:   key,
+			Value: filepath.ToSlash(path),
+		},
+	}
+}
+
+func (d EnsureConfigDirective) Apply(host *sshconfig.Host) {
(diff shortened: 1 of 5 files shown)


internal/ssh/config_file_test.go
```diff
@@ -148,3 +148,78 @@ func TestCheckForLegacyConfigEntries(t *testing.T) {
...
+		assert.Error(t, err)
+		assert.Contains(t, err.Error(), "no .conf files")
+	})
+
+	t.Run("concatenates conf files into unified config and removes directory", func(t *testing.T) {
+		tmp := t.TempDir()
+		testutil.SetHomeDir(t, tmp)
+		legacyDir := filepath.Join(tmp, ".ssh", "topo_config")
+		require.NoError(t, os.MkdirAll(legacyDir, 0o700))
+		require.NoError(t, os.WriteFile(filepath.Join(legacyDir, "topo_board1.conf"), []byte(`Host board1
```
muchzill4: Something I've just realised that we might be doing wrongly across the codebase: using `require.NoError` in arrange step, which should probably be more panick-y instead? `require/assert` produces a failure that looks like an assertion, when really it means the test itself is broken, not the code under test.

internal/ssh/config_file_test.go
```diff
@@ -148,3 +148,78 @@ func TestCheckForLegacyConfigEntries(t *testing.T) {
...
+		testutil.SetHomeDir(t, tmp)
+		require.NoError(t, os.MkdirAll(filepath.Join(tmp, ".ssh", "topo_config"), 0o700))
+
+		err := ssh.MigrateLegacyTopoConfig()
+
+		assert.Error(t, err)
+		assert.Contains(t, err.Error(), "no .conf files")
+	})
+
+	t.Run("concatenates conf files into unified config and removes directory", func(t *testing.T) {
```
muchzill4, line 158: A lot of noise in this test. Perhaps we can improve readability with something along those lines (assumes imaginary functions you'd need):

```go
tmp := t.TempDir()
testutil.SetHomeDir(t, tmp)
legacyDir := filepath.Join(tmp, ".ssh", "topo_config")
legacyDirSlash := filepath.ToSlash(legacyDir)
board1Conf := `Host board1
  IdentityFile ~/.ssh/key1
`
board2Conf := `Host board2
  IdentityFile ~/.ssh/key2
`
sshConf := "Include " + legacyDirSlash + `/*.conf

Host *
`
mustMkdirAll(t, legacyDir)
mustWriteFile(t, filepath.Join(legacyDir, "topo_board1.conf"), board1Conf)
mustWriteFile(t, filepath.Join(legacyDir, "topo_board2.conf"), board2Conf)
mustWriteFile(t, filepath.Join(tmp, ".ssh", "config"), sshConf)

require.NoError(t, ssh.MigrateLegacyTopoConfig())

mergedFile := legacyDirSlash
wantSshConfAfterMigration := "Include " + mergedFile + `

Host *
`
requireFileContents(t, mergedFile, board1Conf+board2Conf)
requireFileContents(t, filepath.Join(tmp, ".ssh", "config"), wantSshConfAfterMigration)
```

internal/ssh/config_file.go
```diff
@@ -107,8 +151,8 @@ func updateConfigFile(path string, host string, directives []ConfigDirective) er
 		return fmt.Errorf("failed to find or create host block: %w", err)
 	}
 
-	for _, directive := range directives {
-		addOrReplaceDirective(hostBlock, directive)
+	for _, modifier := range modifiers {
+		modifier.Apply(hostBlock)
```
muchzill4, line 153: Nice.

internal/ssh/config_file.go
```diff
@@ -155,8 +208,65 @@ func CheckForLegacyTopoConfigEntries() error {
...
 	return nil
 }
+
+func MigrateLegacyTopoConfig() error {
+	legacyDir, err := getConfigFilePath(topoConfigFileName)
+	if err != nil {
+		return err
+	}
+
+	if err := CheckForLegacyTopoConfigEntries(); err != errLegacyConfigEntries {
```
muchzill4: This `err !=` comparison is a bit confusing and tells me that instead of `CheckForLegacyTopoEntries()`, we should have `ListLegacyTopoConfigEntries() ([]string, error)`. Then each caller of `ListLegacyTopoConfigEntries`, can react accordingly. 

My main gripe is that `errLegacyConfigEntries` is either bubbled up or not, depending on the call site, while it seems like it should be specific to the `setup-keys` flow only.

muchzill4 (changes requested):

internal/ssh/config_file_test.go
awphi: Good point - we should probably have some `testutil.MustMkdirAll`, `testutil.MustWriteFile` helpers we can use. This is probably something $AI_AGENT_OF_YOUR_CHOOSING could do across the codebase quite quickly with easily verifiable results.

Happy if I leave as is for consistency now and address more widely later to separate the feat from the test refactoring?

awphi: As it turns out we already have `RequireWriteFile` so will use that here - it still does `require.NoError` but it's easier to change in that one place than all over the place

awphi: Bit the bullet and also added a `RequireMkdirAll` only used in this test package for now - can propagate across the codebase later https://github.com/arm/topo/pull/186/commits/b6b31f98f6c0a484cd0e9846b865e642c8a39bfe

internal/ssh/config_file_test.go
awphi: https://github.com/arm/topo/pull/186/commits/b6b31f98f6c0a484cd0e9846b865e642c8a39bfe

internal/ssh/config_file.go
awphi: I'm not sure it's worth it to have a `ListLegacyTopoConfigEntries` since we'd have to implement ssh config listing logic for a collection of config files that we never actually want to consume - we'd just be checking the length.

WDYT about changing the signature of `CheckForLegacyTopoConfigEntries` to 

```go
func LegacyTopoConfigDirectoryExists()(bool, error) 
```

to better separate the error case from the actual check? Then migrate only needs to consume the bool and `setup-keys` can own `errLegacyConfigEntries` and it doesn't even need to be a bastion error anymore .

Made this change here to expedite things if you agree https://github.com/arm/topo/pull/186/commits/4d4f1f64bdd1cd7bb4a58654d32bc0e6b2c00e03 - can revert if you don't!

muchzill4: We could use the config files listing in the `MigrateConfig` though, no? We could leverage that instead of globbing here. Are you saying that's doing unnecessary work for the other code path?

awphi: Ah sorry I got the wrong end of the stick - yes that makes sense to me

awphi: Actually this would no longer cover all possible cases right? If I had an empty `topo_config/` directory then a `ListLegacyTopoConfigEntries` function would return an empty slice, indicating there's nothing to migrate but we'd still want to delete that directory so the new file could be made in its place. We'd need some sentinel value to mark this as a special case which I'm bit iffy about or to always check for a directory and purge it even when `ListLegacyTopoConfigEntries` returns an empty slice which doesn't seem great either

internal/ssh/config_file.go
```diff
@@ -122,41 +164,106 @@ func updateConfigFile(path string, host string, directives []ConfigDirective) er
...
+	return false, nil
+}
+
+func MigrateLegacyTopoConfig() error {
+	legacyDir, err := getConfigFilePath(topoConfigFileName)
+	if err != nil {
+		return err
+	}
+
+	if exists, err := LegacyTopoConfigDirectoryExists(); err != nil {
```
muchzill4, line 221: Shouldn't `LegacyTopoConfigDirectoryExists` take `legacyDir` as argument? I'm slowly realising that the `testutil.SetHomeDir(t, tmp)` is our tests crying for help. Unsure if it's part of this PR.

muchzill4 (approved):

internal/ssh/config_file.go
awphi: Agreed it's becoming quite a strong smell. I'm keen to get this PR through as is to avoid ballooning further + unblock vscode work but have made a note to revisit!

awphi:
Spotted a small mistake in some final testing where `migrate-ssh` would fail on an empty `topo_config/` but its existence also prevented `setup-keys` so removed that check + corresponding test https://github.com/arm/topo/pull/186/commits/677fd5ef4621299e54b49fef8804121dc3a66f4b

PR #189, opened 2026-04-08
fix: allow Literal IP addresses in topo setup-keys

tgonzalezorlandoarm:
Commit 7bd3a2b23e0e9a92923d6df2347a9f6381623ac7 introduced a regression where `topo setup-keys --target <ip>` is not a valid target anymore:

```
topo setup-keys --target 10.1.197.67
ERR no explicit host config found for 10.1.197.67; note: a per user SSH config entry should be created when setting up keys
```

raw IPs should be accepted.

## Changes 

* Allow for literal IP addresses to be introduced.
* Unify the key file naming convention under `id_<keytype>_topo_<user>_<alias/ip>`.
* Start specifying the `User` directive in the config entry now that we are enforcing a per-user alias/config entry.


internal/ssh/config.go
```diff
@@ -57,6 +57,10 @@ func (c Config) ConnectTimeout(fallback time.Duration) time.Duration {
 }
 
 func IsDestinationAlreadyConfiguredWithAnotherUser(dest Destination) error {
+	if dest.IsIPLiteral() {
```
muchzill4: Why do we want to short circuit the `LookupExplicitHostConfig` check if an ip is given? 🤔 

Perhaps I'm reading this backwards, but why does ip destination get different treatment from non-ip destination if it comes to checking pre-existing values that are potentially associated with a different user?

tgonzalezorlandoarm: done

tgonzalezorlandoarm:
The following comment is for a different PR, but what I'm thinking about now is the edge case in which you have a Host configured in an ssh config but then you reference it with the IP instead of the Host name.

internal/ssh/config.go
```diff
@@ -41,31 +41,35 @@ func NewConfigFromBytes(data []byte) Config {
 	return config
 }
 
-func IsDestinationAlreadyConfiguredWithAnotherUser(dest Destination) error {
-	hostConfig, err := LookupExplicitHostConfig(dest.Host, dest.Port)
+// IsDestinationAlreadyConfiguredWithAnotherUser checks if the destination is already configured with another user.
+// If it is, an error is returned. Otherwise, the user associated with the destination in the SSH config is returned.
+func IsDestinationAlreadyConfiguredWithAnotherUser(dest Destination) (string, error) {
```
yejseo01: `IsDestinationAlreadyConfiguredWithAnotherUser` sounds like a function that would return boolean only. Can we rename this function to `GetUser` instead?

internal/ssh/config.go
```diff
@@ -41,31 +41,35 @@ func NewConfigFromBytes(data []byte) Config {
 	return config
 }
 
-func IsDestinationAlreadyConfiguredWithAnotherUser(dest Destination) error {
-	hostConfig, err := LookupExplicitHostConfig(dest.Host, dest.Port)
+// IsDestinationAlreadyConfiguredWithAnotherUser checks if the destination is already configured with another user.
+// If it is, an error is returned. Otherwise, the user associated with the destination in the SSH config is returned.
```
yejseo01: I don't think this comment is necessary when the function is renamed

internal/ssh/config.go
```diff
@@ -41,31 +41,35 @@ func NewConfigFromBytes(data []byte) Config {
...
+				hostConfig.User,
+			)
+		}
+	} else if !isExplicitHostConfig && !dest.IsIPLiteral() {
+		return "", fmt.Errorf("no explicit host config found for %s", dest.Host)
+	} else if !isExplicitHostConfig && dest.IsIPLiteral() {
+		if dest.User != "" {
+			return dest.User, nil
+		}
+		return hostConfig.User, nil
```
yejseo01:
```suggestion
	if isExplicitHostConfig {
		if dest.User != "" && hostConfig.User != dest.User {
			return "", fmt.Errorf(
				"ssh host/alias %q is already associated with user %q",
				dest.Host,
				hostConfig.User,
			)
		}
	} else if dest.IsIPLiteral() {
		if dest.User != "" {
			return dest.User, nil
		}
		return hostConfig.User, nil
	} else {
		return "", fmt.Errorf("no explicit host config found for %s", dest.Host)
```

internal/ssh/config_resolve_test.go
yejseo01, line 1: does this belong in config_test.go?

Luke-Parkin (approved):

internal/ssh/config_resolve_test.go
```diff
@@ -0,0 +1,101 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestResolveConfiguredUser(t *testing.T) {
+	t.Run("IP literal with no explicit config returns dest user", func(t *testing.T) {
+		config := []byte(`debug1: /etc/ssh/ssh_config line 57: Applying options for *
+user lukpar01
```
yejseo01: maybe not the best to embed the actual user name in the test examples

Luke-Parkin: agreed

internal/ssh/config.go
Luke-Parkin: Good name, I was trying to think of one but failed.

Luke-Parkin: https://github.com/arm/topo/pull/189/commits/43dc76bf770d877f34983a9926953a9bd6320733

internal/ssh/config.go
Luke-Parkin: https://github.com/arm/topo/pull/189/commits/43dc76bf770d877f34983a9926953a9bd6320733

internal/ssh/config_resolve_test.go
Luke-Parkin: https://github.com/arm/topo/pull/189/commits/93674eeda319ce33d6cf771a7895f36048c493c8

internal/ssh/config_test.go
```diff
@@ -56,6 +56,101 @@ user homer
...
+
+		assert.NoError(t, err)
+		assert.Equal(t, "root", got)
+	})
+
+	t.Run("explicit config with no user conflict returns config user", func(t *testing.T) {
+		config := []byte(`debug1: /tmp/.ssh/topo_config line 1: Applying options for board-alias
+user root
+hostname 10.2.2.26
+debug1: /etc/ssh/ssh_config line 57: Applying options for *
```
muchzill4: I think `debug1:` is printed first in real output.

internal/ssh/config.go
```diff
@@ -41,31 +41,32 @@ func NewConfigFromBytes(data []byte) Config {
...
 }
 
-// LookupExplicitHostConfig returns configuration, only if there's an explicit entry for the given host/port
-func LookupExplicitHostConfig(host, port string) (Config, error) {
-	dest := Destination{Host: host, Port: port}
-	output, err := readConfig(dest)
-	if err != nil {
-		return Config{}, err
-	}
+func ResolveConfiguredUser(dest Destination, configOutput []byte) (string, error) {
```
yejseo01, line 52: does this need to be public function? it seems like it's used only in this file?

internal/ssh/config_test.go
```diff
@@ -56,6 +56,101 @@ user homer
 	})
 }
 
+func TestResolveConfiguredUser(t *testing.T) {
```
yejseo01, line 59: I don't think `ResolveConfiguredUser` needs to be public, and we probably don't need test it if it's not used outside the `config.go`

internal/ssh/config.go
Luke-Parkin: Yes, the test file is in a different package.

internal/ssh/config_test.go
Luke-Parkin: Considering we had a pretty major regression and missed it, I think tests are needed here.

yejseo01: what did we miss?

internal/ssh/config.go
yejseo01: but the test is not necessary if the function is private? I meant there's no usage of this function besides from the test for itself.

internal/ssh/config_test.go
```diff
@@ -56,6 +56,101 @@ user homer
...
+	t.Run("IP literal with no explicit config returns dest user", func(t *testing.T) {
```
muchzill4: Let's cleanup the test cases a bit, please double check if I haven't missed something (I think you can get away with 6 instead of 7):

```go
t.Run("with explicit configuration", func() {
   t.Run("it returns destination's user when no user is configured")
   t.Run("it returns destination's user when user is configured as the same user")
   t.Run("it returns configured user when destination's user is not set")
   t.Run("it errors if destination's user doesn't match configured user")
})

t.Run("with catch-all configuration", func() { // <- is there a better name
   t.Run("returns destination's user when it's set")
   t.Run("returns configured user when destination's user is not set")
})
```

muchzill4 (commented):
Thanks for taking this over!

internal/ssh/config_test.go
Luke-Parkin: Read the PR description!

internal/ssh/config.go
Luke-Parkin: It's used on line 49 in config.go, in GetUserFromConfig, which is used in SetupKeys.

yejseo01: That's still inside the same file, so it still can be a private function. Anyway, if the tests brings us good values overall, I'm okay with keeping it public and having the tests.

internal/ssh/config_test.go
Luke-Parkin: https://github.com/arm/topo/pull/189/commits/47a574c6f4b128a8e70085dbb0d9f5b3bfea0f13

internal/ssh/config_test.go
Luke-Parkin: https://github.com/arm/topo/pull/189/commits/47a574c6f4b128a8e70085dbb0d9f5b3bfea0f13

muchzill4 (approved):

yejseo01 (approved):
I want to approve too!

PR #192, opened 2026-04-09
fix: stop prefixing commands with `sh -c` in local runner

muchzill4:
## Changes

- 🐛 Makes `Local` runner usable on Windows, where we don't have access to `sh -c`
- Local runner now uses `shlex.Split` to parse command strings and calls `exec.Command` directly, instead of wrapping every command in `/bin/sh -c`
- Adds `google/shlex` dependency
- Adds tests for local runner `Run`


muchzill4:
As an aside I'd like to re-think the runner interface, I'm not sure the API is _good_. I'd rather we had a way to avoid adding more and more methods to pass stdin, ssh args and combination of both.

Luke-Parkin (approved):

PR #193, opened 2026-04-09
feat: list-candidate-targets command

awphi:
## Changes

- Adds a new hidden vscode command; `list-candidate-targets` which takes a path to an ssh config file and spits out a de-duplicated list of `Host` entries found in said config, recursively evaluating the contents of `Include`d files.
  - Only spits out the `Host` as the most reliable way to resolve ssh directives of a given host is with `ssh -G` rather than reading the config.
  - VSCode will use this to populate a list of suggested ssh targets to add to your target list in the UI. It's exposed by the CLI to avoid taking on another dependency in the extension and to isolate all sshconfig manipulation/traversal to the CLI.
- Adds a new `mapset`-like data structure to support cycle detection. There's a few other places in the codebase that could make use of this DS too so will follow up.

### DIFF
--- a/cmd/topo/vscode.go
+++ b/cmd/topo/vscode.go
@@ -22,9 +22,9 @@ var getProjectCmd = &cobra.Command{
 	},
 }
 
-var listSsh = &cobra.Command{
-	Use:    "list-ssh <config-filepath>",
-	Short:  "Prints a list of ssh hosts defined in the given config file as JSON",
+var listCandidateTargets = &cobra.Command{
+	Use:    "list-candidate-targets <config-filepath>",
+	Short:  "Prints a list of candidate ssh targets defined in the given config file as JSON",
 	Hidden: true,
 	Args:   cobra.ExactArgs(1),
 	RunE: func(cmd *cobra.Command, args []string) error {
@@ -43,5 +43,5 @@ var listSsh = &cobra.Command{
 
 func init() {
 	rootCmd.AddCommand(getProjectCmd)
-	rootCmd.AddCommand(listSsh)
+	rootCmd.AddCommand(listCandidateTargets)
 }
--- a/cmd/topo/vscode.go
+++ b/cmd/topo/vscode.go
@@ -1,8 +1,11 @@
 package main
 
 import (
+	"encoding/json"
+	"fmt"
 	"os"
 
+	"github.com/arm/topo/internal/ssh"
 	"github.com/arm/topo/internal/vscode"
 	"github.com/spf13/cobra"
 )
@@ -19,6 +22,26 @@ var getProjectCmd = &cobra.Command{
 	},
 }
 
+var listSsh = &cobra.Command{
+	Use:    "list-ssh <config-filepath>",
+	Short:  "Prints a list of ssh hosts defined in the given config file as JSON",
+	Hidden: true,
+	Args:   cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		cmd.SilenceUsage = true
+		configPath := args[0]
+
+		hosts := ssh.ListHosts(configPath)
+		data, err := json.Marshal(hosts)
+		if err != nil {
+			return fmt.Errorf("failed to marshal ssh hosts: %w", err)
+		}
+		_, err = fmt.Fprintf(os.Stdout, "%s\n", data)
+		return err
+	},
+}
+
 func init() {
 	rootCmd.AddCommand(getProjectCmd)
+	rootCmd.AddCommand(listSsh)
 }
--- a/internal/collections/set.go
+++ b/internal/collections/set.go
@@ -1,5 +1,10 @@
 package collections
 
+import (
+	"maps"
+	"slices"
+)
+
 type Set[T comparable] struct {
 	elements map[T]struct{}
 }
@@ -17,11 +22,7 @@ func (s *Set[T]) Add(item T) {
 }
 
 func (s *Set[T]) ToSlice() []T {
-	slice := make([]T, 0, len(s.elements))
-	for item := range s.elements {
-		slice = append(slice, item)
-	}
-	return slice
+	return slices.Collect(maps.Keys(s.elements))
 }
 
 func (s *Set[T]) Contains(item T) bool {
--- a/internal/collections/set.go
+++ b/internal/collections/set.go
@@ -1,0 +1,30 @@
+package collections
+
+type Set[T comparable] struct {
+	elements map[T]struct{}
+}
+
+func NewSet[T comparable](items ...T) Set[T] {
+	s := Set[T]{elements: make(map[T]struct{})}
+	for _, item := range items {
+		s.Add(item)
+	}
+	return s
+}
+
+func (s *Set[T]) Add(item T) {
+	s.elements[item] = struct{}{}
+}
+
+func (s *Set[T]) ToSlice() []T {
+	slice := make([]T, 0, len(s.elements))
+	for item := range s.elements {
+		slice = append(slice, item)
+	}
+	return slice
+}
+
+func (s *Set[T]) Contains(item T) bool {
+	_, exists := s.elements[item]
+	return exists
+}
--- a/internal/ssh/list.go
+++ b/internal/ssh/list.go
@@ -1,0 +1,61 @@
+package ssh
+
+import (
+	"strings"
+
+	"github.com/arm/topo/internal/collections"
+	"github.com/arm/topo/internal/output/logger"
+	sshconfig "github.com/kevinburke/ssh_config"
+)
+
+func gatherIncludedConfigPaths(cfg *sshconfig.Config) []string {
+	includedPaths := []string{}
+
+	for _, host := range cfg.Hosts {
+		for _, node := range host.Nodes {
+			if include, ok := node.(*sshconfig.Include); ok {
+				includePath := strings.TrimSpace(strings.TrimPrefix(include.String(), "Include"))
+				if includePath != "" {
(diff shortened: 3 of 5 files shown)


internal/collections/set.go
```diff
@@ -0,0 +1,30 @@
...
+func (s *Set[T]) Add(item T) {
+	s.elements[item] = struct{}{}
+}
+
+func (s *Set[T]) ToSlice() []T {
+	slice := make([]T, 0, len(s.elements))
+	for item := range s.elements {
+		slice = append(slice, item)
+	}
+	return slice
```
muchzill4:
```suggestion
	return slices.Collect(maps.Keys(s.elements))
```
So modern 🧑‍🍳.
 
⚠️ Requires `maps` and `slices` import.

cmd/topo/vscode.go
```diff
@@ -19,6 +22,26 @@ var getProjectCmd = &cobra.Command{
 	},
 }
 
+var listSsh = &cobra.Command{
+	Use:    "list-ssh <config-filepath>",
```
muchzill4: Should this handle any config path or just topo config path? If former then I'm not sure it belongs in `topo`. 🤔

muchzill4: Should it be `list-candidate-targets` / `list-targets`?

muchzill4 (commented):
Tests and code lovely, but I'm not sure about this as a functionality. 

Calling @th3james for opinion.

awphi:
> I'm not sure about this as a functionality.

I'll try to add some more context to plead my case - in the vscode extension we want to support a similar workflow to the "Remote - SSH" extension where when you go to add a target we display a list of all your configured ssh hosts to pick as a convenience feature:

Screenshot 2026-04-10 at 12 33 42

Now, the reason I did this in the CLI as opposed to the vscode extension is to avoid taking on another dependency on the vscode side for sshconfig parsing when we've already got one on the CLI side. However, since its only consumer (at least for now) is the vscode extension it's just a hidden command like `get-project`. I could also see this functionality being useful for a TUI on the CLI side but that's still very up in the air right now.

Happy to walk this back and re-implement in TS with https://www.npmjs.com/package/ssh-config or basic string manipulation if there's agreement this doesn't belong in the CLI.

cmd/topo/vscode.go
awphi: The intention would be to invoke it on `~/.ssh/config`. Feels like this is something ssh should be able to do on its own but alas. Totally agree it's not a specific problem - outlined why I did this in the CLI here; https://github.com/arm/topo/pull/193#issuecomment-4223489491

The only reason I exposed it as a parameter is in case that requirement changes on the vscode side we can avoid having to make CLI changes.

internal/collections/set.go
awphi: https://github.com/arm/topo/pull/193/commits/14b42e3ad1f22ab810b29bb391a8204609723180

cmd/topo/vscode.go
awphi: I like that, more flexible. I know @thegecko was keen on discovering potential targets from other sources beyond just sshconfig - https://github.com/arm/topo/pull/193/commits/309c1e5e135bfe252285b26559574ba9d79c6cfd

thegecko:
I wonder if this shouldn't be 'hidden' in cases third parties want to drive their own UI from Topo?

internal/ssh/list.go
```diff
@@ -0,0 +1,61 @@
...
+	sshconfig "github.com/kevinburke/ssh_config"
+)
+
+func gatherIncludedConfigPaths(cfg *sshconfig.Config) []string {
+	includedPaths := []string{}
+
+	for _, host := range cfg.Hosts {
+		for _, node := range host.Nodes {
+			if include, ok := node.(*sshconfig.Include); ok {
+				includePath := strings.TrimSpace(strings.TrimPrefix(include.String(), "Include"))
```
th3james, line 17: Unfortunately the spec is more complicated than this:

> Multiple path names may be specified and each pathname may contain glob(7) wildcards, tokens as described in the “TOKENS” section, environment variables as described in the “ENVIRONMENT VARIABLES” section and, for user configurations, shell-like ‘~’ references to user home directories.  Wildcards will be expanded and processed in lexical order.

https://man7.org/linux/man-pages/man5/ssh_config.5.html

It would appear that the `ssh_config` library can't help us here either.

This is all making me a bit nervous about whether we're taking on responsibilities that aren't core to Topo. At the end of the day, we have a dependency on people having at least a cursory understanding of SSH. One could argue that anyone who has such a fancy config may not need our help (particularly WRT ENV vars). Either way the current implementation cannot guarantee that it will return all targets across all includes paths.

Are we sure the juice is worth the squeeze in terms of the use case here? Do we intend to list _all_ SSH hosts in the VS Code extension? if there's no use case in the CLI I wonder if this might be better leveraging an SSH extension in VS Code?

awphi: > It would appear that the ssh_config library can't help us here either.

So I've been digging into `ssh_config` a bit more since raising this PR and it seems it already recursively resolves `Include` statements before in a spec-compliant manner so my incomplete version here could be removed. HOWEVER, it does not expose a way to list all the hosts from those `Include`d files (e.g. see open PR adding this here https://github.com/kevinburke/ssh_config/pull/78). Depending on your thoughts on my answer to your second question I'd suggest vendoring this dep and modifying as needed - there's a lot we can strip from it that we're not using and we can cherry-pick in the `GetHosts` method.

> Are we sure the juice is worth the squeeze in terms of the use case here? Do we intend to list all SSH hosts in the VS Code extension?

Imo yes, I'm confident editing my own ssh config but I still like the convenience of the quick pick menu in vscode to quickly ssh into a machine without having to type anything and/or remember my `Host` alias. 

However, if you're really not sure about the value here I'm happy to abandon this PR and we can drop sshconfig target discovery altogether.

>  if there's no use case in the CLI 

I agree there's absolutely no use case in the CLI for this today but I don't know if this will hold forever e.g. we might want similar behaviour in a TUI or if someone vibes up a `lazytopo` or similar (linking back to what Rob said here; https://github.com/arm/topo/pull/193#issuecomment-4223550511). The reason it lives in the CLI today is because we already have an sshconfig-parsing dependency here so may as well utilise it rather than adding another on the vscode side.

> I wonder if this might be better leveraging an SSH extension in VS Code?

This was indeed my first preference and @federicobozzini's initial suggestion but unfortunately the de-facto standard (or potentially just standard it seems) SSH extension in vscode is ["Remote SSH" by Microsoft](https://github.com/microsoft/vscode-remote-release/tree/main/ssh) which is closed-source and exposes no public API we could hook into.

cmd/topo/vscode.go
```diff
@@ -19,6 +22,26 @@ var getProjectCmd = &cobra.Command{
...
+var listCandidateTargets = &cobra.Command{
+	Use:    "list-candidate-targets <config-filepath>",
+	Short:  "Prints a list of candidate ssh targets defined in the given config file as JSON",
+	Hidden: true,
+	Args:   cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		cmd.SilenceUsage = true
+		configPath := args[0]
+
+		hosts := ssh.ListHosts(configPath)
```
yejseo01, line 34: nit: `targets` might be better. Trying to be cautious on mixing the usage of `host` and `target` within the code. I think we use `host` for docker context but `target` for all others

internal/ssh/list.go
yejseo01: > I agree there's absolutely no use case in the CLI for this today but I don't know if this will hold forever 

Is it YAGNI for CLI then?

awphi: > Is it YAGNI for CLI then?

If we don't mind taking on an npm dependency on the vscode side and duplicating some sshconfig parsing code then yes

federicobozzini: > Do we intend to list all SSH hosts in the VS Code extension?

Just to clarify, this PR doesn't list all SSH hosts, only the one in the config path used as an argument. This is also the topo-specific bit, since we plan to have all the topo configs in a separate file.

muchzill4: > Just to clarify, this PR doesn't list all SSH hosts, only the one in the config path used as an argument. This is also the topo-specific bit, since we plan to have all the topo configs in a separate file.

If you are after something that lists all topo hosts, which config is _owned_ by topo, then the ssh config path shouldn't be configurable, as topo's internals know the path already.

awphi: We'll need to know all hosts, not just ones in the topo config file so we can support adding externally managed hosts as targets in the extension. That is to say we're probably going to invoke `topo list-candidate-targets ~/.ssh/config` which will also list all topo-managed targets by virtue of the `Include` statement.

It's configurable in case the user is using an unorthodox ssh config file path they'd like to consume in the vscode extension but maybe we shouldn't care about that.

Important to remember this command is just for the discovery drop-down when adding a new target to the extension and won't be the source of truth for the actual target list. The diagram on the jira ticket is worth referring to here.

federicobozzini: Right. I would agree with Rob that this command should either live in the `vscode` namespace and be hidden  or implemented directly in VS Code. It doesn't seem to be topo-specific enough to live in the public API.

awphi: It is hidden right now, Rob was arguing in favour of making it public. `ListHosts` could certainly be moved to the `vscode` package though but it makes little difference given they're both under `internal/`

awphi:
As discussed offline, we're going to kick the can down the road on this - given there's only one consumer of this functionality atm it probably doesn't warrant being in the CLI.

PR #194, opened 2026-04-09
test: commonise file-based testutils

awphi:
## Changes

- Makes use of our existing file/dir-writing `testutil` functions across the test suites to reduce noise
- Adds a new `testutil.RequireWriteFile` to round out the trio of read file/write file/mkdir


muchzill4 (approved):
Nice!

PR #195, opened 2026-04-09
refactor: simplify SSH authentication probe

muchzill4:
## Changes

- Extract pure functions from `SSHAuthenticationProbe`: `BuildProbeArgs` (options → SSH args) tested with table-driven tests, no mocks
- Remove the separate known host check — host key verification is now handled by `StrictHostKeyChecking=yes` (or `accept-new`) directly on the pubkey probe
- Remove the password fallback — the old password probe used `BatchMode=yes` with `NumberOfPasswordPrompts=0`, meaning it could never actually authenticate with a password. If pubkey fails with an auth error, the fix is `topo setup-keys` regardless
- Remove `ErrPasswordAuthentication` — health check now matches on `ErrAuthFailed` directly to recommend `topo setup-keys`
- Replace `sshRunnerWithExtraArgs` interface with concrete `*runner.SSH`
- Consolidate SSH error classification into `ClassifyStderr` — host key, auth, timeout, and connection patterns all live in one place. The probe translates `ssh.Err*` into probe-domain errors via `errors.Is`, no string re-parsing
- Eliminate `testify/mock` from the test file entirely

Net result: 1 SSH call instead of up to 3, ~100 fewer lines, zero mocks.

### DIFF
--- a/internal/ssh/errors.go
+++ b/internal/ssh/errors.go
@@ -11,19 +11,29 @@ var (
 	ErrAuthFailed        = fmt.Errorf("%w: authentication failed", ErrSSH)
 	ErrConnectionFailed  = fmt.Errorf("%w: connection failed", ErrSSH)
 	ErrConnectionTimeout = fmt.Errorf("%w: connection timed out", ErrSSH)
+	ErrHostKeyUnknown    = fmt.Errorf("%w: host key is not known", ErrSSH)
+	ErrHostKeyChanged    = fmt.Errorf("%w: host key has changed", ErrSSH)
 )
 
 // ClassifyStderr inspects SSH stderr output and returns a typed error when a
 // known failure pattern is detected, or nil if the output is unrecognised.
 func ClassifyStderr(stderr string) error {
 	lower := strings.ToLower(stderr)
+	if strings.Contains(lower, "host key verification failed") {
+		if strings.Contains(lower, "has changed") {
+			return ErrHostKeyChanged
+		}
+		return ErrHostKeyUnknown
+	}
 	if strings.Contains(lower, "timed out") ||
 		strings.Contains(lower, "connection timeout") ||
 		strings.Contains(lower, "did not properly respond after a period of time") {
 		return ErrConnectionTimeout
 	}
 	if strings.Contains(lower, "publickey") ||
-		strings.Contains(lower, "authentication") {
+		strings.Contains(lower, "authentication") ||
+		strings.Contains(lower, "permission denied") ||
+		strings.Contains(lower, "password") {
 		return ErrAuthFailed
 	}
 	if strings.Contains(lower, "connection refused") {
--- a/internal/ssh/errors.go
+++ b/internal/ssh/errors.go
@@ -25,15 +25,10 @@ func ClassifyStderr(stderr string) error {
 		}
 		return ErrHostKeyUnknown
 	}
-	if strings.Contains(lower, "timed out") ||
-		strings.Contains(lower, "connection timeout") ||
-		strings.Contains(lower, "did not properly respond after a period of time") {
+	if strings.Contains(lower, "timed out") {
 		return ErrConnectionTimeout
 	}
-	if strings.Contains(lower, "publickey") ||
-		strings.Contains(lower, "authentication") ||
-		strings.Contains(lower, "permission denied") ||
-		strings.Contains(lower, "password") {
+	if strings.Contains(lower, "permission denied") {
 		return ErrAuthFailed
 	}
 	if strings.Contains(lower, "connection refused") {
--- a/internal/target/ssh_authentication_probe.go
+++ b/internal/target/ssh_authentication_probe.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"errors"
 	"strings"
+
+	"github.com/arm/topo/internal/runner"
 )
 
 var (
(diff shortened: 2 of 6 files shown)


internal/target/ssh_authentication_probe.go
```diff
@@ -3,122 +3,60 @@ package target
...
+	switch {
+	case errors.Is(err, ssh.ErrHostKeyChanged):
+		return ErrHostKeyChanged
+	case errors.Is(err, ssh.ErrHostKeyUnknown):
+		return ErrHostKeyUnknown
+	case errors.Is(err, ssh.ErrAuthFailed):
+		return ErrAuthFailed
+	default:
 		return err
 	}
```
awphi, line 60: Can't we just re-use the `ssh` bastion errors so this collapses down to
```go
func (p SSHAuthenticationProbe) Probe(ctx context.Context) error {
	_, err := p.runner.RunWithArgs(ctx, "true", BuildProbeArgs(p.opts.AcceptNewHostKeys)...)
	return err
}
```
would require changes in `health.go` to lean on those same errors but I'm not sure that's a bad thing

internal/target/ssh_authentication_probe.go
```diff
@@ -3,122 +3,60 @@ package target
...
-		return nil
-	}
-	output := strings.ToLower(out)
-	if strings.Contains(output, "host key verification failed") {
-		if strings.Contains(output, "has changed") {
-			return ErrHostKeyChanged
-		}
-		return ErrHostKeyNew
+// BuildProbeArgs returns SSH arguments for a public key authentication probe.
+func BuildProbeArgs(acceptNewHostKeys bool) []string {
```
awphi: ~~Personally not a fan of comments that could be covered by clearer function names - WDYT about the `target.BuildSSHAuthProbeArgs` or similar?~~

Actually maybe could be a method on`SSHAuthenticationProbeOptions` - can't see a reason not to tightly couple and removes the need for a comment or verbose name?

awphi (approved):
Couple suggestions but nothing blocking again - very nice work! Might be worth getting @tgonzalezorlandoarm if he has a chance before heading away as a second opinion since he's more knowledgable about possible gaps with these probes.

internal/ssh/errors.go
```diff
@@ -11,19 +11,29 @@ var (
...
 	if strings.Contains(lower, "timed out") ||
 		strings.Contains(lower, "connection timeout") ||
 		strings.Contains(lower, "did not properly respond after a period of time") {
 		return ErrConnectionTimeout
 	}
 	if strings.Contains(lower, "publickey") ||
-		strings.Contains(lower, "authentication") {
+		strings.Contains(lower, "authentication") ||
+		strings.Contains(lower, "permission denied") ||
+		strings.Contains(lower, "password") {
```
th3james: Is this potentially concise enough to risk matching a different error? Based on the tests this would be more accurate
```suggestion
		strings.Contains(lower, "password:") {
```

th3james (approved):

internal/target/ssh_authentication_probe.go
```diff
@@ -3,122 +3,60 @@ package target
...
 )
 
 var (
-	ErrPasswordAuthentication = errors.New("key-based SSH authentication is not setup")
-	ErrHostKeyNew             = errors.New("ssh host key is not known")
-	ErrHostKeyChanged         = errors.New("ssh host key has changed")
-	ErrAuthenticationFailure  = errors.New("ssh authentication failed")
+	ErrHostKeyUnknown = errors.New("ssh host key is not known")
+	ErrHostKeyChanged = errors.New("ssh host key has changed")
+	ErrAuthFailed     = errors.New("ssh authentication failed")
```
th3james:
```suggestion
	ErrHostKeyUnknown = errors.New("SSH host key is not known")
	ErrHostKeyChanged = errors.New("SSH host key has changed")
	ErrAuthFailed     = errors.New("SSH authentication failed")
```

internal/target/ssh_authentication_probe.go
muchzill4: Funny you suggest that. I've evaluated exactly what you're saying, but ended up preferring call site relying on "probe domain" errors. Even if they're duplicates. Perhaps you're right that it isn't a bad thing.

th3james (commented):
While we're changing these lines, consider fixing
> Error strings should not be capitalized (unless beginning with proper nouns or acronyms) 

https://go.dev/wiki/CodeReviewComments#error-strings

internal/target/ssh_authentication_probe.go
muchzill4: https://github.com/arm/topo/pull/195/commits/aecf44c281f160ecde34686893dc919d82994e99

internal/target/ssh_authentication_probe.go
muchzill4: https://github.com/arm/topo/pull/195/commits/0107368bf9ff0226672a8a3102e17d5d0521ef25

internal/ssh/errors.go
muchzill4: https://github.com/arm/topo/pull/195/commits/3a5598887cc9d7f8c15f99093cb15609567e4f3d

internal/target/ssh_authentication_probe.go
muchzill4: I'll leave it as is.

PR #196, opened 2026-04-10
refactor: explicit passing of ssh directories

awphi:
## Changes

- Replaces the implicit dependency many of our sshconfig-manipulating functions have on the user's home directory with an explicit one in the form of a function parameter. Why?
  - Allows us to eliminate `testutil.SetHomeDir` - one less thing to forget about when writing tests + denoises them a bit
  - Explicit dependency good, implicit bad 😠 
  -  Hoists the repeated decision of where the user's ssh config dir is located up the stack

### DIFF
--- a/internal/ssh/config_file.go
+++ b/internal/ssh/config_file.go
@@ -185,7 +185,7 @@ func CreateOrModifyConfigFile(sshDir string, dest Destination, modifiers []Confi
 	})
 }
 
-func LegacyTopoConfigDirectoryExists(sshDir string) (bool, error) {
+func IsLegacyTopoConfigDirectory(sshDir string) (bool, error) {
 	info, err := os.Stat(filepath.Join(sshDir, TopoConfigFileName))
 	if err != nil {
 		if os.IsNotExist(err) {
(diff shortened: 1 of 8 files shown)


internal/ssh/config_file.go
```diff
@@ -212,13 +201,9 @@ func LegacyTopoConfigDirectoryExists() (bool, error) {
...
-func MigrateLegacyTopoConfig() error {
-	legacyDir, err := getConfigFilePath(topoConfigFileName)
-	if err != nil {
-		return err
-	}
-
-	if exists, err := LegacyTopoConfigDirectoryExists(); err != nil {
+func MigrateLegacyTopoConfig(sshDir string) error {
+	legacyDir := filepath.Join(sshDir, TopoConfigFileName)
+	if exists, err := LegacyTopoConfigDirectoryExists(sshDir); err != nil {
```
muchzill4: Thoughts?

```suggestion
	if isLegacyDir, err := IsLegacyTopoConfigDirectory(legacyDir); err != nil {
```

awphi: https://github.com/arm/topo/pull/196/commits/2dd6eb2789a89b0ee3a27350da05fcb6f50da1e9

muchzill4 (approved):

PR #198, opened 2026-04-14
refactor: introduce StartContainer with pluggable container specs

muchzill4:
## Changes

- Replace `StartTargetContainer` with `StartContainer(t, spec)` that accepts a `ContainerSpec`
    - ℹ️ I'm planning to leverage the `ContainerSpec` for e2e testing of `setup-keys` which will be a separate PR
    - `DinDContainer` is provided as the first spec, with setup/cleanup hooks for SSH host key management and Docker daemon readiness
- Images are auto-built from Dockerfiles, removing the manual `docker build` prerequisite from CI and local development
- Readiness uses TCP port check (`waitForPort`) instead of SSH-based polling
- Host key acceptance is explicit with cleanup on test teardown, instead of being a hidden side effect of the readiness probe
- `TargetContainer` renamed to `Container`
- Low-level functions decoupled from `*testing.T` — they return errors instead
- Updated `docs/DEVELOPMENT.md` to reflect that container images are built automatically

### DIFF
--- a/internal/testutil/container.go
+++ b/internal/testutil/container.go
@@ -15,7 +15,7 @@ import (
 
 type Container struct {
 	SSHDestination string
-	ContainerName  string
+	Name           string
 }
 
 type ContainerSpec struct {
@@ -72,7 +72,7 @@ func StartContainer(t *testing.T, spec ContainerSpec) *Container {
 
 	c := &Container{
 		SSHDestination: fmt.Sprintf("ssh://root@localhost:%s", port),
-		ContainerName:  containerName,
+		Name:           containerName,
 	}
 
 	if spec.setup != nil {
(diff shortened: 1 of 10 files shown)


internal/testutil/container.go
```diff
@@ -3,80 +3,140 @@ package testutil
...
+	image   string
+	runArgs []string
+	setup   func(c *Container) error
+	cleanup func(c *Container)
+}
 
-type TargetContainer struct {
-	SSHDestination string
-	ContainerName  string
+var DinDContainer = ContainerSpec{
```
yejseo01, line 29: DinD probably mean Docker in Docker right?

yejseo01: Can we just call it `DockerInDockerContainer`?

yejseo01 (approved):
Minor question regarding abbreviation in variable name, but Looks good to me in general.

internal/testutil/container.go
muchzill4: Aligns with https://github.com/muchzill4/topo/blob/90fecdbb051ed9896d483a3b6bf0696f09f8e938/internal/testutil/test-container/Dockerfile#L1.

PR #199, opened 2026-04-14
feat: improve ssh tunnel errors

awphi:
## Changes

- Adds a new `command.RunCommand` and `command.StartCommand` utility to execute `exec.Cmd`s, capturing their stderr into a buffer and surfacing a well-structured, consistent error if the command fails
  - Generic error messages now contain the causal error, command that was executed and its stderr so everything is there on a single line for better debuggability
  - ^ this should address the case where tunnel teardown fails for unknown reasons - have not been able to repro thus far in normal circumstances so unclear what if any we'd need to check if an error state to surface a better message
- Adds an explicit check in the error state for starting the ssh tunnel to check if the port is taken. This only runs in the unhappy path to avoid unneeded delays when the ssh tunnel just works(tm)
- Swaps the teardown failure log message to an error from a warning so it's more in-your-face that it broke


internal/ssh/ssh_tunnel.go
```diff
@@ -4,17 +4,32 @@ import (
...
 
+	"github.com/arm/topo/internal/command"
 	"github.com/arm/topo/internal/operation"
+	"github.com/arm/topo/internal/output/logger"
 )
 
 const TunnelPIDPlaceholder = "<ssh tunnel pid>"
 
+func isPortTaken(port string) bool {
+	conn, err := net.DialTimeout("tcp", fmt.Sprintf("127.0.0.1:%s", port), 2*time.Second)
```
muchzill4, line 22: Isn't 2 seconds too generous?

internal/ssh/ssh_tunnel.go
```diff
@@ -69,12 +84,15 @@ func (s *SSHTunnelStart) Run(w io.Writer) error {
 	cmd := s.Command()
 	cmd.Stdout = w
 	cmd.Stderr = w
```
muchzill4, line 86: This is now meaningless, as the `Start|RunCommand` will overwrite both `Stdout` and `Stderr`, right?

internal/command/exec.go
```diff
@@ -0,0 +1,45 @@
...
+	return execute(cmd, cmd.Start, w)
+}
+
+func RunCommand(cmd *exec.Cmd, w io.Writer) error {
+	return execute(cmd, cmd.Run, w)
+}
+
+func newExecutionError(cmd *exec.Cmd, stderr *bytes.Buffer, err error) error {
+	cmdStr := strings.Join(cmd.Args, " ")
+	return fmt.Errorf("%s failed: %w | stderr: %s", cmdStr, err, stderr.String())
```
muchzill4: If `cmd.Start()` fails, `stderr` will be empty.

internal/ssh/ssh_tunnel.go
```diff
@@ -192,11 +201,8 @@ func (s *SSHTunnelProcessStop) Run(w io.Writer) error {
 		return nil
 	}
 
-	cmd := s.Command()
-	cmd.Stdout = w
-	cmd.Stderr = w
-	if err := cmd.Run(); err != nil {
-		return fmt.Errorf("failed to stop SSH tunnel process: %d", s.Start.Process.Pid)
+	if err := command.RunCommand(s.Command(), w); err != nil {
+		return fmt.Errorf("stopping SSH tunnel process: %d", s.Start.Process.Pid)
```
muchzill4: - Should we wrap `err` here as well?
- Perhaps it should stay "error when stopping SSH tunnel process"?

internal/command/exec.go
```diff
@@ -0,0 +1,45 @@
...
+	lw.mu.Lock()
+	defer lw.mu.Unlock()
+	return lw.w.Write(p)
+}
+
+func execute(cmd *exec.Cmd, run func() error, w io.Writer) error {
+	lw := &lockedWriter{w: w}
+	var stderrBuf bytes.Buffer
+	cmd.Stdout = lw
+	cmd.Stderr = io.MultiWriter(lw, &stderrBuf)
```
muchzill4: This bothers me a bit. The moment you hand off the `exec.Cmd` to `Run|StartCommand` it gets mutated. I'd almost rather have `command.Start|Run(name, args...)`, which mimics stdlib, but does our conventional error handling.

muchzill4: Alternatively, we could learn to live with repetition for now (until it becomes unbearable) and _just_ do:

```go
command.FormatError(args []string, stderr *bytes.Buffer, err error)
```

internal/ssh/ssh_tunnel.go
awphi: Generous as in 2 seconds is too much? My thinking was we don't know what's on that port and it might be slow to respond. Happy to reduce if you think I'm being too cautious or bump if not cautious enough

internal/command/exec.go
awphi: > Alternatively, we could learn to live with repetition for now (until it becomes unbearable) and just do:

This is a bit of a pain because we'd also need to gather stderr all over the place which would involve using the locked writer to prevent data loss with interleaved stdout/stderr :(

awphi: > This bothers me a bit. The moment you hand off the exec.Cmd to Run|StartCommand it gets mutated. I'd almost rather have command.Start|Run(name, args...), which mimics stdlib, but does our conventional error handling.

This is also a pain because we grab the pid from the `exec.Cmd` struct after `Start`ing it to eventually kill it. I suppose we could write a wrapper for `exec.Cmd` that exposes what we need?

internal/ssh/ssh_tunnel.go
awphi: >Should we wrap err here as well?

Yes good call

> Perhaps it should stay "error when stopping SSH tunnel process"?

We're not consistent on this throughout the codebase - we do "failed when X", "error when X", "when X", "X failed". I'll change it here but probably needs a consistency pass

muchzill4: 💯 consistency is king.

internal/command/exec.go
awphi: https://github.com/arm/topo/pull/199/commits/5579600e2894341b58270d5e9c76a0d70a8a8169

internal/ssh/ssh_tunnel.go
awphi: https://github.com/arm/topo/pull/199/commits/5579600e2894341b58270d5e9c76a0d70a8a8169

internal/command/exec.go
awphi: Addressed by removing stderr from the error message

yejseo01 (approved):
LGTM!

PR #200, opened 2026-04-14
fix: skip buildable services during docker compose pull

muchzill4:
## Changes

- Pass `--ignore-buildable` to `docker compose pull` so it skips services that have a `build:` context. Without this, pull tries to fetch locally-built image names from Docker Hub and fails with `pull access denied`.


awphi (approved):
Looks good! One question - if the `image` was indeed pullable and buildable which takes precedence with this flag? I'd assume the build but that seems wasteful albeit better than a failure

muchzill4:
> Looks good! One question - if the `image` was indeed pullable and buildable which takes precedence with this flag? I'd assume the build but that seems wasteful albeit better than a failure

It will always build in that case, which should leverage the docker layer cache.
The value of `image:` will be the name of the built image.

PR #202, opened 2026-04-14
feat: idempotent posix install script

awphi:
## Changes

Does as it says on the tin - code was a collaboration between me and codex but have tested quite thoroughly on my machine, refactored to my tastes and wrote an e2e test suite for it.

See the README on how to use. Would be keen to get review from a linux user. 

Equivalent windows `.ps1` to follow!


scripts/install.sh
```diff
@@ -0,0 +1,200 @@
...
+  while [ $# -gt 0 ]; do
+    case "$1" in
+      --version)
+        [ $# -ge 2 ] || { echo "Error: --version requires a value" >&2; exit 1; }
+        version="$2"; shift 2 ;;
+      --path)
+        [ $# -ge 2 ] || { echo "Error: --path requires a value" >&2; exit 1; }
+        install_dir="$2"; shift 2 ;;
+      -h|--help)
+        sed -n '2,/^$/s/^# \{0,1\}//p' "$0"; exit 0 ;;
```
Luke-Parkin: Could you add a comment here explaining that this reads the header comments of the script and strips the #

(For those of us who can't sight read sed)

scripts/install.sh
```diff
@@ -0,0 +1,200 @@
...
+  echo "${version} ${install_dir}"
+}
+
+resolve_version() {
+  version="$1"
+
+  if [ -z "$version" ]; then
+    echo "Resolving latest version..." >&2
+    page="$(fetch "${BASE_URL}/")"
+    version="$(echo "$page" | sed -n 's/.*\(v[0-9][0-9]*\.[0-9][0-9]*\.[0-9][0-9]*\).*/\1/p' | tail -n1)"
```
Luke-Parkin: Another comment saying this:
1. Regexes all the version numbers and returns a line delaminated list of ver nums

2. Tail fetches the latest (which is at the bottom) would be useful here

scripts/install.sh
```diff
@@ -0,0 +1,200 @@
...
+    exit 1
+  fi
+}
+
+install_binary() {
+  src="$1"
+  install_dir="$2"
+  version="$3"
+
+  install -m 0755 "$src" "${install_dir}/${BINARY_NAME}"
```
Luke-Parkin, line 170: This could error out if topo was previously installed in a directory where the script does not have permission to install.

Not sure what the usual bash script etiquette is, whether we should catch this or just let it throw and the user figures it out.

Luke-Parkin (commented):
Nice script!!

Just a few notes, mainly around commenting scary seds. If not for us, but also for anyone trying to audit the script before running it.

Won't approve as I'd prefer someone better at sh to do so! (But ping me if no one else comes along)

scripts/install.sh
awphi: Decided to kill this because it's scary/fragile + having a `USAGE` global near the top is basically a comment https://github.com/arm/topo/pull/202/commits/53f0bf6e10d28bdd6eda4d7f58f97b94cdb2f713

scripts/install.sh
awphi: good shout https://github.com/arm/topo/pull/202/commits/8b2e0d0ee7792ffc2bb17f2427a65f5ba7e60329

scripts/install.sh
awphi: Imo it's okay - if they've installed with the script previously it would be extremely weird for it to fail since we only check `bin/` dirs in their home dir. 

If they installed by some other means then surfacing the error message would be the most helpful thing we could do imo. Will leave open to hear other opinions though.

Luke-Parkin: Yep am fine with that!

scripts/install.sh
```diff
@@ -0,0 +1,202 @@
...
+
+resolve_version() {
+  version="$1"
+
+  if [ -z "$version" ]; then
+    echo "Resolving latest version..." >&2
+    page="$(fetch "${BASE_URL}/")"
+    # regex to match version strings like v1.2.3, taking the last one with tail to get the latest
+    # as artifactory lists versions in ascending order.
+    version="$(echo "$page" | sed -n 's/.*\(v[0-9][0-9]*\.[0-9][0-9]*\.[0-9][0-9]*\).*/\1/p' | tail -n1)"
```
muchzill4: We should use `sort -V` in case artifactory decided to order differently.

awphi: Sorry got distracted by other things and forgot about this suggestion - https://github.com/arm/topo/pull/202/commits/837d4413aefc344da66592337d4638bc2ca111fd

scripts/install.sh
```diff
@@ -0,0 +1,201 @@
...
+  for candidate in "$HOME/.local/bin" "$HOME/bin"; do
+    case ":${PATH}:" in
+      *":${candidate}:"*)
+        mkdir -p "$candidate" 2>/dev/null && echo "$candidate" && return
+        ;;
+    esac
+  done
+
+  echo "Error: could not find a user-writable directory on PATH." >&2
+  echo "Provide one explicitly with --path, or add ~/.local/bin to your PATH." >&2
```
Luke-Parkin, line 143: We need to tell people, literally give them a command, to tell them how to add this to their $PATH

Luke-Parkin (commented):
One more thing then I think it's ready

Luke-Parkin (approved):
LGTM - please check my readme changes before merge though.

I tested the windows stuff on x86 win11

PR #204, opened 2026-04-15
chore: drop TestNewKeySetup

muchzill4:
## Changes

- Remove `TestNewKeySetup` — it only asserts the return types of a trivial constructor with no branching logic, no error paths, and a vestigial `error` return that is always `nil`. All three sub-cases exercise the same code path.


awphi (approved):
🚀

PR #216, opened 2026-04-16
feat!: remove `topo describe` command

federicobozzini:
## Changes

- `topo describe` removed from public topo API
- update `topo describe` to print to stdout
- added real support for json output


cmd/topo/templates.go
muchzill4: We can get rid of this, right?

muchzill4: This can be removed.

internal/output/templates/describe_test.go
```diff
@@ -0,0 +1,64 @@
...
+			}
+			toPrint := templates.PrintableTargetDescription{HardwareProfile: profile}
+			var out bytes.Buffer
+
+			err := printable.Print(toPrint, &out, term.Plain)
+
+			require.NoError(t, err)
+			var parsed target.HardwareProfile
+			require.NoError(t, yaml.Unmarshal(out.Bytes(), &parsed))
+			assert.Equal(t, profile, parsed)
```
muchzill4: Better assert using [`YAMLEq`](https://github.com/stretchr/testify/blob/v1.11.1/assert/assertions.go#L1878), so we can check the actual contents.

internal/output/templates/describe_test.go
```diff
@@ -0,0 +1,64 @@
...
+			}
+			var out bytes.Buffer
+
+			err := printable.Print(toPrint, &out, term.JSON)
+
+			require.NoError(t, err)
+			assert.Contains(t, out.String(), `"model": "Cortex-A55"`)
+			assert.Contains(t, out.String(), `"cores": 4`)
+			assert.Contains(t, out.String(), `"remoteproc0"`)
+			assert.Contains(t, out.String(), `"totalmemory_kb": 16384`)
```
muchzill4: Use `assert.JSONEq`.

cmd/topo/templates.go
federicobozzini: https://github.com/arm/topo/pull/216/commits/31b8a489f71036ffc022238ba413829e93b5e15f

federicobozzini: https://github.com/arm/topo/pull/216/commits/31b8a489f71036ffc022238ba413829e93b5e15f

internal/output/templates/describe_test.go
federicobozzini: https://github.com/arm/topo/pull/216/commits/4da821ce7e7deeecc98cdc049efc322f021dfd33

internal/output/templates/describe_test.go
federicobozzini: https://github.com/arm/topo/pull/216/commits/504365c9309bd7379fa6bf89248fef91b9d796e2

internal/output/templates/describe.go
```diff
@@ -0,0 +1,30 @@
...
+	target.HardwareProfile
+}
+
+func (d PrintableTargetDescription) AsPlain(_ bool) (string, error) {
+	var buf bytes.Buffer
+	if err := yaml.NewEncoder(&buf).Encode(d.HardwareProfile); err != nil {
+		return "", fmt.Errorf("encode target description as yaml: %w", err)
+	}
+	return buf.String(), nil
+}
```
awphi, line 22: Not sure what the outcome of our democratic vote was but outputting yaml by default seems weird to me. I guess the intention is to pipe it into a file down the line but I don't think we should support yaml until then.

Something prettier or not supporting plain at all given it's in the vscnode namespace would be my preference

federicobozzini: Can we keep this returning yaml in this PR and change it to something different later?

awphi: Sounds reasonable to me, given this is now hidden we can change it without making further breaking changes I suppose!

muchzill4 (approved):

PR #212, opened 2026-04-16
refactor: replace probe structs with plain functions

muchzill4:
## Changes

- Remove `HardwareProbe` and `SSHAuthenticationProbe` structs and their constructors
- Replace `NewProbeXXX().Probe()` pattern with plain `ProbeXXX()` functions
- Update all call sites in `describe`, `templates`, `health`

PR #217, opened 2026-04-16
refactor: split hardware probe into separate files

muchzill4:
## Changes

- Split `hardware_probe.go` into `cpu_probe.go`, `memory_probe.go`, `remoteproc_probe.go`
- Move tests to matching `*_test.go` files
- Add `TestProbeRemoteproc` covering happy path, empty dir, missing dir, and read error
- Move `TestProbeHardware` cases into `TestProbeCPU` (they only tested CPU probing)

## End goal

- `probe.SSHAuthentication` is standalone - as is now

### Progress

- [x] Split hardware probe into sub-probes (**this PR**)
- [ ] Rename `target.Xxx` to `probe.Xxx`
- [ ] Reduce the public surface of the `probe` package (stop testing internals)


internal/target/memory_probe.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+
+	value, err := FindKeyValueInString(key, out)
+	if err != nil {
+		return 0, fmt.Errorf("in checking %s", path)
+	}
+	return value, nil
+}
+
+func FindKeyValueInString(key string, text string) (int64, error) {
```
th3james, line 30: nit: Questionable whether this makes sense as part of the public package interface. Feels like it's only public to simplify testing.

th3james (approved):

internal/target/memory_probe.go
muchzill4: I agree. Planning to do something about it in step 4: `Reduce the public surface of the probe package (stop testing internals)`.

PR #218, opened 2026-04-16
feat: windows install script

awphi:
## Changes

- Adds a powershell windows installation script. This mostly mirrors the behaviour of the posix one with two notable exceptions;
  - Windows convention found in tools like vscode, python or the GH CLI is to install each binary in its own directory. There's no `~/.local/bin` idiom.
  - As a result, instead of requiring every user to add this new topo-specific binary dir to their `PATH`, we do it for them (unless they provide a `-Path` arg, then we don't touch it)
- Updates the README accordingly, moving the manual installation instructions below both automatic scripts so it's less prevalent and avoiding duplicating it under both OS-specific headings. 
- Adds an e2e test suite mirroring the posix one, updates to use go build tags instead of guards so it's _blazingly fast_

Tested on Windows 11 x86 and Arm with/without args and all looked good to me but I'm not going to pretend to be a powershell expert.

### DIFF
--- a/e2e/install_script_windows_test.go
+++ b/e2e/install_script_windows_test.go
@@ -1,0 +1,95 @@
+//go:build windows
+
+package e2e
+
+import (
+	"os"
+	"os/exec"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func runInstallScript(t *testing.T, args ...string) (string, error) {
+	t.Helper()
+
+	path, err := filepath.Abs("../scripts/install.ps1")
+	require.NoError(t, err)
+
+	tmpDir := t.TempDir()
+	homeDir := filepath.Join(tmpDir, "home")
+	localAppDataDir := filepath.Join(tmpDir, "localappdata")
+	appDataDir := filepath.Join(tmpDir, "appdata")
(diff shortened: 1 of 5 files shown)


e2e/install_script_windows_test.go
```diff
@@ -0,0 +1,95 @@
...
+		"-File", path,
+	}, args...)
+	cmd := exec.Command("powershell", cmdArgs...)
+	cmd.Env = append(os.Environ(),
+		"HOME="+homeDir,
+		"USERPROFILE="+homeDir,
+		"LOCALAPPDATA="+localAppDataDir,
+		"APPDATA="+appDataDir,
+		"TEMP="+tempDir,
+		"TMP="+tempDir,
```
muchzill4: Can you explain why we need to set all these variables?

I get that we're trying to isolate ourselves, but this seems excessive. Perhaps it's a Windows thing I don't comprehend. 😉

muchzill4 (approved):
Other than the `env` question LGTM. Nice work on simplifying the installation for our users. 👏

e2e/install_script_windows_test.go
awphi: Yep good call, was being overprotective but we can get away with lot less noise - https://github.com/arm/topo/pull/218/commits/e304982076b2fbc17ffa07f52640447ec810c3a5

PR #219, opened 2026-04-16
feat: Improve clone's logging output

Luke-Parkin:
## Changes

- The UX of topo clone was inconsistent with the rest of the cli's UX.
- This PR Utilises the standard operations sequence to make it look a little nicer.
- It also adds a print dir path step, following responses from UX reviews where users failed to understand where their project had been cloned to.

### DIFF
--- a/internal/arguments/interactive_provider.go
+++ b/internal/arguments/interactive_provider.go
@@ -23,7 +23,7 @@ func (p *InteractiveProvider) Provide(args []Arg) ([]ResolvedArg, error) {
 
 	for i, arg := range args {
 		if i != 0 {
-			fmt.Printf("\n")
+			fmt.Fprintf(p.output, "\n")
 		}
 		_, err := fmt.Fprintf(p.output, "Provide: %s\n", arg.Description)
 		if err != nil {
--- a/internal/arguments/interactive_provider.go
+++ b/internal/arguments/interactive_provider.go
@@ -21,8 +21,11 @@ func (p *InteractiveProvider) Provide(args []Arg) ([]ResolvedArg, error) {
 	var result []ResolvedArg
 	scanner := bufio.NewScanner(p.input)
 
-	for _, arg := range args {
-		_, err := fmt.Fprintf(p.output, "\n%s\n", arg.Description)
+	for i, arg := range args {
+		if i != 0 {
+			fmt.Printf("\n")
+		}
+		_, err := fmt.Fprintf(p.output, "Provide: %s\n", arg.Description)
 		if err != nil {
 			return nil, err
 		}
--- a/internal/arguments/interactive_provider.go
+++ b/internal/arguments/interactive_provider.go
@@ -23,7 +23,10 @@ func (p *InteractiveProvider) Provide(args []Arg) ([]ResolvedArg, error) {
 
 	for i, arg := range args {
 		if i != 0 {
-			fmt.Fprintf(p.output, "\n")
+			_, err := fmt.Fprintf(p.output, "\n")
+			if err != nil {
+				return nil, err
+			}
 		}
 		_, err := fmt.Fprintf(p.output, "Provide: %s\n", arg.Description)
 		if err != nil {
--- a/internal/project/project_test.go
+++ b/internal/project/project_test.go
@@ -56,6 +56,48 @@ func TestInit(t *testing.T) {
 	})
 }
 
+func TestClone(t *testing.T) {
+	t.Run("clones source into destination directory", func(t *testing.T) {
+		dir := t.TempDir()
+		destDir := filepath.Join(dir, "demo")
+		mockSource := mockTemplateSourceWithContent(t, `
+services:
+  app:
+    image: nginx:alpine
+`, "demo-source")
+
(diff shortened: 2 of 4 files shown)


internal/project/project_test.go
```diff
@@ -80,6 +122,19 @@ func mockTemplateSourceWithErrorOnCopy(t *testing.T, errToReturn error, sourceNa
 	return mockSource
 }
 
+func mockCloneSourceWithContent(t *testing.T, content, sourceName string) *mockTemplateSource {
```
muchzill4: Imho we should rename `mockTemplateSourceWithContent` to `mockSourceWithContent` and use it instead of this, which is essentially the same.

internal/project/project_test.go
```diff
@@ -56,6 +56,48 @@ func TestInit(t *testing.T) {
...
+  app:
+    image: nginx:alpine
+`, "demo-source")
+
+		err := project.Clone(destDir, mockSource, arguments.NewStrictProviderChain())
+
+		require.NoError(t, err)
+		composeFilePath := filepath.Join(destDir, template.ComposeFilename)
+		_, statErr := os.Stat(composeFilePath)
+		require.NoError(t, statErr)
```
muchzill4: There's `assert.FileExists`.

internal/arguments/interactive_provider.go
```diff
@@ -21,8 +21,14 @@ func (p *InteractiveProvider) Provide(args []Arg) ([]ResolvedArg, error) {
...
-	for _, arg := range args {
-		_, err := fmt.Fprintf(p.output, "\n%s\n", arg.Description)
+	for i, arg := range args {
+		if i != 0 {
+			_, err := fmt.Fprintf(p.output, "\n")
+			if err != nil {
+				return nil, err
+			}
+		}
+		_, err := fmt.Fprintf(p.output, "Provide: %s\n", arg.Description)
```
muchzill4, line 31: Can you show a screenshot of before / after?

muchzill4:
PR title of a `feat:` ends up in the changelog, is it meaningful to the reader? Say what we have actually improved externally, not what code internals we've used.

internal/project/project_test.go
Luke-Parkin: https://github.com/arm/topo/pull/219/commits/bdb15f82b87d11a7073bb880ab66971dca092a0a

internal/project/project_test.go
Luke-Parkin: https://github.com/arm/topo/pull/219/commits/bdb15f82b87d11a7073bb880ab66971dca092a0a

internal/arguments/interactive_provider.go
Luke-Parkin: image
image

internal/project/project_test.go
```diff
@@ -56,14 +56,55 @@ func TestInit(t *testing.T) {
...
+
+func mockSourceWithContent(t *testing.T, content, sourceName string) *mockTemplateSource {
 	mockSource := &mockTemplateSource{}
 	mockSource.On("CopyTo", mock.Anything).Return(nil).Run(func(args mock.Arguments) {
 		destDir := args.String(0)
 		testutil.RequireMkdirAll(t, destDir)
 		testutil.RequireWriteFile(t, filepath.Join(destDir, template.ComposeFilename), content)
 	})
-	mockSource.On("GetName").Return(sourceName, nil)
+	mockSource.On("GetName").Maybe().Return(sourceName, nil)
```
Luke-Parkin, line 128: Let me know what you think about this... I am trying to figure out why anyone would do a .Maybe() since the tests won't ever fail on it?

muchzill4: I'm on a mission to obliterate mocks from the codebase and use fakes instead...

I think this is fine for now.

internal/arguments/interactive_provider.go
muchzill4: Thanks @Luke-Parkin , makes this much easier to review. 🙏 

---

Is `"Git clone"` really what this PR introduces, I think it says [`"Copy files"`](https://github.com/arm/topo/pull/219/changes#diff-cf4f5504efe89b688e171549faeccdeaa0723a75e9883aa22ef4fee73cee2b64R266) in the code. Former is obviously inaccurate for `dir:` source, latter is somewhat too generic. What if we did:

```go
fmt.Sprintf("Cloning source %s", src)
```

Where "cloning" is because we're doing "topo clone", and you'd have:

- `Cloning source git:https://...`
- `Cloning source dir:/some/path`

---

I don't like "Project directory" as a heading. I can't tell why, but it doesn't align well with other headings.
How about:

```
┌─  Next steps
cd <project_directory>
topo deploy
```

Luke-Parkin: Like this?

Image

I really like the next steps - Do you think this correctly conveys that <they> need to do this? Or whether users will think that this is what topo has done.

muchzill4: Hmm. 

- `🫵 Next steps`
- `All done`
 
And the text could say (note indent to make it visually different from previous logs):

```
Now run:
   cd topo-welcome
   topo deploy
```

muchzill4 (approved):

PR #221, opened 2026-04-17
refactor: rename target probes to probe package

muchzill4:
## Changes

- Move probe files from `internal/target` to `internal/probe` package
- Rename probe functions to fit package namespace (`probe.CPU` instead of `probe.ProbeCPU`)
- Replace `SSHAuthenticationProbeOptions` struct with a `bool` parameter
- Update all call sites across `cmd/topo`, `internal/catalog`, `internal/describe`, `internal/health`

### End goal

- `probe.SSHAuthentication` is standalone

### Progress

- [x] [Split hardware probe into separate files](https://github.com/arm/topo/pull/217)
- [x] Rename `target.Xxx` to `probe.Xxx` (**this PR**)
- [ ] Reduce the public surface of the `probe` package (stop testing internals)


internal/probe/memory_probe_test.go
```diff
@@ -22,7 +22,7 @@ MemFree:        8192000 kB`
 	t.Run("returns error when key not found", func(t *testing.T) {
 		text := `MemTotal:       16384000 kB`
 
-		got, err := target.FindKeyValueInString("MissingKey", text)
+		got, err := probe.FindKeyValueInString("MissingKey", text)
```
awphi, line 25: This doesn't feel like the right package for this function but neither did `target` 🤷 

Tbh I'm not sure how necessary it is to export - we could replace these tests with tests on `probe.Memory` that make use of the fake runner.

Food for though - doesn't need to happen in this PR if at all

awphi (approved):
Nice, much better naming

internal/probe/memory_probe_test.go
muchzill4: Haha. This is the same thing @th3james said on the previous PR. My answer is: this will be handled in step 4: `Reduce the public surface of the probe package (stop testing internals)`.

PR #222, opened 2026-04-17
refactor: always wrap ssh Runner commands in login shell, use runner to check deps

awphi:
## Changes

Two-pronged refactor:

- Removes `command.WrapInLoginShell` from the call site of every command invocation we run on the target and hoists into the ssh runner itself, lowering mental burden remembering if you have to provide it or not. Lmk if I missed a scenario where we would want to execute a command on the target without login shell.
- With this change, we can collapse the two `CommandSuccessfulFn` and `BinaryExistsFn`s we pass to the dependency checker into just passing the runner directly. 
   - This also allows us to clean up the related test suite a fair bit by using `runner.Fake` :)

This refactor is in service of making health checks a bit more flexible - allowing us to write ones that assert things about stdout of given commands which will be required for checks like an "is this version of topo the latest?".

### DIFF
--- a/internal/command/command_test.go
+++ b/internal/command/command_test.go
@@ -5,7 +5,6 @@ import (
 
 	"github.com/arm/topo/internal/command"
 	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
 )
 
 func TestWrapInLoginShell(t *testing.T) {
@@ -18,13 +17,6 @@ func TestWrapInLoginShell(t *testing.T) {
 }
 
 func TestBinaryLookupCommand(t *testing.T) {
-	t.Run("returns wrapped command for valid binary", func(t *testing.T) {
-		got, err := command.BinaryLookupCommand("docker")
-
-		require.NoError(t, err)
-		assert.Equal(t, command.WrapInLoginShell("command -v docker"), got)
-	})
-
 	t.Run("returns error for invalid binary", func(t *testing.T) {
 		got, err := command.BinaryLookupCommand("bad name")
 
--- a/internal/command/command_test.go
+++ b/internal/command/command_test.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/arm/topo/internal/command"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestWrapInLoginShell(t *testing.T) {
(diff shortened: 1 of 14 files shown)


internal/command/command_test.go
```diff
@@ -18,13 +17,6 @@ func TestWrapInLoginShell(t *testing.T) {
 }
 
 func TestBinaryLookupCommand(t *testing.T) {
-	t.Run("returns wrapped command for valid binary", func(t *testing.T) {
```
muchzill4, line 21: We still need a test here for the non-error case.

internal/health/dependencies_test.go
```diff
@@ -56,21 +57,18 @@ func TestPerformChecks(t *testing.T) {
...
+		runner := &runner.Fake{}
 
-		got := health.PerformChecks(context.Background(), deps, mockBinaryExists, mockCommandSuccessful)
+		got := health.PerformChecks(context.Background(), deps, runner)
 
 		want := []health.DependencyStatus{
 			{
 				Dependency: health.Dependency{Binary: "foo", Label: "bar", Checks: []health.Check{health.BinaryExists()}},
-				Error:      mockBinaryExists(context.Background(), "foo"),
+				Error:      errors.New("\"foo\" not found in $PATH"),
```
muchzill4: This couples fake runner's internals into the error we're expecting. I wonder if we shouldn't do something like:

```go
wantFoo := health.DependencyStatus{Dependency: fooDependency}
wantFoo.Error, _ = runner.BinaryExists(fooDependency.Binary)
wantBar ...
want := []health.DependencyStatus{wantFoo, wantBar}
```

muchzill4 (commented):
While I liked the previous approach being explicit, I think this is a good compromise and cleans things up nicely. Nice work. 👍

internal/command/command_test.go
awphi: https://github.com/arm/topo/pull/222/commits/47bd1b7b1afedf7f7ce1986aba11b5980c4b8925

internal/health/dependencies_test.go
awphi: good idea - https://github.com/arm/topo/pull/222/commits/e4a4676cf569c64200e67a0003e4e156220f6cee

muchzill4: Oh, that's even better - for some reason I assumed `BinaryExists` was returning two values.

muchzill4 (approved):

PR #223, opened 2026-04-17
refactor: reduce probe package's public API surface

muchzill4:
## Changes

- Unexport fields that are internal implementation details of how probing works
    - Rewrite CPU and Memory tests to go through public `CPU()`/`Memory()` entry points using runner fakes
- Drop redundant `_probe` suffix from all probe filenames
- Remove unused `Children` field from lscpu output type

### End goal

- `probe.SSHAuthentication` is standalone

### Progress

- [x] [Split hardware probe into separate files](https://github.com/arm/topo/pull/217)
- [x] [Rename `target.Xxx` to `probe.Xxx`](https://github.com/arm/topo/pull/221)
- [x] Reduce the public surface of the `probe` package (**this PR**)


internal/probe/cpu.go
```diff
@@ -26,48 +48,49 @@ func CPU(ctx context.Context, r runner.Runner) ([]HostProcessor, error) {
...
-	Lscpu []LscpuOutputField `json:"lscpu"`
+type lscpuJSON struct {
+	Lscpu []lscpuOutputField `json:"lscpu"`
 }
 
-func (proc *HostProcessor) ExtractArmFeatures() []string {
-	if len(proc.Features) == 0 {
-		return nil
-	}
+type fieldsByModelName map[string][]lscpuOutputField
```
th3james: maps are unordered, which means the sorting here is non-deterministic. This causes test flakiness:
```
go test ./internal/probe -run 'TestProbeCPU/parses multiple processors' -count=50
```

th3james (changes requested):

internal/probe/cpu.go
muchzill4: Nice catch! Thanks. Fixed @ https://github.com/arm/topo/pull/223/commits/9dd41b08cc28e6c0acc1f2bd3ff6b43f7d133fc0.

internal/probe/cpu.go
```diff
@@ -26,48 +48,53 @@ func CPU(ctx context.Context, r runner.Runner) ([]HostProcessor, error) {
...
+	var groups []modelFields
+	idx := -1
+	for _, f := range fields {
+		if f.Field == "Model name:" {
+			groups = append(groups, modelFields{name: f.Data})
+			idx = len(groups) - 1
+			continue
+		}
+		if idx >= 0 {
+			groups[idx].fields = append(groups[idx].fields, f)
```
th3james: nitpick: IMHO slightly clearer without idx
```suggestion
	for _, f := range fields {
		if f.Field == "Model name:" {
			groups = append(groups, modelFields{name: f.Data})
			continue
		}
		if len(groups) > 0 {
			groups[len(groups)-1].fields = append(groups[len(groups)-1].fields, f)
```

th3james (approved):
Very nice

PR #224, opened 2026-04-17
refactor: make dependency checks structs

awphi:
## Changes

- Removes the `CheckKind` enum + corresponding switch and replaces it with a `Check` interface that takes context, runner and the current dependency, performs whatever is needed and returns a fix string and an error as needed.
  - This is more extensible than the previous approach if we, for example, want to assert things about the stdout of a command - for example topo latest version checking or ensuring that `docker info` contains what we expect, even with a zero exit code
- Moves the warning-wrapping logic based on severity out of `PerformChecks` and into the check itself. Currently only used in `BinaryExists` but this seemed to fit better to me.
  - Deletes the old test for this and creates a new one that tests `BinaryExists` in isolation
- Decouples prerequisite installation checking from the health checks themselves - now installation checking is performed after all the checks have passed. i.e. if docker is installed but `docker info` does not pass, we don't check other deps that rely on docker
  - Not sure if this behaviour is quite right - it sounds right to me when you interpret a check as validating the software is installed and ready-to-use i.e. you can't use remoteproc-runtime even if it's installed if docker isn't working.

All in all should read a little bit cleaner and adds a lot more flexibility in what our checks can do - this will be utilised in a follow-up PR for topo version checking!

### DIFF
--- a/internal/health/check.go
+++ b/internal/health/check.go
@@ -7,7 +7,7 @@ import (
 )
 
 type Check interface {
-	Apply(ctx context.Context, r runner.Runner, dep Dependency) (string, error)
+	Run(ctx context.Context, r runner.Runner, dep Dependency) (string, error)
 }
 
 type CheckSeverity int
@@ -22,7 +22,7 @@ type CommandSuccessful struct {
 	Fix string
 }
 
-func (c CommandSuccessful) Apply(ctx context.Context, r runner.Runner, dep Dependency) (string, error) {
+func (c CommandSuccessful) Run(ctx context.Context, r runner.Runner, dep Dependency) (string, error) {
 	_, err := r.Run(ctx, c.Cmd)
 	return c.Fix, err
 }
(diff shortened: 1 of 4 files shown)


internal/health/check.go
```diff
@@ -0,0 +1,41 @@
...
+	SeverityError CheckSeverity = iota
+	SeverityWarning
+)
+
+type CommandSuccessful struct {
+	Cmd string
+	Fix string
+}
+
+func (c CommandSuccessful) Apply(ctx context.Context, r runner.Runner, dep Dependency) (string, error) {
```
muchzill4: `s/Apply/Perform`? or `Run`?

internal/health/dependencies.go
```diff
@@ -161,27 +134,22 @@ func PerformChecks(ctx context.Context, dependencies []Dependency, runner runner
...
-				if check.Severity == SeverityWarning {
-					err = WarningError{Err: err}
-				}
-				fix = check.Fix
 				break
 			}
 		}
 
+		if err == nil && dep.SoftwareEnumID != UnsetSoftwareDependency {
+			exists := runner.BinaryExists(ctx, dep.Binary)
```
muchzill4: Aren't we now doing double `BinaryExists` checks?

awphi: Yes good point - my think was that this was better than coupling installation checks to the presence/result of the `BinaryExists` check but maybe we can just special case it to avoid hitting the target twice https://github.com/arm/topo/pull/224/commits/330db841e82850cd14b966670183950b5053e899

internal/health/check.go
awphi: mm maybe `Run` - I had that originally but went back on it https://github.com/arm/topo/pull/224/commits/9c2f1b80c8798458be55708274002982fe2d47a0

muchzill4 (approved):

PR #226, opened 2026-04-17
fix: handling of SSH timeout error

federicobozzini:
## Changes

- `SSH.BinaryExists` was masking timeout errors as "not found", causing binaries to be falsely reported as missing under high network latency. This also silently skipped any dependencies that depended on the timed-out check (e.g. remoteproc checks gated on Docker).
Timeout errors are now propagated so health reports distinguish "timed out" from "not installed", and timed-out prerequisites no longer cause their dependents to be silently omitted.

Currently:

```sh
$ ./topo health --target imx93 --timeout 1s
Host
----
SSH: ✅ (ssh)
Container Engine: ✅ (docker)

Target
------
Connectivity: ✅
Container Engine: ❌ ("docker" not found on remote target's $PATH)
Hardware Info: ❌ ("lscpu" not found on remote target's $PATH)
```

After this change

```sh
$ ./topo health --target imx93 --timeout 1s
Host
----
SSH: ✅ (ssh)
Container Engine: ✅ (docker)

Target
------
Connectivity: ✅
Container Engine: ❌ (timed out)
Hardware Info: ❌ (timed out)
Subsystem Driver (remoteproc): ℹ️ (no remoteproc devices found)
```

### DIFF
--- a/internal/health/check.go
+++ b/internal/health/check.go
@@ -35,8 +35,14 @@ type BinaryExists struct {
 
 func (b BinaryExists) Run(ctx context.Context, r runner.Runner, dep Dependency) (string, error) {
 	err := r.BinaryExists(ctx, dep.Binary)
-	if b.Severity == SeverityWarning && err != nil && !errors.Is(err, runner.ErrTimeout) {
-		err = WarningError{Err: err}
+	if err == nil {
+		return b.Fix, nil
+	}
+	if errors.Is(err, runner.ErrTimeout) {
+		return "", err
+	}
+	if b.Severity == SeverityWarning {
+		return b.Fix, WarningError{Err: err}
 	}
 	return b.Fix, err
 }
--- a/internal/health/check.go
+++ b/internal/health/check.go
@@ -34,15 +34,14 @@ type BinaryExists struct {
 }
 
 func (b BinaryExists) Run(ctx context.Context, r runner.Runner, dep Dependency) (string, error) {
-	err := r.BinaryExists(ctx, dep.Binary)
-	if err == nil {
-		return b.Fix, nil
+	if err := r.BinaryExists(ctx, dep.Binary); err != nil {
+		if errors.Is(err, runner.ErrTimeout) {
+			return "", err
+		}
+		if b.Severity == SeverityWarning {
+			err = WarningError{Err: err}
+		}
+		return b.Fix, err
 	}
-	if errors.Is(err, runner.ErrTimeout) {
-		return "", err
-	}
-	if b.Severity == SeverityWarning {
-		return b.Fix, WarningError{Err: err}
-	}
-	return b.Fix, err
+	return "", nil
 }
--- a/internal/health/check.go
+++ b/internal/health/check.go
@@ -1,0 +1,42 @@
+package health
+
+import (
+	"context"
+	"errors"
+
+	"github.com/arm/topo/internal/runner"
+)
+
+type Check interface {
+	Run(ctx context.Context, r runner.Runner, dep Dependency) (string, error)
+}
+
+type CheckSeverity int
+
+const (
+	SeverityError CheckSeverity = iota
+	SeverityWarning
+)
+
+type CommandSuccessful struct {
+	Cmd string
+	Fix string
+}
+
+func (c CommandSuccessful) Run(ctx context.Context, r runner.Runner, dep Dependency) (string, error) {
+	_, err := r.Run(ctx, c.Cmd)
+	return c.Fix, err
+}
+
+type BinaryExists struct {
+	Severity CheckSeverity
+	Fix      string
+}
+
+func (b BinaryExists) Run(ctx context.Context, r runner.Runner, dep Dependency) (string, error) {
+	err := r.BinaryExists(ctx, dep.Binary)
+	if b.Severity == SeverityWarning && err != nil && !errors.Is(err, runner.ErrTimeout) {
+		err = WarningError{Err: err}
+	}
+	return b.Fix, err
+}
--- a/internal/health/dependencies.go
+++ b/internal/health/dependencies.go
@@ -2,6 +2,7 @@ package health
 
 import (
 	"context"
+	"errors"
 
 	"github.com/arm/topo/internal/runner"
 )
@@ -152,7 +153,7 @@ func hardwareCapabilityMatches(required []HardwareCapability, available map[Hard
 	return false
 }
 
-func PerformChecks(ctx context.Context, dependencies []Dependency, runner runner.Runner) []DependencyStatus {
+func PerformChecks(ctx context.Context, dependencies []Dependency, r runner.Runner) []DependencyStatus {
 	installed := make(map[SoftwareDependency]struct{})
 	result := make([]DependencyStatus, 0, len(dependencies))
 
@@ -166,15 +167,21 @@ func PerformChecks(ctx context.Context, dependencies []Dependency, runner runner
 		for _, check := range dep.Checks {
 			switch check.Kind {
 			case CheckBinaryExists:
-				err = runner.BinaryExists(ctx, dep.Binary)
+				err = r.BinaryExists(ctx, dep.Binary)
+				if errors.Is(err, runner.ErrTimeout) {
+					if dep.SoftwareEnumID != UnsetSoftwareDependency {
+						installed[dep.SoftwareEnumID] = struct{}{}
+					}
+					break
+				}
 				if err == nil && dep.SoftwareEnumID != UnsetSoftwareDependency {
 					installed[dep.SoftwareEnumID] = struct{}{}
 				}
 			case CheckCommandSuccessful:
-				_, err = runner.Run(ctx, check.Arg)
+				_, err = r.Run(ctx, check.Arg)
 			}
 			if err != nil {
-				if check.Severity == SeverityWarning {
+				if check.Severity == SeverityWarning && !errors.Is(err, runner.ErrTimeout) {
 					err = WarningError{Err: err}
 				}
 				fix = check.Fix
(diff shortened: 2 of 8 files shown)


internal/health/dependencies.go
```diff
@@ -137,9 +136,17 @@ func PerformChecks(ctx context.Context, dependencies []Dependency, runner runner
 		var fix string
 		var err error
 		for _, check := range dep.Checks {
-			fix, err = check.Run(ctx, runner, dep)
-			if _, ok := check.(BinaryExists); ok && err == nil {
-				installed[dep.SoftwareEnumID] = struct{}{}
+			fix, err = check.Run(ctx, r, dep)
+			if _, ok := check.(BinaryExists); ok {
+				if errors.Is(err, runner.ErrTimeout) {
```
muchzill4: What are we trying to achieve here? To me this sounds like "pretending" binary exists if we time out which does sound like potentially surprising behaviour.

1. Does docker exist? We timed out, so it does.
2. Does docker exist? We've checked, it does not. <- at this point I'm confused about Schrodinger's docker

internal/health/check.go
```diff
@@ -34,7 +35,7 @@ type BinaryExists struct {
 
 func (b BinaryExists) Run(ctx context.Context, r runner.Runner, dep Dependency) (string, error) {
 	err := r.BinaryExists(ctx, dep.Binary)
-	if b.Severity == SeverityWarning && err != nil {
+	if b.Severity == SeverityWarning && err != nil && !errors.Is(err, runner.ErrTimeout) {
```
muchzill4: It's a bit odd to me that in case of a timeout, we drop the severity, but retain the fix, which is likely to not be accurate.

Your output shows the issue quite nicely:
```
Remoteproc Shim: ❌ (timed out)
  → run `topo install remoteproc-runtime`
```
If it times out, `topo install` isn't going to fix anything.

- I think this should still be a warning (`X` indicates something crucial to proceed, having working shim is not, no matter the reason)
- It should not say "run `topo install`". We could mention `--timeout` though.

~Also, should we not match `runner.ErrSSH` to cast a wider net?~

internal/runner/ssh.go
```diff
@@ -63,6 +63,9 @@ func (r *SSH) BinaryExists(ctx context.Context, bin string) error {
 		return err
 	}
 	if _, err := r.Run(ctx, cmd); err != nil {
+		if errors.Is(err, ErrTimeout) {
```
muchzill4: Why add this?

muchzill4 (changes requested):

internal/runner/ssh.go
federicobozzini: I'm not sure what are you asking here. I don't see another nice way to make the  timeout error bubble up from here. It's like the whole point of the PR :thinking:

muchzill4: Ah! I'm still in the mindset of `ErrTimeout == ErrSSH`, but that's no longer true. 😓 Ignore me.

muchzill4: Should we do `if errors.Is(err, ssh.ErrSSH) || errors.Is(err, ErrTimeout)`? Both have the same codepath (returning err)

internal/health/dependencies.go
muchzill4: Actually it'd be Schrodingers `remoteproc-runtime` that depends on existence of docker. Anyway, I think we shouldn't mark dependency as installed if we can't determine if it really is.

federicobozzini: Just to make sure I understand what you mean, do you think we should still check (and show) the existance of `remoteproc-runtime` even if we can't check if docker is installed because of a timeout?

internal/runner/ssh.go
federicobozzini: https://github.com/arm/topo/pull/226/commits/d3f4910678d3d42b9977b4ee43cc9db6464549e7

internal/health/check.go
federicobozzini: https://github.com/arm/topo/pull/226/commits/215fa5978c09343fddd7efc9230e014cb2cf6707

internal/health/dependencies.go
muchzill4: No, we should not. If we can't determine if `docker` is installed, we should skip all the binaries that depend on `docker`. In other words, treat docker-check-timeout as if it was docker-is-not-installed.

internal/health/check.go
```diff
@@ -34,8 +35,14 @@ type BinaryExists struct {
...
+	if err == nil {
+		return b.Fix, nil
+	}
+	if errors.Is(err, runner.ErrTimeout) {
+		return "", err
+	}
+	if b.Severity == SeverityWarning {
+		return b.Fix, WarningError{Err: err}
 	}
 	return b.Fix, err
```
muchzill4: Now this has grown a bit, I find it a bit harder to reason about when the error actually happens and why we're always returning the `b.Fix`. How about:

```go
if err := r.BinaryExists(cox, dep.Binary); err != nil {
    if errors.Is(err, runner.ErrTimeout) {
        return "", err
    }
    if b.Severity == SeverityWarning {
        err = WarningError{Err: err}
    }
    return b.Fix, err
}
return "", nil
```

I think it's a bit clearer about the "it errored" path, making it also obvious what happens in case check returns no error.

internal/health/dependencies.go
```diff
@@ -152,6 +160,10 @@ func PerformChecks(ctx context.Context, dependencies []Dependency, runner runner
 			Error:      err,
 			Fix:        fix,
 		})
+
+		if errors.Is(err, runner.ErrTimeout) {
```
muchzill4: I'm not sure we should be assuming that one `ErrTimeout` means everything will time out. While it is true, this layer doesn't necessarily need to know it.

internal/health/dependencies_test.go
```diff
@@ -183,6 +183,67 @@ func TestPerformChecks(t *testing.T) {
 		}
 		assert.Equal(t, want, got)
 	})
+
+	t.Run("propagates timeout error from BinaryExists without masking as not found", func(t *testing.T) {
```
muchzill4: Do we actually need this test? Not found-ness and timeout are concepts of `runner.SSH`.

internal/health/check.go
```diff
@@ -33,9 +34,14 @@ type BinaryExists struct {
...
-	err := r.BinaryExists(ctx, dep.Binary)
-	if b.Severity == SeverityWarning && err != nil {
-		err = WarningError{Err: err}
+	if err := r.BinaryExists(ctx, dep.Binary); err != nil {
+		if errors.Is(err, runner.ErrTimeout) {
+			return "", err
+		}
+		if b.Severity == SeverityWarning {
+			err = WarningError{Err: err}
+		}
```
muchzill4, line 43: I missed that in my snippet, but I still wonder if we shouldn't retain the severity for a dependency, no matter why it fails.

internal/health/dependencies.go
federicobozzini: > I'm not sure we should be assuming that one ErrTimeout means everything will time out. While it is true, this layer doesn't necessarily need to know it.

Why not? To me it seems like a very reasonable assumption but I'm not sure where else we should be handling it if not here.

internal/health/dependencies_test.go
federicobozzini: As you prefer, I think it adds some value but I generally tend to err on the side of over testing.

internal/health/check.go
federicobozzini: I wouldn't.

internal/health/dependencies.go
muchzill4: Primarily, because only runner knows that one `runner.ErrTimeout` means everything else is a timeout. That knowledge should not leak outside of runner's implementation, because it breaks _encapsulation_ and creates nasty (and potentially surprising) coupling.

I'm not sure why you'd want to do this anyway, other than trying to prematurely optimise. I think it's safe to call `runner.BinaryExists` over and over.

internal/health/dependencies_test.go
muchzill4: Let's remove it.

internal/health/dependencies_test.go
```diff
@@ -183,6 +183,67 @@ func TestPerformChecks(t *testing.T) {
...
+			BinaryExistsErr: map[string]error{"docker": runner.ErrTimeout},
+		}
+
+		got := health.PerformChecks(context.Background(), []health.Dependency{dep}, r)
+
+		assert.Len(t, got, 1)
+		assert.ErrorIs(t, got[0].Error, runner.ErrTimeout)
+	})
+
+	t.Run("timeout skips unverified prerequisite dependents and marks the rest as timed out", func(t *testing.T) {
```
muchzill4: Re https://github.com/arm/topo/pull/226/changes#r3119955096, I'd just change this to:

`"timeout skips unverified prerequisite dependents"` 👍

internal/health/dependencies_test.go
federicobozzini: https://github.com/arm/topo/pull/226/commits/af75cb474cabb16d8f63422fc94ca9172ca55243

internal/health/dependencies_test.go
federicobozzini: https://github.com/arm/topo/pull/226/commits/ea6c0b4255946aafb3831c3b09e8ca5a222eb5ba

internal/health/dependencies.go
federicobozzini: https://github.com/arm/topo/pull/226/commits/97818b6e273ceaee55944f2334bbca70d081231b

muchzill4 (approved):
Nice!

PR #228, opened 2026-04-20
refactor: replace `--ignore-buildable` with explicit service names in compose pull

muchzill4:
## Changes

- Add `compose.PullableServices()` which parses compose YAML to identify services without a `build` key
- Replace `docker compose pull --ignore-buildable` with explicit service names passed to `docker compose pull`
- `--ignore-buildable` is docker-specific and not supported by nerdctl or podman
- Enables support for container engines that don't have the convenience of `--ignore-buildable` flag


internal/deploy/operation/docker_compose.go
```diff
@@ -27,8 +30,35 @@ func NewDockerComposeBuild(composeFile string, h command.Host) *DockerCompose {
...
 
-func NewDockerComposePull(composeFile string, h command.Host) *DockerCompose {
-	return NewDockerCompose("Pull images", composeFile, h, []string{"pull", "--ignore-buildable"})
+type DockerComposePull struct {
+	composeFile string
+	host        command.Host
+}
+
+func NewDockerComposePull(composeFile string, h command.Host) *DockerComposePull {
+	return &DockerComposePull{composeFile: composeFile, host: h}
```
tgonzalezorlandoarm, line 37: Have you considered that the `NewDockerComposePull` may return a `DockerCompose` instead? We could do all the compose file parsing logic that is currently performed in `Run` inside `NewDockerComposePull`

internal/compose/parse.go
```diff
@@ -0,0 +1,34 @@
...
+	if err != nil {
+		return nil, err
+	}
+	var cf composeFileSchema
+	if err := yaml.Unmarshal(data, &cf); err != nil {
+		return nil, fmt.Errorf("parsing compose file: %w", err)
+	}
+	var names []string
+	for name, svc := range cf.Services {
+		if svc.Build == nil {
```
tgonzalezorlandoarm: Question out of ignore:
What happens in this case:
```
services:
  zephyr:
    extends:
      file: zephyr-application/compose.yaml
      service: zephyr
```

are we visiting the extended files as well or will `build` be always `nil` here?

internal/compose/parse.go
```diff
@@ -0,0 +1,34 @@
...
+	"gopkg.in/yaml.v3"
+)
+
+type composeFileSchema struct {
+	Services map[string]struct {
+		Build any `yaml:"build"`
+	} `yaml:"services"`
+}
+
+func PullableServices(composeFile io.Reader) ([]string, error) {
```
tgonzalezorlandoarm: will a file like this pull `debug-ui` as well?
```
services:
  app:
    image: myapp

  debug-ui:
    image: my-debug-ui
    profiles: ["debug"]
```

awphi: What's the motivation behind manual yaml parsing here over using `compose.ReadProject`?

internal/deploy/operation/docker_compose.go
```diff
@@ -27,8 +30,35 @@ func NewDockerComposeBuild(composeFile string, h command.Host) *DockerCompose {
...
+}
+
+func (p *DockerComposePull) Description() string { return "Pull images" }
+
+func (p *DockerComposePull) Run(w io.Writer) error {
+	f, err := os.Open(p.composeFile)
+	if err != nil {
+		return fmt.Errorf("reading compose file: %w", err)
+	}
+	defer f.Close() //nolint:errcheck
```
awphi: Question for me more than anything - do you think we should be `nolint`-ing these sorts of defer close error checks or `logger.Error`ing them? We have a mix throughout the codebase atm

internal/compose/parse.go
muchzill4: I felt that `ReadProject` is a bit of a too heavy weight for this. All I need is tiny bit of information from compose file, so it didn't make sense to me to parse all this additional info.

muchzill4: Anything w/o a `build:` will be considered pull-able. What's special about `debug-ui`?

internal/deploy/operation/docker_compose.go
muchzill4: Determining what images to pull feels more natural as part of the `Run`, not construction.

internal/compose/parse.go
muchzill4: Well spotted. I'm definitely not handling extends, so that needs to be improved. 😓

internal/deploy/operation/docker_compose.go
muchzill4: I think it's situational. I think in this case, adding `logger.Error` would be noise that's not really actionable (we're only reading the file, so at worst this is resource leak). Perhaps that's a good rule to follow?

internal/compose/parse.go
muchzill4: Turns out using `compose-go` addresses @tgonzalezorlandoarm comment about `extends`, so two birds with one stone: https://github.com/arm/topo/pull/228/commits/19333cc52fd8e2f4a6430a9702db069f9ecf2b24

internal/compose/parse.go
muchzill4: https://github.com/arm/topo/pull/228/commits/19333cc52fd8e2f4a6430a9702db069f9ecf2b24

awphi (approved):

internal/compose/parse.go
tgonzalezorlandoarm: `docker compose --profile debug up`
is the only way to trigger debug-ui service , so in that case it shouldn't be pullable.
We are currently not proxying this flag through `Topo` so it's not a problem right now, perhaps for the future.

tgonzalezorlandoarm (approved):

PR #230, opened 2026-04-20
fix(templates): report target connectivity failures from probes

tgonzalezorlandoarm:
Currently, `topo templates --target <host>` can fail with a misleading probe-level error such as:

```
"lscpu" not found on remote target's $PATH
```

...even when the real problem is that the target cannot be reached.

The failure comes from the interaction between probing and SSH command lookup:
Hardware compatibility detection starts by probing the target, and the first probe is `CPU()`, which checks for `lscpu` via `runner.(*SSH).BinaryExists`.

In the SSH runner, `BinaryExists` treats non-SSH execution failures as "binary not found", so transport failures can be surfaced as a missing-binary error from the first probe that happened to run.

## Changes

- Preserve SSH and timeout errors in `runner.(*SSH).BinaryExists` instead of rewriting them as missing-binary failures.

### DIFF
--- a/cmd/topo/templates.go
+++ b/cmd/topo/templates.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"errors"
+	"fmt"
 	"os"
 
 	"github.com/arm/topo/internal/catalog"
@@ -31,6 +33,9 @@ var templatesCmd = &cobra.Command{
 			defer cancel()
 			hwProfile, err := probe.Hardware(ctx, r)
 			if err != nil {
+				if errors.Is(err, ssh.ErrSSH) || errors.Is(err, runner.ErrTimeout) {
+					return fmt.Errorf("failed to reach target %s: %w", targetArg, err)
+				}
 				return err
 			}
 			profile = &hwProfile
--- a/internal/runner/ssh.go
+++ b/internal/runner/ssh.go
@@ -62,8 +62,9 @@ func (r *SSH) BinaryExists(ctx context.Context, bin string) error {
 	if err != nil {
 		return err
 	}
+
 	if _, err := r.Run(ctx, cmd); err != nil {
-		if errors.Is(err, ssh.ErrSSH) {
+		if errors.Is(err, ssh.ErrSSH) || errors.Is(err, ErrTimeout) {
 			return err
 		}
 		return fmt.Errorf("%q not found on remote target's $PATH", bin)


cmd/topo/templates.go
```diff
@@ -31,6 +33,9 @@ var templatesCmd = &cobra.Command{
 			defer cancel()
 			hwProfile, err := probe.Hardware(ctx, r)
 			if err != nil {
+				if errors.Is(err, ssh.ErrSSH) || errors.Is(err, runner.ErrTimeout) {
```
muchzill4, line 36: I think it's a wrong place in the stack to handle these errors. If they stem from `BinaryExists` check, it should be `BinaryExists` that returns meaningful errors (`"connectivity error / timeout"`) and not `"not found on $PATH"`.

Ideally, we won't couple the cmd layer to specific kinds of errors a runner (local OR remote) can return.

---

Actually, why even re-wrap it? Can we just pass through the error `probe.Hardware()` returns?

e2e/templates_test.go
```diff
@@ -47,7 +47,18 @@ func TestTemplates(t *testing.T) {
 		testutil.AssertJsonGoldenFile(t, string(out), "testdata/TestTemplatesJson.golden")
 	})
 
-	t.Run("outputs errors as JSON when specified", func(t *testing.T) {
+	t.Run("fails with a connectivity error when the target is unreachable", func(t *testing.T) {
```
muchzill4: I'd probably only unit test this on the runner level (if at all, because it's going to be annoying to do so). E2E is expensive.

muchzill4 (changes requested):

internal/runner/ssh.go
```diff
@@ -62,8 +62,9 @@ func (r *SSH) BinaryExists(ctx context.Context, bin string) error {
 	if err != nil {
 		return err
 	}
+
 	if _, err := r.Run(ctx, cmd); err != nil {
-		if errors.Is(err, ssh.ErrSSH) {
+		if errors.Is(err, ssh.ErrSSH) || errors.Is(err, ErrTimeout) {
```
muchzill4, line 67: Do I understand correctly that this line the actual fix? I'm getting a bit confused by the PR description which states: "even when the real problem is that the target cannot be reached", which should already be handled by `errors.Is(err, ssh.ErrSSH)`. So, what you're addressing here is the timeout issue, which I believe is covered by https://github.com/arm/topo/pull/226.

e2e/templates_test.go
tgonzalezorlandoarm: Done

federicobozzini (changes requested):
@tgonzalezorlandoarm I think you are trying to fix the same issue as #226

tgonzalezorlandoarm:
As pointed out, closing in favour of #226

PR #231, opened 2026-04-20
feat: topo version check in health

awphi:
## Changes

- Adds `version.FetchLatest` which re-implements some of the latest-version grabbing logic from the install scripts in pure go + corresponding tests.
  - I've raised a ticket in the backlog to consider adding a mutable `latest` tag to the publish flow on artifactory in the future but shouldn't block this work as is imo.
- Adds a new `VersionMatches` to fetch + compare two strings and display a nicely-formatted error message if they don't match.
  - Errors encountered while fetching the latest version are only surfaced as a warning log and thus do not result in a red X in the health output
- Adds `topo` itself as a host dependency checked with `topo health` making use of the new `VersionMatches` check
  - Gives a platform-specific fix string to copy-paste to update topo w/ our new scripts!

PR is a little beefier than I'd like - couldn't figure out a way to split this into more neatly without adding unused code, even if temporarily.

### DIFF
--- a/e2e/testdata/TestHealthCheckJson.golden
+++ b/e2e/testdata/TestHealthCheckJson.golden
@@ -3,9 +3,8 @@
     "dependencies": [
       {
         "name": "Topo",
-        "status": "error",
-        "value": "out of date - current: dev, latest version: 4.1.0",
-        "fix": "run `curl -fsSL https://raw.githubusercontent.com/arm/topo/refs/heads/main/scripts/install.sh | sh`"
+        "status": "ok",
+        "value": "topo"
       },
       {
         "name": "SSH",
--- a/e2e/testdata/TestHealthCheckJson.golden
+++ b/e2e/testdata/TestHealthCheckJson.golden
@@ -2,6 +2,12 @@
   "host": {
     "dependencies": [
       {
+        "name": "Topo",
+        "status": "error",
+        "value": "out of date - current: dev, latest version: 4.1.0",
+        "fix": "run `curl -fsSL https://raw.githubusercontent.com/arm/topo/refs/heads/main/scripts/install.sh | sh`"
+      },
+      {
         "name": "SSH",
         "status": "ok",
         "value": "ssh"
--- a/internal/health/check.go
+++ b/internal/health/check.go
@@ -54,7 +54,7 @@ type VersionMatches struct {
 	Fix            string
 }
 
-func (v VersionMatches) Run(ctx context.Context, r runner.Runner, dep Dependency) (string, error) {
+func (v VersionMatches) Run(ctx context.Context, _ runner.Runner, _ Dependency) (string, error) {
 	latest, err := v.FetchLatest(ctx)
 	if err != nil {
 		logger.Warn(fmt.Sprintf("failed to fetch latest version: %v", err))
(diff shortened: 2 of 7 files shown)


e2e/testdata/TestHealthCheckJson.golden
```diff
@@ -1,6 +1,12 @@
 {
   "host": {
     "dependencies": [
+      {
+        "name": "Topo",
+        "status": "error",
+        "value": "out of date - current: dev, latest version: 4.1.0",
```
awphi: I didn't add an escape hatch when `version.Version == "dev"` so we can see the error message but arguably this poor DX for us? Open to ideas

awphi: Nvm, added an escape hatch here https://github.com/arm/topo/pull/231/commits/ab5fd72c9242f55f4a34c7258c4517590467f6ce to get around the fact that fix strings are platform dependent

muchzill4 (commented):
1. Do we want to "error" (❌) when version is not latest? Shouldn't this be more _gentle_? Info?

2. I wonder how this would feel if it wasn't part of the `Host` block. It's a bit weird that it's "a dependency", while we already have `topo` installed. It's almost a separate `Topo` section. 🤔 WDYT?

awphi:
> 1. Do we want to "error" (❌) when version is not latest? Shouldn't this be more _gentle_? Info?

Yep good point!

> 2. I wonder how this would feel if it wasn't part of the `Host` block. It's a bit weird that it's "a dependency", while we already have `topo` installed. It's almost a separate `Topo` section. 🤔 WDYT?

My intention was to make it clear we only need topo installed on the host and not the target but I see your point, even if we don't surface the 'dependency' terminology to the user

That being said, sketching it out has made your suggestion grow on me but will need some refactoring to do this since I'm not sure the dependency terminology we use in the codebase will make sense anymore;

```
Topo
----
Version: ℹ️ (current: v4.1.0, latest: v4.2.0)
  → run `curl -fsSL https://raw.githubusercontent.com/arm/topo/refs/heads/main/scripts/install.sh | sh` to update

...
```

```
Topo
----
Version: ✅  (current: v4.2.0, latest: v4.2.0)

...
```

muchzill4:
Happy for it to land as info in `Host` and we can then decide if we want a separate section. 👍

awphi:
> Happy for it to land as info in Host and we can then decide if we want a separate section. 👍

Okay nice done! https://github.com/arm/topo/pull/231/commits/fd1834ebc3958d680e89c1adbf8b79bc66430883

internal/health/check.go
```diff
@@ -45,3 +47,22 @@ func (b BinaryExists) Run(ctx context.Context, r runner.Runner, dep Dependency)
...
 	return "", nil
 }
+
+type VersionMatches struct {
+	CurrentVersion string
+	FetchLatest    func(ctx context.Context) (string, error)
+	Fix            string
+}
+
+func (v VersionMatches) Run(ctx context.Context, r runner.Runner, dep Dependency) (string, error) {
```
muchzill4: I think we might want to revisit the `Check` interface. Looks like `Run()` with all arguments the arguments is not that useful for `VersionMatches`, perhaps we should be explicit by using `_` for `Runner` and `Dependency`.

muchzill4 (approved):

internal/health/check.go
awphi: Will add to the inbox to have a think about the shape of this

PR #232, opened 2026-04-21
fix: improve error message phrasing consistency

awphi:
## Changes

- Makes a pass over the repo to improve error message consistency to adhere to the form; "failed to X %s: %w"
- Uncapitalises SSH in the ssh auth errors - felt better to me but happy to be disagreed with.
- Removes a couple error logging when closing resources which are not actionable + adds nolint instead

### DIFF
--- a/internal/ssh/config_file.go
+++ b/internal/ssh/config_file.go
@@ -7,7 +7,6 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/arm/topo/internal/output/logger"
 	sshconfig "github.com/kevinburke/ssh_config"
 )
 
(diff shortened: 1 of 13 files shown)


internal/ssh/config_file.go
```diff
@@ -80,13 +79,7 @@ func readConfigFile(path string) (*sshconfig.Config, error) {
...
 		return nil, fmt.Errorf("failed to open topo ssh config file: %w", err)
 	}
-	defer func() {
-		if cfgFile != nil {
-			if err := cfgFile.Close(); err != nil {
-				logger.Error("failed to close topo ssh config file", "error", err)
-			}
-		}
-	}()
+	defer cfgFile.Close() // nolint:errcheck
```
federicobozzini, line 82: Why this change?

awphi: Idea suggested here https://github.com/arm/topo/pull/228#discussion_r3112114812

awphi:
Changed to `fix:` as thought users might need to care about some of this

tgonzalezorlandoarm (approved):

PR #234, opened 2026-04-22
fix: handling of health timeout error for remoteproc

federicobozzini:
## Changes

- Propagate probe errors instead of swallowing them.
- Carry the probe error through to the report layer.
- Display the probe error instead of "no remoteproc devices found".
- Tests for error propagation and reporting.

Before:

```sh
$ ./topo health --target imx93 --timeout 200ms 
Host
----
SSH: ✅ (ssh)
Container Engine: ✅ (docker)

Target
------
Connectivity: ✅
Container Engine: ❌ (timed out)
Hardware Info: ❌ (timed out)
Subsystem Driver (remoteproc): ℹ️ (no remoteproc devices found)
```
After

```sh
$ ./topo health --target imx93 --timeout 200ms 
Host
----
SSH: ✅ (ssh)
Container Engine: ✅ (docker)

Target
------
Connectivity: ✅
Container Engine: ❌ (timed out)
Hardware Info: ❌ (timed out)
Subsystem Driver (remoteproc): ❌ (timed out)
```

### DIFF
--- a/internal/health/status.go
+++ b/internal/health/status.go
@@ -8,7 +8,8 @@ import (
 )
 
 type HardwareProfile struct {
-	RemoteCPU []probe.RemoteprocCPU
+	RemoteCPU    []probe.RemoteprocCPU
+	RemoteCPUErr error
 }
 
 func (h HardwareProfile) Capabilities() map[HardwareCapability]struct{} {
@@ -27,8 +28,9 @@ type HealthStatus struct {
 func ProbeHealthStatus(ctx context.Context, r runner.Runner) HealthStatus {
 	var hs HealthStatus
 
-	remoteprocs, _ := probe.Remoteproc(ctx, r)
+	remoteprocs, err := probe.Remoteproc(ctx, r)
 	hs.Hardware.RemoteCPU = remoteprocs
+	hs.Hardware.RemoteCPUErr = err
 
 	dependenciesToCheck := FilterByHardware(TargetRequiredDependencies, hs.Hardware.Capabilities())
 	hs.Dependencies = PerformChecks(ctx, dependenciesToCheck, r)
--- a/internal/health/status.go
+++ b/internal/health/status.go
@@ -8,8 +8,8 @@ import (
 )
 
 type HardwareProfile struct {
-	RemoteCPU    []probe.RemoteprocCPU
-	RemoteCPUErr error
+	RemoteCPU []probe.RemoteprocCPU
+	Err       error
 }
 
 func (h HardwareProfile) Capabilities() map[HardwareCapability]struct{} {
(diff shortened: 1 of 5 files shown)


internal/health/status.go
```diff
@@ -8,7 +8,8 @@ import (
 )
 
 type HardwareProfile struct {
-	RemoteCPU []probe.RemoteprocCPU
+	RemoteCPU    []probe.RemoteprocCPU
+	RemoteCPUErr error
```
muchzill4: This can be just `Err`, right?

internal/probe/remoteproc.go
```diff
@@ -13,8 +13,11 @@ type RemoteprocCPU struct {
 
 func Remoteproc(ctx context.Context, r runner.Runner) ([]RemoteprocCPU, error) {
 	var remoteProcs []RemoteprocCPU
-	out, err := r.Run(ctx, "ls /sys/class/remoteproc")
-	if err != nil || out == "" {
+	out, err := r.Run(ctx, "ls /sys/class/remoteproc 2>/dev/null || true")
```
federicobozzini: I'm not crazy about this change, but we need a way to handle the case where the remoteproc kernel module is not installed.

internal/probe/remoteproc.go
```diff
@@ -14,7 +14,10 @@ type RemoteprocCPU struct {
 func Remoteproc(ctx context.Context, r runner.Runner) ([]RemoteprocCPU, error) {
 	var remoteProcs []RemoteprocCPU
 	out, err := r.Run(ctx, "ls /sys/class/remoteproc")
-	if err != nil || out == "" {
+	if err != nil {
+		return remoteProcs, err
+	}
+	if out == "" {
```
muchzill4: Previous one was ok I think. If you are after explicit returns we'd also need to change returns to empty slices. I think it'd be too much.

internal/probe/remoteproc.go
muchzill4: Why add the redirect? Seems outside of scope of this PR.

federicobozzini: You missed my comment:

> I'm not crazy about this change, but we need a way to handle the case where the remoteproc kernel module is not installed.

As an example this change failed on the CI because alpine doesn't have it.

muchzill4: > You missed my comment:

I must've been mid-review. 👍 

> As an example this change failed on the CI because alpine doesn't have it.

Can you link to CI failure? I'm surprised this suddenly started to fail. The `|| true` should handle the case of `ls` trying to list a directory that doesn't exist.

federicobozzini: https://github.com/arm/topo/actions/runs/24781239237

internal/health/status.go
federicobozzini: I'm not sure. If tomorrow we wanted to support a different type of subsystem driver, there could be a different failure for it.

muchzill4: YAGNI

internal/probe/remoteproc.go
federicobozzini: https://github.com/arm/topo/pull/234/commits/486c0f10c3815217f3c8bfec5de003ac4019fd7d

internal/health/status.go
federicobozzini: https://github.com/arm/topo/pull/234/commits/eae0281308f2549d917305a6a8570ff7d5d76d83

internal/probe/remoteproc.go
muchzill4: Oh my.. I didn't notice that `|| true` was added here as well.

Perhaps `test -d /sys/class/remoteproc` will work better in that case?

muchzill4 (approved):

PR #235, opened 2026-04-22
feat: skip-version-checks flag in topo health

awphi:
## Changes

- Adds a `--skip-version-checks` to `topo health` for use by consumers that don't want to hit artifactory every time health is checked
  - Opted to pluralise this in case we add other version checks to `topo health` in the future. Maybe too pre-emptive?
- Adds util to filter all version checks from a list of dependencies that is used when this flag is set :)

Importantly I opted not to filter out dependencies whose only check is of type `VersionMatches` to keep the list of dependencies consistent.

### DIFF
--- a/cmd/topo/health.go
+++ b/cmd/topo/health.go
@@ -12,7 +12,10 @@ import (
 	"github.com/spf13/cobra"
 )
 
-const acceptNewHostFlag = "accept-new-host-keys"
+const (
+	acceptNewHostFlag     = "accept-new-host-keys"
+	skipVersionChecksFlag = "skip-version-checks"
+)
 
 var healthCmd = &cobra.Command{
 	Use:   "health",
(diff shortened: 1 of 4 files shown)


cmd/topo/health.go
```diff
@@ -27,13 +30,19 @@ var healthCmd = &cobra.Command{
 		if err != nil {
 			panic(fmt.Sprintf("internal error: %s flag not registered: %v", acceptNewHostFlag, err))
 		}
+
+		skipVersionCheck, err := cmd.Flags().GetBool(skipVersionChecksFlag)
```
muchzill4: I think we'd benefit from `TOPO_SKIP_VERSION_CHECKS` env variable too.

awphi: good idea https://github.com/arm/topo/pull/235/commits/2ef5aa0f1bf2938df0db1b1c3e978a23aa3ffff6

internal/health/check.go
```diff
@@ -66,3 +67,14 @@ func (v VersionMatches) Run(ctx context.Context, _ runner.Runner, _ Dependency)
 
 	return v.Fix, InfoError{Err: fmt.Errorf("out of date - current: %s, latest version: %s", v.CurrentVersion, latest)}
 }
+
+func FilterVersionChecks(deps []Dependency) []Dependency {
```
muchzill4: Should this be `Drop|Exclude|Without`? I never know with "filter" which way it is filtering.

cmd/topo/health.go
```diff
@@ -12,7 +12,25 @@ import (
...
 
-const acceptNewHostFlag = "accept-new-host-keys"
+const (
+	acceptNewHostFlag     = "accept-new-host-keys"
+	skipVersionChecksFlag = "skip-version-checks"
+)
+
+const skipVersionChecksEnvVar = "TOPO_SKIP_VERSION_CHECKS"
+
+func resolveSkipVersionChecks(cmd *cobra.Command) bool {
```
muchzill4: Should we put this closer to the bottom? My thinking is always "public api" first.

muchzill4 (approved):

cmd/topo/health.go
awphi: https://github.com/arm/topo/pull/235/commits/cb7a8f0efec12d56fb7cd51f8d81cbd8f366a2aa

internal/health/check.go
awphi: https://github.com/arm/topo/pull/235/commits/cb7a8f0efec12d56fb7cd51f8d81cbd8f366a2aa

muchzill4 (approved):

PR #236, opened 2026-04-23
feat: upgrade command

awphi:
## Changes

- Adds a `topo upgrade` command. General steps are;
  - Check for latest version from artifactory, if already on it then bail
  - If not then construct the archive URL
  - Download the archive into memory
  - Extract the archive into a temp file on disk, in the same directory as the current running `topo` binary. This ensures atomic renames always work as there's no cross-filesystem errors to worry about.
  - On windows; rename current topo to topo.old, rename temp file to topo
    - Windows does not support deleting running programs, so we do this rename dance. This also means we leave a `topo.exe.old` on their disk which we overwrite next time `topo upgrade` is executed
  - Otherwise; rename temp file to `topo`, deleting the current running binary + set execute perms on it
- Adds a `version.Placeholder` constant to avoid hard-coding "dev" all over the place
- Makes a best effort to warn users of any files we created and failed to cleanup and errors if we failed to install and couldn't restore the old version with logging.

## Next Steps

After this PR the plan is to remove the upgrading mechanism from the install scripts and point users towards `topo upgrade` when it's already installed. This makes them more akin to bootstrap scripts rather than idempotent install/update scripts, centralizing update logic in the CLI.

## Testing

Recommend testing this by changing `version.Version` to something like "0.0.1" to avoid hitting the "dev" escape hatch.

---

Apologies for the big diff as well! Couldn't see a clear point to split it aside from things like dropping windows support temporarily which I'm not keen on. Feel free to push back if you think it's not reviewable.

### DIFF
--- a/cmd/topo/upgrade.go
+++ b/cmd/topo/upgrade.go
@@ -2,7 +2,7 @@ package main
 
 import (
 	"fmt"
-	"runtime"
+	"os"
 
 	"github.com/arm/topo/internal/output/term"
 	"github.com/arm/topo/internal/upgrade"
@@ -19,40 +19,26 @@ var upgradeCmd = &cobra.Command{
 		cmd.SilenceUsage = true
 		outputFormat := resolveOutput(cmd)
 
+		var spinner *term.Spinner
+		if outputFormat == term.Plain {
+			spinner = term.StartSpinner(os.Stderr, "Upgrading topo...")
+		}
+
 		ctx, cancel := contextWithTimeout(cmd)
 		defer cancel()
 
-		current := version.Version
-		var latest string
-		err := term.WithSpinner(outputFormat, "Checking for updates...", func() (err error) {
-			latest, err = version.FetchLatest(ctx, version.ArtifactoryBaseURL)
-			return err
-		})
+		newVersion, err := upgrade.Upgrade(ctx, spinner)
+		spinner.Stop()
 		if err != nil {
 			return err
 		}
 
-		if current == latest || current == version.Dev {
-			fmt.Printf("topo %s is already up to date\n", current)
+		if newVersion == version.Version {
+			fmt.Printf("Already running the latest version of topo (%s)\n", newVersion)
 			return nil
 		}
 
-		fmt.Printf("Upgrading topo from %s to %s\n", current, latest)
-
-		binPath, err := upgrade.CurrentBinaryPath()
-		if err != nil {
-			return fmt.Errorf("failed to determine current binary path: %w", err)
-		}
-
-		err = term.WithSpinner(outputFormat, fmt.Sprintf("Downloading topo %s...", latest), func() error {
-			downloadURL := upgrade.ArtifactoryDownloadURL(runtime.GOOS, runtime.GOARCH, latest)
-			return upgrade.Install(ctx, binPath, downloadURL)
-		})
-		if err != nil {
-			return err
-		}
-
-		fmt.Printf("Upgraded topo to %s\n", latest)
+		fmt.Printf("Upgraded topo to %s\n", newVersion)
 		return nil
 	},
 }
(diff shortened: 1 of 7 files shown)


cmd/topo/upgrade.go
```diff
@@ -0,0 +1,71 @@
...
+		if err != nil {
+			return err
+		}
+
+		fmt.Printf("Upgraded topo to %s\n", latest)
+		return nil
+	},
+}
+
+func withSpinner(outputFormat term.Format, msg string, fn func() error) error {
```
muchzill4: Neat! Should this live in `spinner`? 🤷

internal/version/version.go
```diff
@@ -1,8 +1,10 @@
 package version
 
 var (
+	Placeholder = "dev"
```
muchzill4: Maybe call it `Dev`?

internal/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,192 @@
...
+		ctx, cancel := context.WithCancel(context.Background())
+		cancel()
+
+		err := upgrade.Install(ctx, createFakeBinary(t), srv.URL)
+
+		assert.Error(t, err)
+	})
+}
+
+func TestArtifactoryDownloadURL(t *testing.T) {
```
muchzill4, line 156: I wonder if this would read better as table tests.

cmd/topo/upgrade.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/arm/topo/internal/version"
+	"github.com/spf13/cobra"
+)
+
+var upgradeCmd = &cobra.Command{
+	Use:   "upgrade",
+	Short: "Upgrade topo to the latest version",
+	Long:  "Download and install the latest version of topo from Artifactory, replacing the current binary in place.",
+	Args:  cobra.ExactArgs(0),
+	RunE: func(cmd *cobra.Command, args []string) error {
```
muchzill4, line 18: NIT: This feels a bit too logic dense for the cli layer. Ideally the responsibilities of a cli layer should end at parsing input and translating that to a domain/service call, which return value is then translated back to the cli realm.

I appreciate this is a bit difficult to achieve with spinner right now. Perhaps we could have `upgrade.Verb()` take a progress reporter interface, which would wrap spinner. Inside of the upgrade code, you could then say: `reporter.Step("Checking for updates...")` or such.

awphi: Added an interface implemented by `Spinner` for doing this - quite like how the result looks but lmk if you disagree.

Also condensed all the logic originally in this layer into `upgrade.Upgrade` - it feels a bit wrong to not be able to unit test this function but I suppose that's what e2e tests are for (although there aren't any for this command quite yet :P)

https://github.com/arm/topo/pull/236/commits/bc00fe2cfe710c5c1fd4500f08699c45db11aea6

internal/upgrade/upgrade_test.go
awphi: https://github.com/arm/topo/pull/236/commits/8de40353c36818d6aac61d95eb27226dbedccb7d

cmd/topo/upgrade.go
awphi: https://github.com/arm/topo/pull/236/commits/90579b36893747826f099f68b9e04b0de1ad9ba6 (though rendered obsolete by the spinner rejig)

internal/version/version.go
awphi: https://github.com/arm/topo/pull/236/commits/90579b36893747826f099f68b9e04b0de1ad9ba6

cmd/topo/upgrade.go
muchzill4: _chef's kiss_

muchzill4 (approved):

PR #238, opened 2026-04-23
refactor: always enable SSH multiplexing

muchzill4:
## Changes

- Remove `SSHOptions` — SSH multiplexing is now always enabled (except on Windows)
- Simplify `NewSSH` and `For` signatures by removing the options parameter
- Move private functions (`exec`, `multiplexArgs`) to bottom of file


awphi (approved):
🚀

PR #239, opened 2026-04-23
refactor: remove unnecessary dir existence check for remoteproc

federicobozzini:
## Changes

- As the tytle says, checking for the existence of `/sys/class/remoteproc` is unnecessary. Check removed.


muchzill4 (approved):
The diff I crave.

PR #240, opened 2026-04-24
fix: handle ssh's too many authentication failures in health check

tgonzalezorlandoarm:
## Description

We are currently only suggesting `setup-keys` when we try to authenticate via SSH and we get a `permission denied` error. We categorize this error as an Authentication Error.

We could, however, also get a `too many authentication failures` error when trying to connect, which should also be an Authentication Error, but we are failing to categorize it as such.

`too many authentication failures` suggests that we haven't setup the keys for that target, although this is not always the case:

It could be that keys for that target are already setup but there are too many keys in the SSH agent (so it attempts several keys before attempting the correct one). This can happen if the user has not added the key for that target in their .ssh/config , which is a configuration mistake from behalf of the user.

In any case, running setup-keys will resolve the issue, as the newly created key is also added to the topo_config.

## Changes

- fail the health check on SSH's `too many authentication failures` and suggest `setup-keys` to resolve the issue.

### DIFF
--- a/internal/ssh/errors.go
+++ b/internal/ssh/errors.go
@@ -28,7 +28,7 @@ func ClassifyStderr(stderr string) error {
 	if strings.Contains(lower, "timed out") {
 		return ErrConnectionTimeout
 	}
-	if strings.Contains(lower, "permission denied") {
+	if strings.Contains(lower, "permission denied") || strings.Contains(lower, "too many authentication failures") {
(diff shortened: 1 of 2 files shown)


federicobozzini (commented):
Can we add a test for this please?

muchzill4 (approved):

PR #241, opened 2026-04-24
fix: distinguish SSH auth exhaustion from invalid credentials

tgonzalezorlandoarm:
## Description
   
We already classify `too many authentication failures` as a distinct SSH error, but the health report is still treating it like a generic auth failure.

## Changes

Handle this case explicitly in the probe and health layers so the user gets a more accurate diagnostic of the failure's cause:
- map ssh `too many authentication failures` to a dedicated probe error

This keeps `permission denied` and `too many authentication failures` separate, which better reflects the likely cause of the failure.

### DIFF
--- a/internal/health/health.go
+++ b/internal/health/health.go
@@ -160,7 +160,7 @@ func connectivityCheck(status ConnectionStatus) HealthCheck {
 
 	check.Value = status.Error.Error()
 	switch {
-	case errors.Is(status.Error, probe.ErrAuthFailed):
+	case errors.Is(status.Error, probe.ErrAuthFailed) || errors.Is(status.Error, probe.ErrTooManyAuthFails):
 		check.Fix = &Fix{
 			Description: "Configure SSH keys on remote target",
 			Command:     fmt.Sprintf("topo setup-keys --target %s", status.Destination),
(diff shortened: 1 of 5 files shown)


internal/health/health.go
```diff
@@ -165,6 +165,11 @@ func connectivityCheck(status ConnectionStatus) HealthCheck {
 			Description: "Configure SSH keys on remote target",
 			Command:     fmt.Sprintf("topo setup-keys --target %s", status.Destination),
 		}
+	case errors.Is(status.Error, probe.ErrTooManyAuthFails):
+		check.Fix = &Fix{
+			Description: "Configure SSH to use the correct key for this target or run the setup-keys command",
+			Command:     fmt.Sprintf("topo setup-keys --target %s", status.Destination),
+		}
```
muchzill4: I'd probably fold this into the `ErrAuthFailed` case above. I'm not convinced "Configure SSH to use the correct key for this target" provides value over just saying "run `topo setup-keys`", which fill fix the issue.

tgonzalezorlandoarm: done

muchzill4 (approved):

PR #242, opened 2026-04-24
refactor: pluralise HW profile  properties

federicobozzini:
## Changes

- `host` renamed to `hosts` in `topo describe` output
- internal renaming to make sure properties are pluralised correctly
- tagged as refactor because `topo describe` is not part of the public API

### DIFF
--- a/internal/health/health_test.go
+++ b/internal/health/health_test.go
@@ -49,7 +49,7 @@ func TestGenerateTargetReport(t *testing.T) {
 	t.Run("when remoteproc devices are found, SubsystemDriver status is ok and includes device names", func(t *testing.T) {
 		ts := health.Status{
 			Hardware: health.HardwareProfile{
-				RemoteCPU: []probe.RemoteprocCPU{{Name: "m4_0"}, {Name: "m4_1"}},
+				RemoteCPUs: []probe.RemoteprocCPU{{Name: "m4_0"}, {Name: "m4_1"}},
 			},
 		}
 
--- a/internal/health/health_test.go
+++ b/internal/health/health_test.go
@@ -49,7 +49,7 @@ func TestGenerateTargetReport(t *testing.T) {
 	t.Run("when remoteproc devices are found, SubsystemDriver status is ok and includes device names", func(t *testing.T) {
 		ts := health.Status{
 			Hardware: health.HardwareProfile{
-				RemoteProcessors: []probe.RemoteProc{{Name: "m4_0"}, {Name: "m4_1"}},
+				RemoteProcessors: []probe.RemoteProcessor{{Name: "m4_0"}, {Name: "m4_1"}},
 			},
 		}
 
--- a/internal/health/health_test.go
+++ b/internal/health/health_test.go
@@ -49,7 +49,7 @@ func TestGenerateTargetReport(t *testing.T) {
 	t.Run("when remoteproc devices are found, SubsystemDriver status is ok and includes device names", func(t *testing.T) {
 		ts := health.Status{
 			Hardware: health.HardwareProfile{
-				RemoteCPUs: []probe.RemoteprocCPU{{Name: "m4_0"}, {Name: "m4_1"}},
+				RemoteProcessors: []probe.RemoteProc{{Name: "m4_0"}, {Name: "m4_1"}},
 			},
 		}
 
--- a/internal/health/status.go
+++ b/internal/health/status.go
@@ -8,7 +8,7 @@ import (
 )
 
 type HardwareProfile struct {
-	RemoteProcessors []probe.RemoteProc
+	RemoteProcessors []probe.RemoteProcessor
 	Err              error
 }
 
@@ -28,7 +28,7 @@ type HealthStatus struct {
 func ProbeHealthStatus(ctx context.Context, r runner.Runner) HealthStatus {
 	var hs HealthStatus
 
-	remoteProcessors, err := probe.Remoteprocs(ctx, r)
+	remoteProcessors, err := probe.Remoteproc(ctx, r)
 	hs.Hardware.RemoteProcessors = remoteProcessors
 	hs.Hardware.Err = err
 
--- a/internal/health/status.go
+++ b/internal/health/status.go
@@ -8,13 +8,13 @@ import (
 )
 
 type HardwareProfile struct {
-	RemoteCPUs []probe.RemoteprocCPU
-	Err        error
+	RemoteProcessors []probe.RemoteProc
+	Err              error
 }
 
 func (h HardwareProfile) Capabilities() map[HardwareCapability]struct{} {
 	capabilities := make(map[HardwareCapability]struct{})
-	if len(h.RemoteCPUs) > 0 {
+	if len(h.RemoteProcessors) > 0 {
 		capabilities[Remoteproc] = struct{}{}
 	}
 	return capabilities
@@ -28,8 +28,8 @@ type HealthStatus struct {
 func ProbeHealthStatus(ctx context.Context, r runner.Runner) HealthStatus {
 	var hs HealthStatus
 
-	remoteprocs, err := probe.Remoteproc(ctx, r)
-	hs.Hardware.RemoteCPUs = remoteprocs
+	remoteProcessors, err := probe.Remoteprocs(ctx, r)
+	hs.Hardware.RemoteProcessors = remoteProcessors
 	hs.Hardware.Err = err
 
 	dependenciesToCheck := FilterByHardware(TargetRequiredDependencies, hs.Hardware.Capabilities())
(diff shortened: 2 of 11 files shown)


internal/health/status.go
```diff
@@ -8,13 +8,13 @@ import (
 )
 
 type HardwareProfile struct {
-	RemoteCPU []probe.RemoteprocCPU
-	Err       error
+	RemoteCPUs []probe.RemoteprocCPU
```
federicobozzini: Personally I would also rename this to `RemoteCpus` but I'm not too fussy about it.

internal/probe/hardware.go
```diff
@@ -8,25 +8,25 @@ import (
 )
 
 type HardwareProfile struct {
-	HostProcessor []HostProcessor `yaml:"host" json:"host"`
-	RemoteCPU     []RemoteprocCPU `yaml:"remoteprocs" json:"remoteprocs,omitempty"`
-	TotalMemoryKb int64           `yaml:"totalmemory_kb" json:"totalmemory_kb"`
+	HostProcessors []HostProcessor `yaml:"hosts" json:"hosts"`
```
awphi: The yaml/json tag renaming makes this a breaking change so need a `!` in the PR title.

I also wonder if we want to take this opportunity to consider if 'Host' is the best term we can come up with for this. While it's technically accurate it's a bit confusing with our host/target terminology atm.

internal/output/templates/describe_test.go
```diff
@@ -49,10 +49,10 @@ totalmemory_kb: 16384
 		t.Run("renders valid JSON with all fields", func(t *testing.T) {
 			toPrint := templates.PrintableTargetDescription{
 				HardwareProfile: probe.HardwareProfile{
-					HostProcessor: []probe.HostProcessor{
+					HostProcessors: []probe.HostProcessor{
 						{Model: "Cortex-A55", Cores: 4, Features: []string{"asimd", "sve"}},
 					},
-					RemoteCPU: []probe.RemoteprocCPU{
+					RemoteCPUs: []probe.RemoteprocCPU{
```
awphi: Another weird naming thing we could consider consolidating here is `Processor` vs `CPU` - WDYT?

internal/probe/hardware.go
federicobozzini: > The yaml/json tag renaming makes this a breaking change so need a ! in the PR title.

I mentioned it in the description:

> tagged as refactor because topo describe is not part of the public API

> I also wonder if we want to take this opportunity to consider if 'Host' is the best term we can come up with for this. While it's technically accurate it's a bit confusing with our host/target terminology atm.

I think this might definitely deserve a discussion, but i would not do it in this PR.

internal/output/templates/describe_test.go
federicobozzini: I agree, it would be better to settle on one term here.

internal/probe/hardware.go
awphi: > I mentioned it in the description

Ah sorry missed it - I don't agree though, the command is not hidden and appears in `topo --help` - I would definitely consider it part of the public API even if not outlined in the README.

federicobozzini: I don't see it listed and [it shouldn't be](https://github.com/arm/topo/blob/main/cmd/topo/vscode.go#L31). Are you sure?

```sh
$ topo --help
Topo CLI

Usage:
  topo [command]

Available Commands:
  clone       Clone an example project
  completion  Generate the autocompletion script for the specified shell
  deploy      Deploy services using the compose file
  extend      Add services from a template to the compose file
  health      Check the target host environment
  help        Help about any command
  init        Initialise a new project in the current directory
  install     Install components to the target
  service     Manage services in compose files
  setup-keys  Generate SSH keys for the target and install the public key on the target host
  stop        Stop a currently running deployment
  templates   List available service templates
  upgrade     Upgrade topo to the latest version

Flags:
  -h, --help            help for topo
  -o, --output string   output format: plain or json (default "plain")
  -v, --version         version for topo

Use "topo [command] --help" for more information about a command.
```

awphi: Ah completely forgot we'd hidden it. You'd think I'd fact check myself but alas Friday afternoon - sorry for the noise!

internal/health/status.go
muchzill4: I swear it used to be the case that you had to use ID instead of Id in golang, otherwise linter would yell at you. Same would probably go for CPU vs Cpu, SSH vs Ssh, etc. It no longer seems to be the case? 

However.. If we were to break the convention we've painstakingly established, I think we should do it in one big sweep in a separate PR.

internal/output/templates/describe_test.go
awphi: Would vote for processor since `remoteproc` is an established term so sticking to its language seems like the logical thing to do.

`HostProcessors` and `RemoteProcessors`?

federicobozzini: I think technically speaking `RemoteCpu` is even wrong, because something like a DSP (something managed by remoteproc and possibly targeted by topo) is a processor but not a CPU.

federicobozzini: This should do the trick https://github.com/arm/topo/pull/242/commits/26f50d3a25754daf56ac6478633521b67956d87c

internal/health/health_test.go
```diff
@@ -49,7 +49,7 @@ func TestGenerateTargetReport(t *testing.T) {
 	t.Run("when remoteproc devices are found, SubsystemDriver status is ok and includes device names", func(t *testing.T) {
 		ts := health.Status{
 			Hardware: health.HardwareProfile{
-				RemoteCPU: []probe.RemoteprocCPU{{Name: "m4_0"}, {Name: "m4_1"}},
+				RemoteProcessors: []probe.RemoteProc{{Name: "m4_0"}, {Name: "m4_1"}},
```
muchzill4: `RemoteProcessors` and `RemoteProc`? Can we make it consistent?

internal/probe/remoteproc.go
```diff
@@ -8,12 +8,12 @@ import (
...
 )
 
-type RemoteprocCPU struct {
+type RemoteProc struct {
 	Name string `yaml:"name" json:"name"`
 }
 
-func Remoteproc(ctx context.Context, r runner.Runner) ([]RemoteprocCPU, error) {
-	var remoteProcs []RemoteprocCPU
+func Remoteprocs(ctx context.Context, r runner.Runner) ([]RemoteProc, error) {
```
muchzill4: `Remoteprocs`, but then `RemoteProc`. 😕 

I think probe should remain `Remoteproc`, because it's a `Remoteproc` probe.
It could return `[]RemoteProcessor`.

internal/probe/hardware.go
```diff
@@ -8,25 +8,25 @@ import (
 )
 
 type HardwareProfile struct {
-	HostProcessor []HostProcessor `yaml:"host" json:"host"`
-	RemoteCPU     []RemoteprocCPU `yaml:"remoteprocs" json:"remoteprocs,omitempty"`
-	TotalMemoryKb int64           `yaml:"totalmemory_kb" json:"totalmemory_kb"`
+	HostProcessors   []HostProcessor `yaml:"hostProcessors" json:"hostProcessors"`
+	RemoteProcessors []RemoteProc    `yaml:"remoteProcessors" json:"remoteProcessors,omitempty"`
+	TotalMemoryKb    int64           `yaml:"totalmemory_kb" json:"totalmemory_kb"`
```
awphi: Seems weird to be mixing camelCase and snake_case in our yaml/json tags, especially now within the same object.

I think our convention thus far has been camelCase so maybe change totalmemory_kb to match?

internal/probe/remoteproc.go
federicobozzini: https://github.com/arm/topo/pull/242/commits/5ff9685f33492a396076cafd4946d94366f2e6de

internal/health/health_test.go
federicobozzini: https://github.com/arm/topo/pull/242/commits/5ff9685f33492a396076cafd4946d94366f2e6de

internal/probe/hardware.go
federicobozzini: https://github.com/arm/topo/pull/242/commits/4bfe0b66a8cbd8f319e49d867e4f0adb02cd6396

internal/probe/host_processors_test.go
muchzill4: Should've been `TestCPU`, so historically wrong. Now should be `TestHostProcessors`.

federicobozzini: https://github.com/arm/topo/pull/242/commits/841511c216629ddc0aa9d60a32a12b81fce814e8

awphi (approved):

PR #245, opened 2026-04-27
feat: direct install script users to topo upgrade

awphi:
## Changes

- Removes the ability of the install scripts to replace `topo` binaries already on the PATH in-place and directs them towards `topo upgrade`
- Overriding existing binaries when using the `--path` flag continues to function as normal
- Technically a breaking change to the install scripts but didn't mark as such in the PR title since it's not breaking to the CLI

### DIFF
--- a/e2e/install_script_posix_test.go
+++ b/e2e/install_script_posix_test.go
@@ -84,8 +84,6 @@ func TestInstallScript(t *testing.T) {
 		require.NoError(t, err, "script failed: %s", out)
 		assert.Contains(t, out, "topo is already installed")
 		assert.Contains(t, out, "topo upgrade")
-		assert.NotContains(t, out, "Installing topo")
-		assert.NotContains(t, out, "Installed topo")
 	})
 
 	t.Run("fails on unknown flag", func(t *testing.T) {
--- a/e2e/install_script_posix_test.go
+++ b/e2e/install_script_posix_test.go
@@ -8,6 +8,7 @@ import (
 	"path/filepath"
 	"testing"
 
+	"github.com/arm/topo/internal/testutil"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
@@ -19,12 +20,18 @@ func scriptPath(t *testing.T) string {
 	return path
 }
 
-func runInstallScript(t *testing.T, args ...string) (string, error) {
+func runInstallScriptWithEnv(t *testing.T, env []string, args ...string) (string, error) {
 	t.Helper()
 	cmdArgs := append([]string{scriptPath(t)}, args...)
 	cmd := exec.Command("sh", cmdArgs...)
+	cmd.Env = append(os.Environ(), env...)
 	out, err := cmd.CombinedOutput()
 	return string(out), err
+}
+
+func runInstallScript(t *testing.T, args ...string) (string, error) {
+	t.Helper()
+	return runInstallScriptWithEnv(t, nil, args...)
 }
 
 func TestInstallScript(t *testing.T) {
@@ -57,7 +64,7 @@ func TestInstallScript(t *testing.T) {
 		assert.FileExists(t, filepath.Join(dir, "topo"))
 	})
 
-	t.Run("can reinstall in-place", func(t *testing.T) {
+	t.Run("can override existing binary with --path", func(t *testing.T) {
 		dir := t.TempDir()
 		_, err := runInstallScript(t, "--path", dir)
 		require.NoError(t, err)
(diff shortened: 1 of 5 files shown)


e2e/install_script_posix_test.go
```diff
@@ -68,6 +74,20 @@ func TestInstallScript(t *testing.T) {
...
+		requireWriteDummyExecutable(t, filepath.Join(dir, "topo"))
+		pathWithTopo := dir + string(os.PathListSeparator) + os.Getenv("PATH")
+
+		out, err := runInstallScriptWithEnv(t, []string{"PATH=" + pathWithTopo})
+
+		require.NoError(t, err, "script failed: %s", out)
+		assert.Contains(t, out, "topo is already installed")
+		assert.Contains(t, out, "topo upgrade")
+		assert.NotContains(t, out, "Installing topo")
+		assert.NotContains(t, out, "Installed topo")
```
muchzill4: Fragile and doesn't give much confidence.
```suggestion
```

e2e/install_script_windows_test.go
```diff
@@ -73,6 +78,20 @@ func TestInstallScriptWindows(t *testing.T) {
 		assert.FileExists(t, filepath.Join(dir, "topo.exe"))
 	})
 
+	t.Run("tells user to use upgrade when topo is already installed", func(t *testing.T) {
+		dir := t.TempDir()
+		requireWriteDummyExecutable(t, filepath.Join(dir, "topo.exe"))
+		pathWithTopo := dir + string(os.PathListSeparator) + os.Getenv("PATH")
```
muchzill4:
```suggestion
		topoInstallDir := t.TempDir()
		requireWriteDummyExecutable(t, filepath.Join(topoInstallDir, "topo.exe"))
		pathWithTopo := topoInstallDir + string(os.PathListSeparator) + os.Getenv("PATH")
```

scripts/install.sh
```diff
@@ -109,6 +109,23 @@ build_download_url() {
...
 }
 
+exit_if_topo_already_installed() {
+  requested_install_dir="$1"
+
+  if [ -n "$requested_install_dir" ]; then
+    return
+  fi
+
+  existing="$(command -v "$BINARY_NAME" 2>/dev/null || true)"
```
muchzill4, line 113: Shouldn't this be more clever given `--path`? 

It seems that even if I actually follow the instructions and provide `--path`, it will still detect `topo` that already exists in `$PATH`?

I.e.: Unless I'm mistaken here's what can happen:
```
$ command -v topo
/usr/bin/topo
$ ./install.sh --path /hello/
topo is already installed at /usr/bin/topo
```

awphi: That should be what the `if [ -n "$requested_install_dir" ];` check does just above.

When I run your steps I see the following - do you see something different?

```
topo/scripts on  install-script-topo-upgrade [$?⇡] 
❯ command -v topo            
/Users/adahod01/.local/bin/topo

topo/scripts on  install-script-topo-upgrade [$⇡] 
❯ ./install.sh --path ./hello
Resolving latest version...
Installing topo v5.0.0
Downloading https://artifacts.tools.arm.com/topo/v5.0.0/macos/topo_darwin_arm64.tar.gz...
Installed topo v5.0.0 to ./hello/topo
Run 'topo --help' to get started
```

e2e/install_script_posix_test.go
awphi: https://github.com/arm/topo/pull/245/commits/7d55c8ab95bb30f3847b7c35105d856810eed06c

e2e/install_script_windows_test.go
awphi: https://github.com/arm/topo/pull/245/commits/7a120f2f065018c65cd6da03f0e0e997b47e371f

scripts/install.sh
muchzill4: I just tried and you're right that it works. I admit I was running it in my head. 😅  

I think I got confused by the fact `exit_if_topo_already_installed` also exits if a custom install path is given, but then topo isn't installed. I think I'd move the `if [ -n ]` to `main`.

muchzill4 (approved):

scripts/install.sh
awphi: Mm yes that is a bit of a confusing name 😅  - hoisted the condition into main and inverted! https://github.com/arm/topo/pull/245/commits/982f4b8767502c4e25ec5e871a40d2f1240bd0c4

scripts/install.sh
```diff
@@ -190,11 +192,15 @@ install_binary() {
 main() {
   parse_args "$@"
 
+  if [ -z "$ARG_INSTALL_DIR" ]; then
+    exit_if_topo_already_installed "$ARG_INSTALL_DIR"
```
muchzill4: No need to give it args?

awphi: No need indeed 🤦 Surprised shellcheck didn't catch this https://github.com/arm/topo/pull/245/commits/1095fab1121abdc6d9c5aa641d4ae22fc6aa72a9

PR #246, opened 2026-04-27
fix: Improve confusing tunnel security messaging

muchzill4:
## Changes

- Stop displaying curl failures in the logs, because failing curl is actually good in this case
- Explain the purpose of the check
- Make error/success messages more meaningful and potentially actionable

### DIFF
--- a/internal/ssh/ssh_tunnel.go
+++ b/internal/ssh/ssh_tunnel.go
@@ -122,11 +122,12 @@ func (ct *CheckRemoteForwardNotExposed) Run(w io.Writer) error {
 
 	hostName := NewConfig(ct.TargetDest).HostName
 	cmd := exec.Command("curl", fmt.Sprintf("%s:%s", hostName, ct.Port), "--max-time", "1")
-	cmd.Stdout = w
-	cmd.Stderr = w
+	// curl failing is the success path here; suppress its output so a
+	// "Failed to connect" line doesn't appear in the deployment log.
+	cmd.Stdout = io.Discard
+	cmd.Stderr = io.Discard
 
-	err := cmd.Run()
-	if err == nil {
+	if err := cmd.Run(); err == nil {
 		return fmt.Errorf("remote sshd is exposing the forwarded port %s on its network (likely GatewayPorts=yes); the local registry is reachable without SSH auth", ct.Port)
 	}
 
--- a/internal/ssh/ssh_tunnel.go
+++ b/internal/ssh/ssh_tunnel.go
@@ -104,9 +104,9 @@ type CheckRemoteForwardNotExposed struct {
 	Port       string
 }
 
-// Detects a remote sshd with GatewayPorts enabled, which would bind the -R
-// forward to 0.0.0.0 on the target instead of its loopback, exposing the
-// local registry to the target's network.
+// Checks whether the RemoteForward port exposes the registry to the target's
+// network, rather than being limited to target loopback. This can happen when
+// sshd permits non-loopback remote forwards, such as GatewayPorts.
 func NewCheckRemoteForwardNotExposed(targetDest Destination, port string) *CheckRemoteForwardNotExposed {
 	return &CheckRemoteForwardNotExposed{TargetDest: targetDest, Port: port}
 }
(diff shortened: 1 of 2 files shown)


internal/ssh/ssh_tunnel.go
```diff
@@ -99,41 +99,42 @@ func (s *SSHTunnelStart) Run(w io.Writer) error {
...
+type CheckRemoteForwardNotExposed struct {
 	TargetDest Destination
 	Port       string
 }
 
-func NewCheckSSHTunnelSecurity(targetDest Destination, port string) *CheckSSHTunnelSecurity {
-	return &CheckSSHTunnelSecurity{TargetDest: targetDest, Port: port}
+// Detects a remote sshd with GatewayPorts enabled, which would bind the -R
+// forward to 0.0.0.0 on the target instead of its loopback, exposing the
+// local registry to the target's network.
```
th3james: Potentially a bit clearer
```suggestion
// Checks whether the RemoteForward port exposes the registry to the target's
// network, rather than being limited to target loopback. This can happen when
// sshd permits non-loopback remote forwards, such as GatewayPorts.
```

internal/ssh/ssh_tunnel.go
```diff
@@ -99,41 +99,42 @@ func (s *SSHTunnelStart) Run(w io.Writer) error {
...
 
-	err := cmd.Run()
-	if err == nil {
-		return fmt.Errorf("ssh tunnel to %s is not secure: able to access registry port without authentication", ct.TargetDest)
+	hostName := NewConfig(ct.TargetDest).HostName
+	cmd := exec.Command("curl", fmt.Sprintf("%s:%s", hostName, ct.Port), "--max-time", "1")
+	// curl failing is the success path here; suppress its output so a
+	// "Failed to connect" line doesn't appear in the deployment log.
+	cmd.Stdout = io.Discard
+	cmd.Stderr = io.Discard
```
th3james, line 128: I think this is actually a regression because we're suppressing diagnostic information, but we're not validating that the error is the one we're expecting. According to the curl man page, we can actually use the status codes to be more targeted and only expect status code `7` - "Failed to connect to host". This means host resolution or malformed URLs would be errors.
https://linux.die.net/man/1/curl

th3james (changes requested):

internal/ssh/ssh_tunnel.go
muchzill4: I've suppressed the output, that is correct, however I'd refrain from calling it diagnostic information. For the end user this is meaningless. We could debug log what curl says instead.

But to clarify, I've not introduced a regression, this code was already broken, which is something I'm realising now. Curl will error when trying to poke a protocol it doesn't understand. In case of ssh forwarding it is simply TCP, and trying to curl the port I'm getting `curl: (56) Recv failure: Connection reset by peer`.

Now, I'm not sure if we can rely on status codes, because `curl` on windows is not what you think. Perhaps `curl.exe` behaves like curl with status codes. Will check when I get my hands on a windows box.

yejseo01: so do you think we need to check the port security without using curl?

th3james: Yes, I see your point. The registry uses HTTP, so I believe the curl test should work correctly if the registry is running (@yejseo01 should be able to confirm, she tested it). The next question becomes whether we should be testing pure TCP or coupling the check the registry running (probably just TCP?).

The thing I can't remember is why we're doing a curl test rather than using Go's networking libs - feels like it would be easier to guarantee the portability of that approach.

Let's integrate this PR with just the UX fixes and address the robustness separately.

th3james (approved):

internal/ssh/ssh_tunnel.go
yejseo01: Yes. As long as the registry is running the curl check should be fine.

muchzill4: > The thing I can't remember is why we're doing a curl test rather than using Go's networking libs - feels like it would be easier to guarantee the portability of that approach.

Depending on `CGO_ENABLED` it may or may not be able to lookup the host using the hostname. I learned that yesterday after trying to do a simple tcp dial in golang.

> Yes. As long as the registry is running the curl check should be fine.

As far as I can tell, there's nothing in the code that guarantees registry running before this curl executes. The operations are ordered correctly, but we run registry in the background, so who knows if it's listening or not.
I've tried setting `GatewayPorts yes` on remote and was not able to make this check fail. 🤷

yejseo01: I'm surprised that happened. I thought the ordered operations would be good enough for most of the check to be fine. did you make some sort of artificial delay to hit it?

muchzill4: Can you share how you've been testing this? Perhaps I'm doing something wrong.

PR #247, opened 2026-04-28
ci: follow the rename to topo-publisher

yejseo01:
## Changes

- Follow the rename of topo-publisher
- release will not work until this PR is merged


muchzill4:
The way we have changelog setup, `fix:` is user facing. Is this relevant for the changelog or should we use `ci:` instead?

muchzill4 (approved):

federicobozzini (approved):

PR #250, opened 2026-04-29
fix: accurate `topo --version`

federicobozzini:
## Changes

- topo version made overridable


muchzill4 (approved):
Nice. TIL that ldflags don't override non const!

PR #251, opened 2026-04-30
fix!: improve accuracy of port probing

muchzill4:
## Changes

- Tighten the heuristic used to determine whether a port is open on the remote host.
    - `curl` can succeed whenever it reaches a service speaking any protocol it understands at the given `<url>` such as HTTP, FTP, WS, or SMTP.
    - As a result, the current check can depend on the registry being ready behind the tunnelled port, rather than only checking whether the remote port is reachable.
    - This PR changes the check to base decision on `curl`'s exit code. Exit code `7` means "failed to connect to host", which is the signal we need when verifying that a remote port is not exposed to non-loopback addresses.
- Keep using `curl` because Go's built-in networking may not resolve the hostname in the same environment.


internal/ssh/ssh_tunnel.go
```diff
@@ -120,19 +121,31 @@ func (ct *CheckRemoteForwardNotExposed) Run(w io.Writer) error {
...
+	}
+	return fmt.Errorf("remote sshd might be exposing the forwarded port %s on its network (likely GatewayPorts=yes); the local registry may be reachable without SSH auth", ct.Port)
+}
+
+func IsRemotePortDefinitelyNotListening(host, port string) bool {
+	curl := "curl"
+	if runtime.GOOS == "windows" {
+		curl = "curl.exe"
+	}
+	cmd := exec.Command(curl, fmt.Sprintf("%s:%s", host, port), "--max-time", "5")
```
muchzill4: I had to bump the `--max-time`, because the hostname resolution is slow in CI.
When testing on local network:
- connecting from my Mac to remote linux box, check fails after ~9ms
- connecting from Windows to remote linux box, check fails after ~2000ms

We're probably going to have to watch this space, because it has characteristics of an intermittently failing piece of functionality.

internal/ssh/ssh_tunnel.go
```diff
@@ -120,19 +121,31 @@ func (ct *CheckRemoteForwardNotExposed) Run(w io.Writer) error {
...
 	cmd.Stdout = io.Discard
 	cmd.Stderr = io.Discard
+	_ = cmd.Run()
 
-	if err := cmd.Run(); err == nil {
-		return fmt.Errorf("remote sshd is exposing the forwarded port %s on its network (likely GatewayPorts=yes); the local registry is reachable without SSH auth", ct.Port)
+	// Only curl exit 7 ("Failed to connect to host") proves no TCP listener
+	// answered. Anything else — connection succeeded, DNS failure, timeout,
+	// curl missing — leaves us unable to certify the port is not not listening.
```
awphi, line 140: In these circumstances are we okay with preventing deployment with no escape hatch? 

This puts a hard requirement on `curl` on the host which we didn't have before (probably by coincidence but you know) so at the very least should probably be in the health check.

awphi: Also think this probably makes this a breaking change, though quite niche ofc

internal/ssh/ssh_tunnel_test.go
```diff
@@ -136,6 +136,41 @@ func TestCheckRemoteForwardNotExposed(t *testing.T) {
...
+		require.NoError(t, err)
+		t.Cleanup(func() { _ = listener.Close() })
+		_, port, err := net.SplitHostPort(listener.Addr().String())
+		require.NoError(t, err)
+
+		assert.False(t, ssh.IsRemotePortDefinitelyNotListening("127.0.0.1", port))
+	})
+
+	t.Run("it returns false when curl exits with any other non-7 error", func(t *testing.T) {
+		// `.invalid` is reserved by RFC 6761 to never resolve, so curl
```
awphi, line 157: TIL!

internal/ssh/ssh_tunnel.go
muchzill4: https://github.com/arm/topo/pull/251/commits/c00fed93147f5f5b5181f82dd12fc19e0e9c8a81

awphi (approved):
Hard dependency on curl is unfortunate but at least accurate now. Very unfortunate there's no pure go way to solve this

PR #252, opened 2026-04-30
feat: opinionated default install dir + shell-specific command to add to PATH in install script 

awphi:
## Changes

- Moves away from the "let's find a dir already on the PATH to shove topo into and tell the user to add it if we can't find one" and replaces it with "always install to `~/.local/bin`, then check if its on the PATH and if not give the user a one-liner to do it"
- Minor test changes to:
  - Assert the correct version is downloaded when passing `--version`
  - Shell-specific instructions are included to add to PATH based on the users' default shell
  - Does not prompt to add to PATH if install dir is already on it

Have tested on all the included shells and works for me - not too familiar with fish idioms. The add path utility I've used seemed neat but not sure if this is standard.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -3,7 +3,7 @@
 set -eu
 
 USAGE="POSIX-portable idempotent installer for topo.
-Downloads a release from the Arm artifactory server and installs the application.
+Downloads and installs a topo release from the Arm Artifactory server.
 
 Usage:
   sh install.sh [--version VERSION] [--path DIRECTORY]
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -3,15 +3,14 @@
 set -eu
 
 USAGE="POSIX-portable idempotent installer for topo.
-Downloads a release from the Arm artifactory server and installs the binary
-to \$HOME/.local/bin.
+Downloads a release from the Arm artifactory server and installs the application.
 
 Usage:
   sh install.sh [--version VERSION] [--path DIRECTORY]
 
 Options:
   --version VERSION   Install a specific version (e.g. v4.0.0). Default: latest.
-  --path DIRECTORY    Install the binary into DIRECTORY instead of \$HOME/.local/bin."
+  --path DIRECTORY    Install to a custom directory. Default: \$HOME/.local/bin."
 
 BASE_URL="https://artifacts.tools.arm.com/topo"
 BINARY_NAME="topo"
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -3,15 +3,15 @@
 set -eu
 
 USAGE="POSIX-portable idempotent installer for topo.
-Downloads a release from the Arm artifactory server and places the binary
-on the current user's PATH.
+Downloads a release from the Arm artifactory server and installs the binary
+to \$HOME/.local/bin.
 
 Usage:
   sh install.sh [--version VERSION] [--path DIRECTORY]
 
 Options:
   --version VERSION   Install a specific version (e.g. v4.0.0). Default: latest.
-  --path DIRECTORY    Install the binary into DIRECTORY instead of auto-detecting."
+  --path DIRECTORY    Install the binary into DIRECTORY instead of \$HOME/.local/bin."
 
 BASE_URL="https://artifacts.tools.arm.com/topo"
 BINARY_NAME="topo"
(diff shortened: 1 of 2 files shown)


scripts/install.sh
```diff
@@ -172,17 +174,15 @@ install_binary() {
   version="$3"
 
   install -m 0755 "$src" "${install_dir}/${BINARY_NAME}"
-  echo "Installed ${BINARY_NAME} ${version} to ${install_dir}/${BINARY_NAME}"
+  echo "Downloaded ${BINARY_NAME} ${version} to ${install_dir}/${BINARY_NAME}"
```
muchzill4, line 176: I think `Installed` sounded better.

scripts/install.sh
```diff
@@ -3,15 +3,15 @@
 set -eu
 
 USAGE="POSIX-portable idempotent installer for topo.
-Downloads a release from the Arm artifactory server and places the binary
-on the current user's PATH.
+Downloads a release from the Arm artifactory server and installs the binary
+to \$HOME/.local/bin.
```
muchzill4: Should we skip this and instead say that `--path` default is `$HOME/.local/bin`?

muchzill4 (commented):
Should we mention to the user if `topo` is already installed anywhere?

scripts/install.sh
awphi: My thinking was that I'm not sure I'd consider a random binary on disk that's not on my PATH yet "installed" - WDYT?

scripts/install.sh
awphi: https://github.com/arm/topo/pull/252/commits/da381dd1b0f4d64380a46f7e9e53a853a795e736

awphi:
> Should we mention to the user if topo is already installed anywhere?

This PR doesn't change the existing atomic replacement behaviour i.e. if they run this script with topo installed at `/usr/local/bin/topo` with `/usr/local/bin` on the PATH we'll still attempt to put it there.

Open to changing this behaviour given perms issues we've seen with users who manually installed to privileged dirs prior to the existence of these scripts but not sure what the path forward should be? Bail out and tell them to delete the old install?

Either way sounds like a separate PR to me

scripts/install.sh
```diff
@@ -110,42 +109,44 @@ build_download_url() {
...
 
-  echo "Error: could not find a user-writable directory on PATH." >&2
-  echo "Provide one explicitly with --path, or add $preferred_dir to your PATH with the following command:" >&2
-  echo "  export PATH=\"\$PATH:$preferred_dir\"" >&2
-  echo "To persist across terminal restarts, add the export line to your shell's configuration file (e.g. ~/.bashrc, ~/.zshrc)" >&2
-  exit 1
+path_setup_command() {
+  install_dir="$1"
+
+  case "${SHELL:-}" in
```
awphi, line 136: Uses the default shell rather than the current shell since we expect this script to be executed with `sh`. This seemed to be in line with other install scripts like https://mise.run/

scripts/install.sh
```diff
@@ -110,42 +109,44 @@ build_download_url() {
...
+    */zsh)
+      echo "echo 'export PATH=\"\$PATH:${install_dir}\"' >> ~/.zshrc; export PATH=\"\$PATH:${install_dir}\""
```
awphi, line 138: Couldn't get these more concise - other tools seem to provide a `topo activate zsh` etc. helper that gets piped into `source` and `.<shell>rc`

scripts/install.sh
muchzill4: Ah.. that's fair.

scripts/install.sh
muchzill4: But.. wouldn't they have to be on the path for `<tool> activate` to work? 🤔

awphi: They use the absolute path to the downloaded but not yet installed binary

scripts/install.sh
```diff
@@ -3,15 +3,14 @@
 set -eu
 
 USAGE="POSIX-portable idempotent installer for topo.
-Downloads a release from the Arm artifactory server and places the binary
-on the current user's PATH.
+Downloads a release from the Arm artifactory server and installs the application.
```
muchzill4:
```suggestion
Downloads and installs a topo release from the Arm Artifactory server.
```

muchzill4 (approved):

PR #253, opened 2026-04-30
ci: add prerelease

yejseo01:
## Changes

- Enable pre-release to make e2e testing of release process less hazardous

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -137,7 +137,7 @@ jobs:
         run: |
           gh workflow run sign-and-release.yml \
             --repo Arm-Debug/topo-publisher \
-            --ref use-publisher \
+            --ref v5 \
             --field run-id=${{ github.run_id }} \
             --field release-version=${VERSION} \
             --field release-tag=${VERSION} \
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -2,6 +2,12 @@ name: Release
 
 on:
   workflow_dispatch:
+    inputs:
+      prerelease:
+        description: 'Create a pre-release candidate'
+        required: true
+        type: boolean
+        default: false
 
 permissions:
   contents: write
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -29,11 +35,20 @@ jobs:
...
-          if ! next="$(svu next)"; then
-            echo "svu failed" >&2
-            exit 1
+          if [[ "$PRERELEASE" == "true" ]]; then
+            current="$(svu current)"
+            if [[ "$current" == *-rc.* ]]; then
+              next="$(svu prerelease --prerelease rc)"
+            else
+              next="$(svu next --prerelease rc.1)"
+            fi
```
federicobozzini, line 47: Is this actually necessary or can we just do this?

```suggestion
            next="$(svu next --prerelease rc)"
```

.github/workflows/release.yaml
```diff
@@ -29,11 +35,20 @@ jobs:
...
         run: |
-          if ! next="$(svu next)"; then
-            echo "svu failed" >&2
-            exit 1
+          if [[ "$PRERELEASE" == "true" ]]; then
+            current="$(svu current)"
+            if [[ "$current" == *-rc.* ]]; then
+              next="$(svu prerelease --prerelease rc)"
+            else
+              next="$(svu next --prerelease rc.1)"
```
awphi: I think this misuses the term "release candidate" (assuming that's what is meant by rc) - these are ready to use binaries we're considering calling vX.Y.Z but really a short-lived intermediate step to producing a binary we're definitely calling vX.Y.Z. 

Maybe `unsigned` instead? Still think it's confusing to put these artifacts in the publicly documented and consumed topo release artifactory

awphi (commented):
As @Luke-Parkin pointed out, we need to be mindful of `topo upgrade` and the release scripts which have a dependency on the https://artifacts.tools.arm.com/devx-topo/ index to work out what the latest version is.

Currently tags on the end of version numbers are stripped so while this change wouldn't result in someone downloading an unsigned binary by accident it would break these upgrade mechanisms while a `-rc` release is the latest.

I guess we need to adjust the upgrade mechanisms to account for this but we need to wait for users to upgrade to topo version that has these changes. Alternatively (my preference) - upload these "release candidates" into a separate folder/repo on artifactory

PR #259, opened 2026-05-07
refactor: manually calculate image names instead of using `docker compose`

muchzill4:
## Changes

- Calculate image names using data from compose file instead of relying on `docker compose config --images`
   - Enables support for container engines that follow different image name semantics

### DIFF
--- a/internal/compose/compose_test.go
+++ b/internal/compose/compose_test.go
@@ -2,6 +2,7 @@ package compose_test
 
 import (
 	"path/filepath"
+	"sort"
 	"testing"
 
 	"github.com/arm/topo/internal/compose"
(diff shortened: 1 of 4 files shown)


internal/compose/compose_test.go
```diff
@@ -188,6 +189,106 @@ func TestRegisterVolumes(t *testing.T) {
...
+      service: build-base
+`)
+
+		got, err := compose.ImageNames(path)
+
+		require.NoError(t, err)
+		assert.Equal(t, []string{"duff:latest", "springfield-api"}, got)
+	})
+
+	t.Run("returns sorted output", func(t *testing.T) {
```
awphi, line 260: Seems to overlap with the first test "returns explicit and generated image names in sorted order"

muchzill4: D'oh! I wanted an explicit sort order test and failed to fixup the rest - thanks for spotting this: https://github.com/arm/topo/pull/259/commits/cf1b992de9c1ac6caccfa74c993d51979ec9117d

awphi (approved):
Sorry have been context switching way too much and completely forgot I promised to re-review. LGTM!

PR #261, opened 2026-05-07
feat!: add explicit command to fix target health issues

federicobozzini:
## Changes

- added new `fixCommand` property to the output result of the health check. This property doesn't appear in plain output, only json.

### DIFF
--- a/e2e/setup_keys_test.go
+++ b/e2e/setup_keys_test.go
@@ -21,11 +21,13 @@ func TestSetupKeysJourney(t *testing.T) {
 	out := runTopo(t, topo, "health", "--target", container.SSHDestination)
 	assert.Contains(t, out, "Connectivity: ❌ (ssh host key is unknown)")
 	assert.Contains(t, out, "Trust the target's SSH host key")
+	assert.Contains(t, out, fmt.Sprintf("Cmd: topo health --target %s --accept-new-host-keys", container.SSHDestination))
 
 	Step(t, "health with accept-new-host-keys trusts host and suggests setup-keys")
 	out = runTopo(t, topo, "health", "--target", container.SSHDestination, "--accept-new-host-keys")
 	assert.Contains(t, out, "Connectivity: ❌ (ssh authentication failed)")
 	assert.Contains(t, out, "Configure SSH keys on remote target")
+	assert.Contains(t, out, fmt.Sprintf("Cmd: topo setup-keys --target %s", container.SSHDestination))
 
 	Step(t, "setup-keys generates keys and installs them on the target")
 	askpass := writeAskPassScript(t, sshRootPassword)
--- a/e2e/setup_keys_test.go
+++ b/e2e/setup_keys_test.go
@@ -20,14 +20,12 @@ func TestSetupKeysJourney(t *testing.T) {
 	Step(t, "health reports unknown host key and suggests accept-new-host-keys")
 	out := runTopo(t, topo, "health", "--target", container.SSHDestination)
 	assert.Contains(t, out, "Connectivity: ❌ (ssh host key is unknown)")
-	wantFix := fmt.Sprintf("run `topo health --target %s --accept-new-host-keys", container.SSHDestination)
-	assert.Contains(t, out, wantFix)
+	assert.Contains(t, out, "Trust the target's SSH host key")
 
 	Step(t, "health with accept-new-host-keys trusts host and suggests setup-keys")
 	out = runTopo(t, topo, "health", "--target", container.SSHDestination, "--accept-new-host-keys")
 	assert.Contains(t, out, "Connectivity: ❌ (ssh authentication failed)")
-	wantFix = fmt.Sprintf("run `topo setup-keys --target %s`", container.SSHDestination)
-	assert.Contains(t, out, wantFix)
+	assert.Contains(t, out, "Configure SSH keys on remote target")
 
 	Step(t, "setup-keys generates keys and installs them on the target")
 	askpass := writeAskPassScript(t, sshRootPassword)
--- a/e2e/setup_keys_test.go
+++ b/e2e/setup_keys_test.go
@@ -20,13 +20,13 @@ func TestSetupKeysJourney(t *testing.T) {
 	Step(t, "health reports unknown host key and suggests accept-new-host-keys")
 	out := runTopo(t, topo, "health", "--target", container.SSHDestination)
 	assert.Contains(t, out, "Connectivity: ❌ (ssh host key is unknown)")
-	assert.Contains(t, out, "Trust the target's SSH host key")
+	assert.Contains(t, out, "Fix: Trust the target's SSH host key")
 	assert.Contains(t, out, fmt.Sprintf("Cmd: topo health --target %s --accept-new-host-keys", container.SSHDestination))
 
 	Step(t, "health with accept-new-host-keys trusts host and suggests setup-keys")
 	out = runTopo(t, topo, "health", "--target", container.SSHDestination, "--accept-new-host-keys")
 	assert.Contains(t, out, "Connectivity: ❌ (ssh authentication failed)")
-	assert.Contains(t, out, "Configure SSH keys on remote target")
+	assert.Contains(t, out, "Fix: Configure SSH keys on remote target")
 	assert.Contains(t, out, fmt.Sprintf("Cmd: topo setup-keys --target %s", container.SSHDestination))
 
 	Step(t, "setup-keys generates keys and installs them on the target")
--- a/internal/health/dependencies.go
+++ b/internal/health/dependencies.go
@@ -52,6 +52,7 @@ var HostRequiredDependencies = []Dependency{
 			},
 			CurrentVersion: version.Version,
 			Fix:            "run `topo upgrade`",
+			FixCommand:     "topo upgrade",
 		}},
 	},
 	{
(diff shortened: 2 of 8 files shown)


muchzill4 (changes requested):
There's a lot of duplication between `Fix` and `FixCommand`, can you think of a way where we could use `FixCommand` as the source of truth for the command, and `Fix` only for the _meta_ text?

Something like (pseudo-template):

```
{{Issue}}
   {{Fix}}
   {{If FixCommand}}
     -> run `{{FixCommand}}`
```

federicobozzini:
> There's a lot of duplication between `Fix` and `FixCommand`, can you think of a way where we could use `FixCommand` as the source of truth for the command, and `Fix` only for the _meta_ text?

This was also my first thought but:
- often fixes are more than just `run '<fixCommand>'`.
- some issues don't have a `fixCommand` associated, `fix` is just a generic suggestion.

I think this is one of those cases where removing dupilications makes things messier rather than cleaner.

internal/health/dependencies.go
```diff
@@ -168,6 +173,9 @@ func PerformChecks(ctx context.Context, dependencies []Dependency, runner runner
 			}
 
 			if err != nil {
+				if provider, ok := check.(fixCommandProvider); ok {
```
muchzill4: This is quite horrible. 😓 If you want to have a fix that might have additional data (a specific command), it would be much suitable as part of a custom `Fix` structure which would be returned by the `check.Run()`.

federicobozzini: Is this what you had in mind? https://github.com/arm/topo/pull/261/commits/efc9706d77ef88897ef12fa82423029800b99d63

I would agree it looks cleaner.

internal/health/health.go
```diff
@@ -184,7 +188,8 @@ func generateDependencyReport(statuses []DependencyStatus) []HealthCheck {
 				hc.Status = CheckStatusError
 			}
 			hc.Value = ds.Error.Error()
-			hc.Fix = ds.Fix
+			hc.Fix = ds.Fix.Text
```
federicobozzini: This might not be the best but it would avoid a breaking change. I also think `Fix` is a good name so I would personally keep it like this.

internal/health/dependencies.go
muchzill4: It does, but I think we should propagate this to `HealthCheck` too. And again, I stand my ground about duplicating command and text - let's not do it that way.

I'd have:
```
Fix{Text|Description: "Configure SSH keys on remote target", Command: "topo setup-keys"}
```

That way you have a description of the fix and an optional copy-pasteable command user can use. We'd need to tweak the template to consume this data nicely.

federicobozzini: In this case, do you think it would be more reasonable for the json output to be:

```json
  "fix": "Configure SSH keys on remote target",
  "fixCommand": "topo health --target ssh://stm32-yejin --accept-new-host-keys"
```

or:

```json
  "fix": {
    "description": "Configure SSH keys on remote target",
    "command": "topo health --target ssh://stm32-yejin --accept-new-host-keys"
  }
```

I'm leaning more towards the second option even if technically it would be a breaking change.

muchzill4: I think second option is indeed better. 👍 

- I started to wonder if "description" works for command-less fixes, but I guess it's ok if we treat it as the description of the fix, not the command. I.e.: in one instance we'll have:

    ```json
    "fix": {
       "description": "Ensure current user can run docker commands"
    }
    ```

    Does that look good? Perhaps it does. 🤔 

- I wonder how we make it look nice in the CLI, perhaps:

    ```
    Target
    ------
    Connectivity: ❌ (ssh authentication failed)
      Fix: Configure SSH keys on remote target
      Cmd: topo setup-keys --target ssh://host
    ```

---

I'm quite excited by this change, it's going to be great to have a separate field for the command to execute.

federicobozzini: Some example of how it looks now:

```
topo health --target xxx
Host
----
Topo: ✅ (topo)
SSH: ✅ (ssh)
Curl: ✅ (curl)
Container Engine: ✅ (docker)

Target
------
Connectivity: ✅
Container Engine: ✅ (docker)
Remoteproc Runtime: ⚠️ ("remoteproc-runtime" not found on remote target's $PATH)
  Fix: Install the remoteproc runtime
  Cmd: topo install remoteproc-runtime
Remoteproc Shim: ⚠️ ("containerd-shim-remoteproc-v1" not found on remote target's $PATH)
  Fix: Install the remoteproc runtime
  Cmd: topo install remoteproc-runtime
Hardware Info: ✅ (lscpu)
```

```
topo health --target yyy
Host
----
Topo: ✅ (topo)
SSH: ✅ (ssh)
Curl: ✅ (curl)
Container Engine: ✅ (docker)

Target
------
Connectivity: ❌ (ssh host key is unknown)
  Fix: Trust the target's SSH host key
  Cmd: topo health --target yyy --accept-new-host-keys
```

internal/health/dependencies_test.go
```diff
@@ -135,7 +135,30 @@ func TestPerformChecks(t *testing.T) {
...
+				},
+			},
+		}
+		runner := &runner.Fake{}
+
+		got := health.PerformChecks(context.Background(), []health.Dependency{dep}, runner)
+
+		assert.Len(t, got, 1)
+		assert.Equal(t, "topo install remoteproc-runtime", got[0].Fix.Command)
+		assert.Equal(t, "Install the remoteproc runtime", got[0].Fix.Description)
```
muchzill4: I think a single `assert.Equal` instead of comparing field by field is a bit nicer, also tests fail better when you add fields.

```go
want := Fix{...}
assert.Equal(t, want, got[0].Fix)
```

internal/health/health_test.go
```diff
@@ -130,6 +133,38 @@ func TestHostReport(t *testing.T) {
 			want := `{ "dependencies": [] }`
 			assert.JSONEq(t, want, string(b))
 		})
+
+		t.Run("includes command when fix is set", func(t *testing.T) {
```
muchzill4: Isn't this essentially testing that go json marshaller works? I think a more interesting case is `omitempty`, which is not standard(?).

Should we also do `omitempty` for `Command`? 🤔

internal/output/templates/health_test.go
```diff
@@ -151,7 +158,10 @@ func TestPrintHealthReport(t *testing.T) {
 			err := printable.Print(toPrint, &out, term.Plain)
 
 			require.NoError(t, err)
-			assert.Contains(t, out.String(), "Apply Working Hands Cream")
+			assert.Contains(t, out.String(), "Skin Care: ⚠️")
+			assert.Contains(t, out.String(), "  Fix: Apply Working Hands Cream")
+			assert.Contains(t, out.String(), "  Cmd: topo moisturise")
+			assert.NotContains(t, out.String(), "→")
```
muchzill4: Why do this?

e2e/setup_keys_test.go
```diff
@@ -20,14 +20,14 @@ func TestSetupKeysJourney(t *testing.T) {
 	Step(t, "health reports unknown host key and suggests accept-new-host-keys")
 	out := runTopo(t, topo, "health", "--target", container.SSHDestination)
 	assert.Contains(t, out, "Connectivity: ❌ (ssh host key is unknown)")
-	wantFix := fmt.Sprintf("run `topo health --target %s --accept-new-host-keys", container.SSHDestination)
-	assert.Contains(t, out, wantFix)
+	assert.Contains(t, out, "Trust the target's SSH host key")
+	assert.Contains(t, out, fmt.Sprintf("Cmd: topo health --target %s --accept-new-host-keys", container.SSHDestination))
```
muchzill4, line 24: Should we be consistent with our assertions? If we assert "Cmd: ...", we should also assert "Fix: ....". Alternatively drop the prefixes altogether.

internal/health/health.go
```diff
@@ -159,11 +159,23 @@ func connectivityCheck(status ConnectionStatus) HealthCheck {
 	check.Value = status.Error.Error()
 	switch {
 	case errors.Is(status.Error, probe.ErrAuthFailed):
-		check.Fix = fmt.Sprintf("run `topo setup-keys --target %s` to configure ssh keys", status.Destination)
+		command := fmt.Sprintf("topo setup-keys --target %s", status.Destination)
+		check.Fix = &Fix{
+			Description: "Configure SSH keys on remote target",
+			Command:     command,
+		}
```
muchzill4:
```suggestion
		check.Fix = &Fix{
			Description: "Configure SSH keys on remote target",
			Command: fmt.Sprintf("topo setup-keys --target %s", status.Destination),
		}
```

Perhaps we can inline the variable in each `case`?

muchzill4 (commented):
Thanks for introducing the nilable `Fix` and tweaking descriptions all over the shop. This is going to be much better than what we had. 🦾

e2e/setup_keys_test.go
federicobozzini: https://github.com/arm/topo/pull/261/commits/77111c226632076a17e811060e0c21da95ca6886

internal/health/health.go
federicobozzini: https://github.com/arm/topo/pull/261/commits/c488e37a258c8821f1fd71c6a0b54b0080e7d152

internal/output/templates/health_test.go
federicobozzini: I added to check myself, but should have not been committed. https://github.com/arm/topo/pull/261/commits/855f1895b16beb1275dee41c7aabb27e5a74812a

internal/health/dependencies_test.go
federicobozzini: https://github.com/arm/topo/pull/261/commits/f1a80a310af3c764095707adfed23ca434b74bbd

internal/health/health_test.go
federicobozzini: https://github.com/arm/topo/pull/261/commits/56d00506a9cd8081e047768d0807d535dd88bc98

> Should we also do omitempty for Command? 🤔

I'm not sure I get what you mean. We already do `omitempty` for it, don't we?

muchzill4 (approved):
Needs marking as breaking change and LGTM. Nice work.

PR #262, opened 2026-05-07
feat: add destination to health output

awphi:
## Changes

- Short n sweet; adds `Destination` (plain) and `destination` (json) to topo health output:

```
...

Target
------
Destination: ssh://root@xyz:4040
...
```

Thought about putting next to "Target" as the section header but felt weird + we need to start measuring how long to make the hyphen border beneath so went with the easiest possible change for now. JSON output is hopefully uncontentious.


internal/health/health_test.go
```diff
@@ -117,6 +117,18 @@ func TestGenerateTargetReport(t *testing.T) {
 		assert.Equal(t, health.CheckStatusError, got.Connectivity.Status)
 		assert.Equal(t, "run `ssh-keygen -R my-target` to remove the old host key, then retry", got.Connectivity.Fix)
 	})
+
+	t.Run("includes destination when specified", func(t *testing.T) {
```
awphi: This test checks very little actual logic but I can already foresee the "no new code without tests" sign being tapped. Point is if any reviewers suggest removing it I wouldn't mind 👀

e2e/health_test.go
```diff
@@ -41,8 +42,10 @@ func TestHealthCheck(t *testing.T) {
 	t.Run("outputs JSON when specified", func(t *testing.T) {
 		out, err := runCheckHealth(topo, container, "--output", "json")
 
+		normalizedOut := strings.ReplaceAll(out, container.SSHDestination, "ssh://root@localhost:<port>")
```
awphi: Not in love with this but alternatives I could think of seemed uglier (i.e. aliasing the test container ssh string or adding wildcard support to golden files)

internal/health/health_test.go
Luke-Parkin: I'd argue that the e2e test is probably enough here.

e2e/health_test.go
muchzill4: This is a nice approach, but it does suffer from a problem: what if we hardcoded `ssh://root@localhost:<port>` in the template? I think we should do a pre-flight check to see if `container.SSHDestination` is actually there (once?), then do the normalisation.

 Another thing that's a bit confusing is `<port>`. Why is it enclosed in `<>`? Perhaps we replace it with: `<SSH_DESTINATION>` or such, it makes it clearer in the golden file that the entire thing is "templated"?

Now that I said "templated", it's giving me more ideas.. 😉 We could have a template for the golden file and render it with `container.SSHDestination`, but that's probably a bit much.

awphi: >  I think we should do a pre-flight check to see if container.SSHDestination is actually there (once?), then do the normalisation.

That is already there :)

---

I've realised a flaw in this approach though which is that it breaks the automatic updating of golden files with `UPDATE_GOLDEN` which I'm not happy about. I've inverted it so the golden file is the thing we normalize as opposed to the command output. https://github.com/arm/topo/pull/262/commits/37aefc8afdf6a5ccd034a2cc021346b468df6400

internal/health/health_test.go
awphi: https://github.com/arm/topo/pull/262/commits/a4d409f611f4e512d53885c6a99da3dbff426fc4

e2e/health_test.go
muchzill4: What if:

```go
testutil.AssertJsonGoldenFile(t, out, "testdata/TestHealthCheckJson.golden", map[string]any{                                                                                                          
      "target": map[string]any{                                                                                                                                                                          
         "destination": "<dynamic>",                                                                                                                                                                     
      },                                                                                                                                                                                                 
})  
```

Sorry for potential double comment, GitHub UI is being unhelpful today.

awphi: Is this what you had in mind? Avoided it originally since the deep merge util seemed like overhead but the API for the actual tests is definitely nicer https://github.com/arm/topo/pull/262/commits/bb5d4e278574a285edb3db8a79ce776420a59aa3

muchzill4 (approved):
👍

PR #263, opened 2026-05-07
ci: Pre release tag management

Luke-Parkin:
## Changes

This adds support for pre-release tags into the workload. Resulting in this behaviour:

- Make a Normal release: `v1.3.0`
- Make a first pre-release: `v1.4.0-pre.1`
- Make Second pre-release: `v1.4.0-pre.2`
- Make a normal release: `v1.4.0`
- Make Second pre-release: `v1.5.0-pre.1`

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -36,10 +36,44 @@ jobs:
       - name: Calculate next version
         id: calculate_next_version
         run: |
-          if ! next="$(svu next)"; then
+          open_pre_base="$(
+            git tag --list 'v*-pre.*' |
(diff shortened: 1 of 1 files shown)


yejseo01:
I fail to understand how this would work better for `topo upgrade` compared to this PR - https://github.com/arm/topo/pull/253

Can I get an explanation?

yejseo01:
How can I make a pre-release with this patch? the release flow doesn't take a pre-release option?

Luke-Parkin:
> How can I make a pre-release with this patch? the release flow doesn't take a pre-release option?

I can't paste a link, but look at lines 4/5/6 of that file. Pre-release is an input
image

(should probably have a better text but it's there!)

yejseo01 (approved):
LGTM

PR #265, opened 2026-05-08
fix: topo install command needs to include target

federicobozzini:
## Changes

- added `--target` flag to the `topo install` fix command

###Before
```
$ topo health --target xxx
Host
----
Topo: ✅ (topo)
SSH: ✅ (ssh)
Curl: ✅ (curl)
Container Engine: ✅ (docker)

Target
------
Destination: ssh://xxx
Connectivity: ✅
Container Engine: ✅ (docker)
Remoteproc Runtime: ⚠️ ("remoteproc-runtime" not found on remote target's $PATH)
  Fix: Install the remoteproc runtime
  Cmd: topo install remoteproc-runtime
Remoteproc Shim: ⚠️ ("containerd-shim-remoteproc-v1" not found on remote target's $PATH)
  Fix: Install the remoteproc runtime
  Cmd: topo install remoteproc-runtime
Hardware Info: ✅ (lscpu)
```

###After:
```
$ topo health --target xxx
Host
----
Topo: ✅ (topo)
SSH: ✅ (ssh)
Curl: ✅ (curl)
Container Engine: ✅ (docker)

Target
------
Destination: ssh://xxx
Connectivity: ✅
Container Engine: ✅ (docker)
Remoteproc Runtime: ⚠️ ("remoteproc-runtime" not found on remote target's $PATH)
  Fix: Install the remoteproc runtime
  Cmd: topo install remoteproc-runtime --target ssh://xxx
Remoteproc Shim: ⚠️ ("containerd-shim-remoteproc-v1" not found on remote target's $PATH)
  Fix: Install the remoteproc runtime
  Cmd: topo install remoteproc-runtime --target ssh://xxx
Hardware Info: ✅ (lscpu)
```

### DIFF
--- a/internal/health/dependencies_test.go
+++ b/internal/health/dependencies_test.go
@@ -8,6 +8,7 @@ import (
 	"github.com/arm/topo/internal/command"
 	"github.com/arm/topo/internal/health"
 	"github.com/arm/topo/internal/runner"
+	"github.com/arm/topo/internal/ssh"
 	"github.com/stretchr/testify/assert"
 )
 
@@ -19,7 +20,7 @@ func TestDependencyFormat(t *testing.T) {
 	})
 
 	t.Run("target dependencies are of the correct format", func(t *testing.T) {
-		for _, dep := range health.TargetRequiredDependencies {
+		for _, dep := range health.TargetRequiredDependencies(ssh.NewDestination("user@my-target")) {
 			assert.NoError(t, command.ValidateBinaryName(dep.Binary))
 		}
 	})
@@ -29,7 +30,7 @@ func TestDependencyFormat(t *testing.T) {
 		seenEnums := make(map[health.SoftwareDependency]string)
 
 		t.Run("There are no duplicate SoftwareEnumID assignments", func(t *testing.T) {
-			for _, dep := range health.TargetRequiredDependencies {
+			for _, dep := range health.TargetRequiredDependencies(ssh.NewDestination("user@my-target")) {
 				if dep.SoftwareEnumID != health.UnsetSoftwareDependency {
 					if existingDep, exists := seenEnums[dep.SoftwareEnumID]; exists {
 						t.Errorf("Duplicate SoftwareEnumID %d assigned to both %q and %q", dep.SoftwareEnumID, existingDep, dep.Binary)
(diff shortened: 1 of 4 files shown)


internal/health/dependencies_test.go
```diff
@@ -50,6 +51,23 @@ func TestDependencyFormat(t *testing.T) {
 	})
 }
 
+func TestTargetRequiredDependencies(t *testing.T) {
+	t.Run("remoteproc install fix command includes the target", func(t *testing.T) {
+		deps := health.TargetRequiredDependencies(ssh.NewDestination("user@my-target"))
+
+		var got string
+		for _, dep := range deps {
```
muchzill4: It's almost as if this test was screaming at us: don't do it this way.

I have some reservations about making `TargetDependencies` a function for the sake of making a single fix contain the `--target`. I'm not sure if this is the best way to model this. Did you explore any other way to do this? 🤔

federicobozzini: Another idea I had was to change [generateDependencyReport](https://github.com/arm/topo/blob/75dabca6f00f00fb82cd168611d48b9cd4286098/internal/health/health.go#L146) but the result is awful. I definitely prefer this approach. I think it allows decently with how [connectivityCheck](https://github.com/arm/topo/blob/75dabca6f00f00fb82cd168611d48b9cd4286098/internal/health/health.go#L126) works.

Luke-Parkin (approved):

internal/health/dependencies_test.go
```diff
@@ -19,7 +20,7 @@ func TestDependencyFormat(t *testing.T) {
 	})
 
 	t.Run("target dependencies are of the correct format", func(t *testing.T) {
-		for _, dep := range health.TargetRequiredDependencies {
+		for _, dep := range health.TargetRequiredDependencies(ssh.NewDestination("user@my-target")) {
```
muchzill4:
```suggestion
		for _, dep := range health.TargetRequiredDependencies(ssh.NewDestination("does-not-matter-for-this-test")) {
```

I dislike the fact that to fulfil the api we now have to pass this test-irrelevant argument. Perhaps changing the value to "does-not-matter-for-this-test" would make it more explicit to not look at it.

internal/health/dependencies_test.go
muchzill4: Hm. Perhaps it's ok. I don't have a better idea.

We should however try change the testing strategy. It loops without breaking and assumes check index.

Two options that might work:
- `assert.Contains()`, which would assume the body of the entire remoteproc-runtime check - easiest
- test helpers:
    - `dep, err := findDependencyByBinary(deps, "remoteproc-runtime") (Dependency, error)`
    - `asset.Contains(t, wantBinaryExistsCheck, dep.Checks)`

internal/health/dependencies_test.go
federicobozzini: https://github.com/arm/topo/pull/265/commits/dd9318a97848f0358ecc456e12aac6fade5f8d5a

internal/health/dependencies_test.go
federicobozzini: I think second approach is cleaner https://github.com/arm/topo/pull/265/commits/dd9318a97848f0358ecc456e12aac6fade5f8d5a

muchzill4 (approved):

PR #266, opened 2026-05-18
feat: Support Homebrew managed installations of Topo

muchzill4:
#### `topo upgrade` changes

- Hide `upgrade` sub-command when we detect `topo` is managed by brew
    - It is still invokable, so that muscle memory (if any) doesn't error with "unknown sub-command"
- If `upgrade` is invoked, refuse to upgrade homebrew if it's managed by brew

#### `scripts/install.sh` changes

- Refuse to install if `--path` looks homebrew-like


cmd/topo/upgrade.go
```diff
@@ -28,7 +28,9 @@ var upgradeCmd = &cobra.Command{
 		defer cancel()
 
 		newVersion, err := upgrade.Upgrade(ctx, spinner)
-		spinner.Stop()
+		if spinner != nil {
+			spinner.Stop()
+		}
```
awphi, line 33: Yikes good spot

internal/upgrade/upgrade.go
```diff
@@ -17,6 +17,14 @@ import (
 )
 
 func Upgrade(ctx context.Context, reporter term.ProgressReporter) (string, error) {
+	binPath, err := CurrentBinaryPath()
+	if err != nil {
+		return "", fmt.Errorf("failed to determine current binary path: %w", err)
+	}
+	if IsTopoBinaryManagedExternally(binPath) {
+		return "", fmt.Errorf("topo was installed by Homebrew; upgrade it with: brew upgrade topo")
+	}
```
awphi: I don't like mentioning homebrew in the error message when the check is just called `IsTopoBinaryManagedExternally`. WDYT about 

```go
func GetExternallyManagedUpgradeCommand(binPath string) string {
	if isHomebrewManagedBinary(binPath) {
		return "brew upgrade topo"
	}

	return nil
}

// potentially don't even need this
func IsTopoBinaryManagedExternally(binPath string) bool {
	return GetExternallyManagedUpgradeCommand(binPath) != nil
}
```

and using the former here instead like:

```go
externalUpgradeCommand := GetExternallyManagedUpgradeCommand(binPath)
if externalUpgradeCommand != nil {
  return "", fmt.Errorf("topo was installed by an external package manager; upgrade it with: %s", GetExternallyManagedUpgradeCommand(binPath))
}
```

scripts/install.sh
```diff
@@ -110,6 +110,11 @@ build_download_url() {
 
 resolve_install_dir() {
   install_dir="${1:-$HOME/.local/bin}"
+  if is_homebrew_managed_dir "$install_dir"; then
+    echo "Error: ${install_dir} is managed by Homebrew" >&2
+    echo "Install with Homebrew instead, or choose a non-Homebrew path such as \$HOME/.local/bin." >&2
```
awphi, line 119:
```suggestion
    echo "Upgrade with Homebrew instead, or choose a non-Homebrew path such as \$HOME/.local/bin." >&2
```

also maybe worth putting a copy-pasta command to brew upgrade in here?

awphi (commented):
One other missing piece is that the version check that exists on `topo health` (dependencies.go) will currently suggest users run `topo upgrade` when out of date which is now not always accurate.

scripts/install.sh
muchzill4: Do we know if it's an upgrade path or install path though? 🤔 Perhaps KISS?

scripts/install.sh
```diff
@@ -121,6 +126,24 @@ resolve_install_dir() {
...
 
+is_homebrew_managed_dir() {
+  install_dir="$1"
+
+  case "$install_dir" in
+    /opt/homebrew|/opt/homebrew/*|/home/linuxbrew/.linuxbrew|/home/linuxbrew/.linuxbrew/*|/usr/local/Cellar|/usr/local/Cellar/*|/usr/local/Homebrew|/usr/local/Homebrew/*)
+      return 0 ;;
+  esac
+
+  case "$(uname -s)" in
```
muchzill4: Note to self: this is a bit weird. Should we not allow installing to `/usr/local/bin` if user desires to do so?

internal/upgrade/command.go
```diff
@@ -0,0 +1,25 @@
...
+	case IsBinaryManagedByHomebrew(binPath):
+		return false, "brew upgrade topo"
+	default:
+		return true, "topo upgrade"
+	}
+}
+
+func IsBinaryManagedByUs(binPath string) bool {
+	isManagedByUs, _ := GetUpgradeCommand(binPath)
+	return isManagedByUs
```
th3james, line 19: ?
```suggestion
	return !IsBinaryManagedByHomebrew(binPath)
```

scripts/install.sh
th3james: Yeah, it feels like we shouldn't do this?

th3james (approved):

internal/upgrade/command.go
awphi: Disagree - it's a little awkward for sure but means when we inevitably add scoop, apt, yum, or whatever support it only needs to be done in one place; `GetUpgradeCommand`

awphi (approved):

internal/upgrade/command.go
muchzill4: This would work, but I wanted it to be "self documenting". I.e.: reading `IsBinaryManagedByUs` gives you an example of `GetUpgradeCommand` use-case. This also future proofs single entry point (YAGNI?) for adding more package managers in the future.

awphi (approved):

PR #267, opened 2026-05-18
fix: explicitly bind tunnel to ipv4 loopback on target

awphi:
## Changes

- Explicitly binds to ipv4 loopback (127.0.0.1:PORT) on the target side when creating the reverse tunnel. On some systems, just specifying the port will fallback to binding it to the ipv6 loopback if ipv4 one is taken silently. This is particularly annoying if you happen to be deploying to a target which is also running a topo registry (e.g. our DGX spark) as it means the deploy continues until the pull step where it fails since it will pull from the on-target registry running on the ipv4 loopback, not your local one available on the ipv6 loopback. 🙄 
- Removes the `isPortTaken` check when ssh tunnel creation fails - since the tunnel is always spawned after the registry is created on the host machine, the port will always be taken so the check is misleading and overly aggressive. Instead replaces it with some guidance in the generic error message.
- Removes some low value tests which were mainly built for the `isPortTaken` check

These changes together would've saved me an hour or two or debugging this afternoon.

### DIFF
--- a/internal/ssh/ssh_tunnel.go
+++ b/internal/ssh/ssh_tunnel.go
@@ -4,27 +4,16 @@ import (
 	"crypto/sha256"
 	"fmt"
 	"io"
-	"net"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"runtime"
-	"time"
 
 	"github.com/arm/topo/internal/command"
 	"github.com/arm/topo/internal/operation"
 )
 
 const TunnelPIDPlaceholder = "<ssh tunnel pid>"
-
-func isPortTaken(port string) bool {
-	conn, err := net.DialTimeout("tcp", fmt.Sprintf("127.0.0.1:%s", port), 2*time.Second)
-	if err != nil {
-		return false
-	}
-	conn.Close() // nolint:errcheck
-	return true
-}
 
 func ControlSocketPath(targetHost string) string {
 	hash := sha256.Sum256([]byte(targetHost))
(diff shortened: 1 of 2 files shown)


internal/ssh/ssh_tunnel.go
```diff
@@ -86,12 +75,8 @@ func (s *SSHTunnelStart) Run(w io.Writer) error {
 		run = cmd.Run
 	}
 	if err := run(); err != nil {
-		if isPortTaken(s.Port) {
-			return fmt.Errorf("port already in use: %s - specify a different port or stop the existing process", s.Port)
-		}
-
 		formattedError := command.FormatError(cmd.Args, err)
-		return fmt.Errorf("failed to open ssh tunnel: %w", formattedError)
+		return fmt.Errorf("failed to open ssh tunnel: %w - ensure port %s is free or specify a different one with `--registry-port`", formattedError, s.Port)
```
muchzill4, line 79: Shouldn't we parse the error message we're getting? Right now, we're going to suggest `--registry-port` for any failure of the `ssh` command, unsure if that's valid. `isPortTaken` made it more valid. 🤔

muchzill4 (approved):

internal/ssh/ssh_tunnel.go
awphi: `isPortTaken` was fairly misleading as outlined in the PR description since the port would always be taken on the host since we open the ssh tunnel after starting the registry.

What we really want to check is if the port is taken on the target but I can't see a way to do this without adding a dependency on the target like `nc` or `lsof` or whatever which I'd like to avoid. 

We could potentially parse the stdout of the tunnel opening command? S[eems "for listen port" would be the string we'd be searching for ](https://github.com/openssh/openssh-portable/blob/cf6c0b3b94cdc223f1b8be1ef2d93e993af5d976/ssh.c#L1965) but it could fail to bind to the port for other reasons I suppose

Luke-Parkin (approved):

PR #268, opened 2026-05-18
feat: topo ps

Luke-Parkin:
## Changes

- Add `topo ps` command.
- Topo PS returns useful information about containers deployed from that project. 
- It also remaps 0.0.0.0 to the correct domain or IP (if provided via target), or resolves ssh config to the ip address, so users can quickly connect to their web apps from their hosts!
```
~/repos/topo-cli/topo-welcome main !2 ?1 ❯ topo ps --target localhost
Image              Status             Ports
topo-welcome-app   Up About an hour   localhost:8000⏎                                                                                                                                                                          ~/repos/topo-cli/topo-welcome main !2 ?1 ❯
```

```
topo ps --target orion
topo-welcome-app   Up 2 hours⏎                                                                                                                                                                                                 ~/repos/topo-cli/topo-welcome main !2 ?1 ❯ topo ps --target orion
Image              Status          Ports
topo-welcome-app   Up 10 seconds   10.2.4.68:8000⏎                                                                                                                                                                             ~/repos/topo-cli/topo-welcome main !2 ?1 ❯
```

### DIFF
--- a/cmd/topo/ps.go
+++ b/cmd/topo/ps.go
@@ -1,13 +1,10 @@
 package main
 
 import (
-	"bytes"
-	"encoding/json"
 	"os"
-	"strings"
 
+	"github.com/arm/topo/internal/deploy"
 	"github.com/arm/topo/internal/deploy/command"
-	"github.com/arm/topo/internal/deploy/operation"
 	"github.com/arm/topo/internal/output/printable"
 	"github.com/arm/topo/internal/output/templates"
 	"github.com/arm/topo/internal/ssh"
@@ -16,10 +13,10 @@ import (
 
 var topoPsCmd = &cobra.Command{
 	Use:   "ps",
-	Short: "List services in a currently running deployment",
-	Long: `List services that are already running on the target host using definitions in the compose file.
+	Short: "List project containers running on the target",
+	Long: `List running containers on the target host for the current Compose project.
 
-The compose file (compose.yaml) must be in the current working directory, as this is used to select containers to be viewed.
+The compose.yaml must be in the current working directory, as this is used to select containers to be viewed.
 `,
 	Args: cobra.ExactArgs(0),
 	RunE: func(cmd *cobra.Command, _ []string) error {
@@ -36,33 +33,23 @@ The compose file (compose.yaml) must be in the current working directory, as thi
 			return err
 		}
 
-		dest := command.NewHostFromDestination(ssh.NewDestination(targetArg))
-
-		ps := operation.NewDockerComposePs(composeFile, dest)
-
-		var rawPsOut bytes.Buffer
-
-		if err := ps.Run(&rawPsOut); err != nil {
+		dest := ssh.NewDestination(targetArg)
+		containers, err := deploy.ListRunningContainers(composeFile, command.NewHostFromDestination(dest))
+		if err != nil {
 			return err
 		}
 
-		decoder := json.NewDecoder(&rawPsOut)
-		containers := []templates.ContainerStatus{}
-		for decoder.More() {
-			var container templates.ContainerStatus
-			if err := decoder.Decode(&container); err != nil {
-				return err
+		hostName := ssh.NewConfig(dest).HostName
+		rows := make([]templates.ContainerStatus, len(containers))
+		for i, c := range containers {
+			rows[i] = templates.ContainerStatus{
+				Image:   c.Image,
+				Status:  c.Status,
+				Address: deploy.PublishedAddress(c.Ports, hostName),
 			}
-			if i := strings.Index(container.Ports, "->"); i != -1 {
-				container.Ports = container.Ports[:i]
-			}
-			container.Ports = strings.ReplaceAll(container.Ports, "0.0.0.0", ssh.NewConfig(ssh.NewDestination(targetArg)).HostName)
-			containers = append(containers, container)
 		}
 
-		toPrint := templates.PrintablePSReport{Containers: containers}
-
-		return printable.Print(toPrint, os.Stdout, outputFormat)
+		return printable.Print(templates.PrintablePSReport{Containers: rows}, os.Stdout, outputFormat)
 	},
 }
 
--- a/cmd/topo/ps.go
+++ b/cmd/topo/ps.go
@@ -13,8 +13,8 @@ import (
 
 var topoPsCmd = &cobra.Command{
 	Use:   "ps",
-	Short: "List project containers running on the target",
-	Long: `List running containers on the target host for the current Compose project.
+	Short: "List running containers on the target for the current Compose project.",
+	Long: `List running containers on the target for the current Compose project.
 
 The compose.yaml must be in the current working directory, as this is used to select containers to be viewed.
 `,
--- a/cmd/topo/ps.go
+++ b/cmd/topo/ps.go
@@ -15,11 +15,8 @@ var topoPsCmd = &cobra.Command{
 	Short: "List services in a currently running deployment",
 	Long: `List services that are already running on the target host using definitions in the compose file.
 
-Executing this command does not remove the containers.
-
-The compose file (compose.yaml) must be in the current working directory, as this is used to select the containers to be stopped.
-
-Use --dry-run to see what commands would be executed without actually running them.`,
+The compose file (compose.yaml) must be in the current working directory, as this is used to select containers to be viewed.
+`,
 	Args: cobra.ExactArgs(0),
 	RunE: func(cmd *cobra.Command, _ []string) error {
 		cmd.SilenceUsage = true
--- a/cmd/topo/ps.go
+++ b/cmd/topo/ps.go
@@ -47,7 +47,7 @@ The compose file (compose.yaml) must be in the current working directory, as thi
 		}
 
 		decoder := json.NewDecoder(&rawPsOut)
-		var containers []templates.ContainerStatus
+		containers := []templates.ContainerStatus{}
 		for decoder.More() {
 			var container templates.ContainerStatus
 			if err := decoder.Decode(&container); err != nil {
--- a/cmd/topo/ps.go
+++ b/cmd/topo/ps.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"os"
+	"strings"
 
 	"github.com/arm/topo/internal/deploy/command"
 	"github.com/arm/topo/internal/deploy/operation"
@@ -52,6 +53,10 @@ The compose file (compose.yaml) must be in the current working directory, as thi
 			if err := decoder.Decode(&container); err != nil {
 				return err
 			}
+			if i := strings.Index(container.Ports, "->"); i != -1 {
+				container.Ports = container.Ports[:i]
+			}
+			container.Ports = strings.ReplaceAll(container.Ports, "0.0.0.0", targetArg)
 			containers = append(containers, container)
 		}
 
--- a/cmd/topo/ps.go
+++ b/cmd/topo/ps.go
@@ -1,0 +1,48 @@
+package main
+
+import (
+	"os"
+
+	"github.com/arm/topo/internal/deploy"
+	"github.com/arm/topo/internal/deploy/command"
+	"github.com/arm/topo/internal/ssh"
+
+	"github.com/spf13/cobra"
+)
+
+var topoPsCmd = &cobra.Command{
+	Use:   "ps",
+	Short: "List services in a currently running deployment",
+	Long: `List services that are already running on the target host using definitions in the compose file.
+
+Executing this command does not remove the containers.
+
+The compose file (compose.yaml) must be in the current working directory, as this is used to select the containers to be stopped.
+
+Use --dry-run to see what commands would be executed without actually running them.`,
+	Args: cobra.ExactArgs(0),
+	RunE: func(cmd *cobra.Command, _ []string) error {
+		cmd.SilenceUsage = true
+
+		targetArg, err := requireTarget(cmd)
+		if err != nil {
+			return err
+		}
+
+		composeFile, err := getComposeFileName()
+		if err != nil {
+			return err
+		}
+
+		dest := command.NewHostFromDestination(ssh.NewDestination(targetArg))
+
+		ps := deploy.NewDeploymentPs(composeFile, dest)
+
+		return ps.Run(os.Stdout)
+	},
+}
+
+func init() {
+	addTargetFlag(topoPsCmd)
+	rootCmd.AddCommand(topoPsCmd)
+}
--- a/cmd/topo/ps.go
+++ b/cmd/topo/ps.go
@@ -34,22 +34,13 @@ The compose.yaml must be in the current working directory, as this is used to se
 		}
 
 		dest := ssh.NewDestination(targetArg)
-		containers, err := deploy.ListRunningContainers(composeFile, command.NewHostFromDestination(dest))
+		hostName := ssh.NewConfig(dest).HostName
+		containers, err := deploy.ListRunningContainers(composeFile, command.NewHostFromDestination(dest), hostName)
 		if err != nil {
 			return err
 		}
 
-		hostName := ssh.NewConfig(dest).HostName
-		rows := make([]templates.ContainerStatus, len(containers))
-		for i, c := range containers {
-			rows[i] = templates.ContainerStatus{
-				Image:   c.Image,
-				Status:  c.Status,
-				Address: deploy.PublishedAddress(c.Ports, hostName),
-			}
-		}
-
-		return printable.Print(templates.PrintablePSReport{Containers: rows}, os.Stdout, outputFormat)
+		return printable.Print(templates.PrintablePSReport{Containers: containers}, os.Stdout, outputFormat)
 	},
 }
 
--- a/cmd/topo/ps.go
+++ b/cmd/topo/ps.go
@@ -56,7 +56,7 @@ The compose file (compose.yaml) must be in the current working directory, as thi
 			if i := strings.Index(container.Ports, "->"); i != -1 {
 				container.Ports = container.Ports[:i]
 			}
-			container.Ports = strings.ReplaceAll(container.Ports, "0.0.0.0", targetArg)
+			container.Ports = strings.ReplaceAll(container.Ports, "0.0.0.0", ssh.NewConfig(ssh.NewDestination(targetArg)).HostName)
 			containers = append(containers, container)
 		}
 
--- a/cmd/topo/ps.go
+++ b/cmd/topo/ps.go
@@ -1,12 +1,15 @@
 package main
 
 import (
+	"bytes"
+	"encoding/json"
 	"os"
 
-	"github.com/arm/topo/internal/deploy"
 	"github.com/arm/topo/internal/deploy/command"
+	"github.com/arm/topo/internal/deploy/operation"
+	"github.com/arm/topo/internal/output/printable"
+	"github.com/arm/topo/internal/output/templates"
 	"github.com/arm/topo/internal/ssh"
-
 	"github.com/spf13/cobra"
 )
 
@@ -20,6 +23,7 @@ The compose file (compose.yaml) must be in the current working directory, as thi
 	Args: cobra.ExactArgs(0),
 	RunE: func(cmd *cobra.Command, _ []string) error {
 		cmd.SilenceUsage = true
+		outputFormat := resolveOutput(cmd)
 
 		targetArg, err := requireTarget(cmd)
 		if err != nil {
@@ -33,9 +37,27 @@ The compose file (compose.yaml) must be in the current working directory, as thi
 
 		dest := command.NewHostFromDestination(ssh.NewDestination(targetArg))
 
-		ps := deploy.NewDeploymentPs(composeFile, dest)
+		ps := operation.NewDockerComposePs(composeFile, dest)
 
-		return ps.Run(os.Stdout)
+		var rawPsOut bytes.Buffer
+
+		if err := ps.Run(&rawPsOut); err != nil {
+			return err
+		}
+
+		decoder := json.NewDecoder(&rawPsOut)
+		var containers []templates.ContainerStatus
+		for decoder.More() {
+			var container templates.ContainerStatus
+			if err := decoder.Decode(&container); err != nil {
+				return err
+			}
+			containers = append(containers, container)
+		}
+
+		toPrint := templates.PrintablePSReport{Containers: containers}
+
+		return printable.Print(toPrint, os.Stdout, outputFormat)
 	},
 }
 
--- a/e2e/deploy_test.go
+++ b/e2e/deploy_test.go
@@ -36,6 +36,7 @@ func TestDeploy(t *testing.T) {
 		port, err := testutil.GetContainerPublicPort(container.Name, "8080")
 		require.NoError(t, err)
 		assertResponseBody(t, fmt.Sprintf("http://localhost:%s/", port), expectedResponse)
+		requirePS(t, topo, projectDir, container.SSHDestination)
 	})
 
 	t.Run("Clone and deploy", func(t *testing.T) {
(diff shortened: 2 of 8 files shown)


cmd/topo/ps.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/arm/topo/internal/ssh"
+	"github.com/spf13/cobra"
+)
+
+var topoPsCmd = &cobra.Command{
+	Use:   "ps",
+	Short: "List services in a currently running deployment",
+	Long: `List services that are already running on the target host using definitions in the compose file.
+
+The compose file (compose.yaml) must be in the current working directory, as this is used to select containers to be viewed.
```
th3james: IMHO talking about containers is more accurate than services. Worth noting that `docker ps` is actually an alias for `docker container ls`
```suggestion
	Short: "List project containers running on the target",
	Long: `List running containers on the target host for the current Compose project.

The compose.yaml must be in the current working directory, as this is used to select containers to be viewed.
```

cmd/topo/ps.go
```diff
@@ -0,0 +1,72 @@
...
+		for decoder.More() {
+			var container templates.ContainerStatus
+			if err := decoder.Decode(&container); err != nil {
+				return err
+			}
+			if i := strings.Index(container.Ports, "->"); i != -1 {
+				container.Ports = container.Ports[:i]
+			}
+			container.Ports = strings.ReplaceAll(container.Ports, "0.0.0.0", ssh.NewConfig(ssh.NewDestination(targetArg)).HostName)
+			containers = append(containers, container)
```
th3james: Feels like there's some pure-function behaviour here which could be extracted and tested in isolation?

cmd/topo/ps.go
```diff
@@ -0,0 +1,72 @@
...
+		}
+
+		composeFile, err := getComposeFileName()
+		if err != nil {
+			return err
+		}
+
+		dest := command.NewHostFromDestination(ssh.NewDestination(targetArg))
+
+		ps := operation.NewDockerComposePs(composeFile, dest)
```
muchzill4: I'm not sure if this is operation-like. Operations are supposed to be "steps in a multi-step workflow".

In addition `NewDockerComposePs` is misaligned with other operations: everywhere else, the output isn't JSON and isn't supposed to be parsed by the caller.

I'd much rather this was `deploy.ListRunningContainers(composeFile, dest) []ContainerInfo` or something along those lines.

As an aside, the moment you see code reaching into internals of another package (`operation` is internals of `deploy`) a warning light should trigger.

cmd/topo/ps.go
```diff
@@ -0,0 +1,72 @@
...
+		containers := []templates.ContainerStatus{}
+		for decoder.More() {
+			var container templates.ContainerStatus
+			if err := decoder.Decode(&container); err != nil {
+				return err
+			}
+			if i := strings.Index(container.Ports, "->"); i != -1 {
+				container.Ports = container.Ports[:i]
+			}
+			container.Ports = strings.ReplaceAll(container.Ports, "0.0.0.0", ssh.NewConfig(ssh.NewDestination(targetArg)).HostName)
```
muchzill4: Is it `Ports`? That name was always a bit odd to me.

Luke-Parkin: Yep, that's valid. Changed to `address`
https://github.com/arm/topo/pull/268/commits/22fe879ef503793a363843c830580005763f8c65

cmd/topo/ps.go
Luke-Parkin: This is fair, I tried to fit it into operations because I thought that was the preferred way to do it, but per your suggestion I think this is a lot nicer?

https://github.com/arm/topo/pull/268/commits/22fe879ef503793a363843c830580005763f8c65

cmd/topo/ps.go
```diff
@@ -0,0 +1,59 @@
...
+			return err
+		}
+
+		hostName := ssh.NewConfig(dest).HostName
+		rows := make([]templates.ContainerStatus, len(containers))
+		for i, c := range containers {
+			rows[i] = templates.ContainerStatus{
+				Image:   c.Image,
+				Status:  c.Status,
+				Address: deploy.PublishedAddress(c.Ports, hostName),
```
muchzill4: I think it's weird that the consumer of `deploy.ListRunningContainers` has to do `deploy.PublishedAddress` to attach metadata to `deploy.Container`. Since `deploy` already couples to `ssh`, we should probably push the calculation of `PublishedAddress` to `deploy` too.

I always find that a really good way to think about the slice of functionality you're putting in any transport layer is: how hard would it be to use in another transport layer. Say, if we were to duplicate `topo ps` as http endpoint, does the current internal api make sense?

Luke-Parkin: https://github.com/arm/topo/pull/268/commits/b995879698112be79d4fb2adc5945daced605eb2

cmd/topo/ps.go
Luke-Parkin: Fixed in https://github.com/arm/topo/pull/268/commits/22fe879ef503793a363843c830580005763f8c65

cmd/topo/ps.go
Luke-Parkin: Yep, pulled out here https://github.com/arm/topo/pull/268/commits/22fe879ef503793a363843c830580005763f8c65

internal/deploy/ps.go
```diff
@@ -0,0 +1,59 @@
...
+	cmd := command.DockerCompose(h, composeFile, "ps", "--format", "json")
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+	if err := cmd.Run(); err != nil {
+		return nil, fmt.Errorf("docker compose ps: %w: %s", err, strings.TrimSpace(stderr.String()))
+	}
+
+	containers := []Container{}
+	decoder := json.NewDecoder(&stdout)
+	for decoder.More() {
```
muchzill4, line 50: TIL

internal/deploy/ps_test.go
```diff
@@ -0,0 +1,39 @@
+package deploy
```
muchzill4: We tend to black box test.

internal/deploy/ps.go
```diff
@@ -0,0 +1,59 @@
...
+	Address string `json:"Address"`
+}
+
+type rawContainer struct {
+	Image  string `json:"Image"`
+	Status string `json:"Status"`
+	Ports  string `json:"Ports"`
+}
+
+func ListRunningContainers(composeFile string, h command.Host, hostName string) ([]Container, error) {
```
muchzill4, line 24: This function suffers from doing a bit much, thus it's hard to test in isolation. It combines executing a command and parsing the output.

Imho a better pattern would be to do what we do in `ssh` package already:
- `NewConfig` - wires up external call + parsing of output (untested)
- `NewConfigFromBytes` - parses the output (tested) - gives an escape hatch if you have ssh config output at hand

In here, we could do:
- `ListRunningContainers`
- `ParseRunningContainers` - a bit of a mouthful, but you get the gist, might be able to use it with `podman` too, this would be your black box test entry point

Food for thought.

e2e/deploy_test.go
```diff
@@ -36,6 +36,7 @@ func TestDeploy(t *testing.T) {
 		port, err := testutil.GetContainerPublicPort(container.Name, "8080")
 		require.NoError(t, err)
 		assertResponseBody(t, fmt.Sprintf("http://localhost:%s/", port), expectedResponse)
+		requirePS(t, topo, projectDir, container.SSHDestination)
```
muchzill4:
```suggestion
		assertResponseBody(t, fmt.Sprintf("http://localhost:%s/", port), expectedResponse)

		requirePS(t, topo, projectDir, container.SSHDestination)
```

e2e/deploy_test.go
```diff
@@ -53,6 +54,7 @@ func TestDeploy(t *testing.T) {
 		port, err := testutil.GetContainerPublicPort(container.Name, "8080")
 		require.NoError(t, err)
 		assertResponseBody(t, fmt.Sprintf("http://localhost:%s/", port), expectedResponse)
+		requirePS(t, topo, cloneDir, container.SSHDestination)
```
muchzill4:
```suggestion
		assertResponseBody(t, fmt.Sprintf("http://localhost:%s/", port), expectedResponse)

		requirePS(t, topo, cloneDir, container.SSHDestination)
```

e2e/deploy_test.go
muchzill4: Perhaps good time to update the description too.

e2e/deploy_test.go
```diff
@@ -125,6 +127,18 @@ func assertResponseBody(t *testing.T, url, wantBody string) {
 	assert.Equal(t, wantBody, string(body))
 }
 
+func requirePS(t *testing.T, topo, projectDir, sshDestination string) {
```
muchzill4: - Can you group related functions together? This belongs with other `requireTopoThing`?
- Can we make it less hardcode-y it assumes container name and port? I noticed `requireExtend` does the same and sobbed.

internal/deploy/ps.go
Luke-Parkin: [Yep, ](https://github.com/arm/topo/pull/268/commits/603e89fa2824c780f2ad14628f2caab530104f70)

e2e/deploy_test.go
Luke-Parkin: https://github.com/arm/topo/pull/268/commits/604688b854d8fdfe308b2aa7e91500d6f3ca0c2b

e2e/deploy_test.go
Luke-Parkin: https://github.com/arm/topo/pull/268/commits/604688b854d8fdfe308b2aa7e91500d6f3ca0c2b

internal/deploy/ps_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+		got := deploy.RemapAddresses(input, "myhost")
+
+		assert.Equal(t, []deploy.Container{{Address: "myhost:8080"}}, got)
+	})
+
+	t.Run("returns an empty slice when given no containers", func(t *testing.T) {
+		got := deploy.RemapAddresses(nil, "myhost")
+
+		assert.Equal(t, []deploy.Container{}, got)
```
muchzill4:
```suggestion
		assert.Empty(t, got)
```

But I also question the value of this test.

internal/deploy/ps_test.go
```diff
@@ -0,0 +1,80 @@
...
+		input := []deploy.RawContainer{
+			{Image: "web", Status: "Up", Ports: "0.0.0.0:8080->80/tcp"},
+		}
+
+		got := deploy.RemapAddresses(input, "myhost")
+
+		assert.Equal(t, []deploy.Container{{Image: "web", Status: "Up", Address: "myhost:8080"}}, got)
+	})
+
+	t.Run("leaves ports untouched when hostName is empty", func(t *testing.T) {
```
muchzill4:
```suggestion
	t.Run("leaves ports untouched when hostname is empty", func(t *testing.T) {
```

internal/deploy/ps_test.go
```diff
@@ -0,0 +1,80 @@
...
+		_, err := deploy.ParseRunningContainers("{not json")
+
+		assert.Error(t, err)
+	})
+}
+
+func TestRemapAddresses(t *testing.T) {
+	t.Run("strips the container-side port mapping and substitutes the hostname", func(t *testing.T) {
+		input := []deploy.RawContainer{
+			{Image: "web", Status: "Up", Ports: "0.0.0.0:8080->80/tcp"},
```
muchzill4: I'm assuming that having `Image:` and `Status:` here is to cover the case that they are retained. I think it belongs in a separate test. This test should only focus on substitution bit.

In other words: avoid polluting the test with data that's not relevant to the test case, it makes it a bit more expensive to reason about.

internal/deploy/ps_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+	t.Run("leaves addresses without 0.0.0.0 untouched", func(t *testing.T) {
+		input := []deploy.RawContainer{{Ports: "127.0.0.1:8080"}}
+
+		got := deploy.RemapAddresses(input, "myhost")
+
+		assert.Equal(t, []deploy.Container{{Address: "127.0.0.1:8080"}}, got)
+	})
+
+	t.Run("keeps only the first mapping when several are published", func(t *testing.T) {
```
muchzill4: Why is this a behaviour we want? Seems order dependant too, so might be that 8443 is _rewritten_, but 80 isn't.

internal/output/templates/ps.go
```diff
@@ -0,0 +1,48 @@
...
+
+func (r PrintablePSReport) AsPlain(isTTY bool) (string, error) {
+	tmpl, err := template.
+		New("ps").
+		Parse(PSTemplate)
+	if err != nil {
+		return "", err
+	}
+	var buf bytes.Buffer
+	w := tabwriter.NewWriter(&buf, 0, 0, 3, ' ', 0)
```
muchzill4: Thoughts about reducing magic numbers and doing:

```
const (
   minwidth = 0
   tabwidth = 0
   ...
)
```

🤷

internal/deploy/ps_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+	"github.com/arm/topo/internal/deploy"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestParseRunningContainers(t *testing.T) {
+	t.Run("decodes the NDJSON stream emitted by docker compose ps", func(t *testing.T) {
+		input := `{"Image":"web","Status":"Up 5 minutes","Ports":"0.0.0.0:8080->80/tcp"}
+{"Image":"db","Status":"Up 5 minutes","Ports":""}`
```
yejseo01, line 14:
```suggestion
		input := strings.Join([]string{
        `{"Image":"web", "Status":"Up 5 minutes", "Ports":"0.0.0.0:8080->80/tcp"}`,
        `{"Image":"db",  "Status":"Up 5 minutes", "Ports":""}`,
  }, "\n")
```

internal/deploy/ps_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+func TestRemapAddresses(t *testing.T) {
+	t.Run("strips the container-side port mapping and substitutes the hostname", func(t *testing.T) {
+		input := []deploy.RawContainer{
+			{Image: "web", Status: "Up", Ports: "0.0.0.0:8080->80/tcp"},
+		}
+
+		got := deploy.RemapAddresses(input, "myhost")
+
+		assert.Equal(t, []deploy.Container{{Image: "web", Status: "Up", Address: "myhost:8080"}}, got)
```
yejseo01: nit: I think this is more readable personally

```suggestion
        want := []deploy.Container{{Image: "web", Status: "Up", Address: "myhost:8080"}}
		assert.Equal(t, want, got)
```

internal/deploy/ps_test.go
```diff
@@ -0,0 +1,80 @@
...
+	})
+
+	t.Run("leaves ports untouched when hostName is empty", func(t *testing.T) {
+		input := []deploy.RawContainer{{Ports: "0.0.0.0:8080->80/tcp"}}
+
+		got := deploy.RemapAddresses(input, "")
+
+		assert.Equal(t, []deploy.Container{{Address: "0.0.0.0:8080->80/tcp"}}, got)
```
yejseo01: https://github.com/arm/topo/pull/268/changes#r3281399429

internal/deploy/ps_test.go
```diff
@@ -0,0 +1,80 @@
...
+	})
+
+	t.Run("leaves addresses without 0.0.0.0 untouched", func(t *testing.T) {
+		input := []deploy.RawContainer{{Ports: "127.0.0.1:8080"}}
+
+		got := deploy.RemapAddresses(input, "myhost")
+
+		assert.Equal(t, []deploy.Container{{Address: "127.0.0.1:8080"}}, got)
```
yejseo01: https://github.com/arm/topo/pull/268/changes#r3281399429

internal/deploy/ps_test.go
```diff
@@ -0,0 +1,80 @@
...
+	})
+
+	t.Run("keeps only the first mapping when several are published", func(t *testing.T) {
+		input := []deploy.RawContainer{{Ports: "0.0.0.0:8080->80/tcp, 0.0.0.0:8443->443/tcp"}}
+
+		got := deploy.RemapAddresses(input, "myhost")
+
+		assert.Equal(t, []deploy.Container{{Address: "myhost:8080"}}, got)
```
yejseo01: https://github.com/arm/topo/pull/268/changes#r3281399429

internal/output/templates/ps.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+type PrintablePSReport struct {
+	Containers []deploy.Container `json:"containers"`
+}
+
+const PSTemplate = `{{if .}}Image	Status	Address
+{{- range .}}
+{{.Image}}	{{.Status}}	{{.Address}}
+{{- end }}{{else}}No containers deployed from this project are running.{{end}}`
```
yejseo01, line 20: got this idea from ai, but I think I like it
```suggestion
  const PSTemplate = `{{if .}}` +
        `Image  Status  Address
  ` +
        `{{range .}}` +
        `{{.Image}}     {{.Status}}     {{.Address}}
  ` +
        `{{end}}` +
        `{{else}}No containers deployed from this project are running.{{end}}`
```

internal/output/templates/ps.go
yejseo01: magic numbers? I suspect theses numbers have to do with buffer sizes, but since there are more than one random numbers fed in this function, might as well make them clear what they are?

yejseo01 (commented):
Looks good. Couple of really nitty stuff.

One question: do we need to update the user guide or any other docs? if we have to, can it be part of this PR so I can try running it?

internal/deploy/ps_test.go
Luke-Parkin: I think it's a reasonable test to have, I don't think it harms to have it?

internal/deploy/ps_test.go
Luke-Parkin: Is there any reason you prefer this? This just seems more complex to me and more unreadable.

internal/output/templates/ps.go
Luke-Parkin: https://github.com/arm/topo/pull/268/commits/76e6b35ee904fb6b9ff6981a51365ea3d9c3fccb

internal/output/templates/ps.go
Luke-Parkin: I think this makes it harder to read, tbh.

internal/output/templates/ps.go
Luke-Parkin: I've added one for the number 3, to indicate what it means- I think the rest are obvious with the function signature, tbh.

internal/deploy/ps_test.go
Luke-Parkin: Fair, I thought it was noise but I think you're right. fixed. https://github.com/arm/topo/pull/268/commits/92570f5686f167ec99f558072f59321f12e9a656

internal/deploy/ps_test.go
Luke-Parkin: https://github.com/arm/topo/pull/268/commits/92570f5686f167ec99f558072f59321f12e9a656
Split!

internal/deploy/ps_test.go
Luke-Parkin: https://github.com/arm/topo/pull/268/commits/5f7626bfa9bcca2ac671a07d4a227a99462dfdb8

internal/deploy/ps_test.go
Luke-Parkin: https://github.com/arm/topo/pull/268/commits/5f7626bfa9bcca2ac671a07d4a227a99462dfdb8

internal/deploy/ps_test.go
Luke-Parkin: https://github.com/arm/topo/pull/268/commits/5f7626bfa9bcca2ac671a07d4a227a99462dfdb8

cmd/topo/ps.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/arm/topo/internal/output/printable"
+	"github.com/arm/topo/internal/output/templates"
+	"github.com/arm/topo/internal/ssh"
+	"github.com/spf13/cobra"
+)
+
+var topoPsCmd = &cobra.Command{
+	Use:   "ps",
+	Short: "List project containers running on the target",
+	Long: `List running containers on the target host for the current Compose project.
```
awphi:
```suggestion
	Long: `List running containers on the target for the current Compose project.
```

We display containers running via remoteproc too right?

cmd/topo/ps.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/arm/topo/internal/deploy/command"
+	"github.com/arm/topo/internal/output/printable"
+	"github.com/arm/topo/internal/output/templates"
+	"github.com/arm/topo/internal/ssh"
+	"github.com/spf13/cobra"
+)
+
+var topoPsCmd = &cobra.Command{
+	Use:   "ps",
+	Short: "List project containers running on the target",
```
awphi: This reads to me like it's more akin to `docker ps` than `docker compose ps`. Maybe we can just re-use the first line of the long desc?

internal/deploy/ps.go
```diff
@@ -0,0 +1,84 @@
...
+	Image  string `json:"Image"`
+	Status string `json:"Status"`
+	Ports  string `json:"Ports"`
+}
+
+type Container struct {
+	Image   string `json:"Image"`
+	Status  string `json:"Status"`
+	Address string `json:"Address"`
+}
```
awphi, line 22: I think we've been using snakeCase for our json tags elsewhere, can we stick to that?

internal/deploy/ps_test.go
```diff
@@ -0,0 +1,91 @@
...
+	})
+
+	t.Run("leaves addresses without 0.0.0.0 untouched", func(t *testing.T) {
+		input := []deploy.RawContainer{{Ports: "127.0.0.1:8080"}}
+
+		got := deploy.RemapAddresses(input, "myhost")
+
+		want := []deploy.Container{{Address: "127.0.0.1:8080"}}
+		assert.Equal(t, want, got)
+	})
```
awphi, line 75: I wonder if we should hide non-accessible addresses from `topo ps` output? Or at least indicate its the target localhost loopback rather than on the host.

Trying to think about the newbie who writes a template, topo deploys and gets confused why they can't reach their app at `127.0.0.1:8080` on their host they copy-pasted from `topo ps`

Just food for thought though - not blocking this PR

internal/deploy/ps.go
```diff
@@ -0,0 +1,84 @@
...
+
+func RemapAddresses(raws []RawContainer, hostName string) []Container {
+	containers := make([]Container, len(raws))
+	for i, raw := range raws {
+		containers[i] = Container{
+			Image:   raw.Image,
+			Status:  raw.Status,
+			Address: publishedAddress(raw.Ports, hostName),
+		}
```
awphi, line 67: Again, perhaps scope for future work and I may have missed some discussion on this but I think it's a shame we don't display more here - particularly id, name and status as they're all useful debugging tools

awphi (commented):
Nice! Couple small nits and some ideas follow-ups to make this even more useful. Can approve once the help strings + json tags are tweaked

internal/deploy/ps_test.go
Luke-Parkin: I had originally had something like this, but we felt it was too clever. I would like to revisit this in the future though

internal/deploy/ps.go
Luke-Parkin: Good catch. https://github.com/arm/topo/pull/268/commits/aa99f630a9762f7e046020c0ba2da62d28a20096

cmd/topo/ps.go
Luke-Parkin: https://github.com/arm/topo/pull/268/commits/aa99f630a9762f7e046020c0ba2da62d28a20096

cmd/topo/ps.go
Luke-Parkin: Good catch https://github.com/arm/topo/pull/268/commits/aa99f630a9762f7e046020c0ba2da62d28a20096

awphi (approved):
LGTM (once GH fixes itself 🙄)

PR #269, opened 2026-05-19
docs: add Topo glossary

ZachMaso:
## Changes

- Add `docs/glossary.md` with definitions for core Topo concepts.
- Link to the glossary from the README after the "Who is this for?" section.

## Validation

- Documentation-only change; no code tests run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 Discover and deploy containerised software to Arm hardware over SSH.
 
-Point Topo at any Arm-based Linux device to discover software templates which showcase its capabilities. Pick one and Topo helps you configure it for your use case, then deploys it in minutes. The result? A standard Docker Compose project to learn from, modify, or use as a starting point for your own work.
+Point Topo at any Arm-based Linux device to discover software Templates which showcase its capabilities. Pick one and Topo helps you configure it for your use case, then deploys it in minutes. The result? A standard Docker Compose project to learn from, modify, or use as a starting point for your own work.
 
 Already have a Compose project? Topo gives you a fast, incremental build-deploy loop over SSH.
 
@@ -14,7 +14,9 @@ Already have a Compose project? Topo gives you a fast, incremental build-deploy 
 
 **You want a faster edit-build-deploy loop.** Build on your laptop and deploy to a Pi or Jetson over SSH. Rebuilds are incremental, so after the first deploy you're often iterating in seconds.
 
-**You have a heterogeneous device and want to use all of it.** Your board has coprocessors like a Cortex-M that normally need separate toolchains and manual firmware loading. Topo and [remoteproc-runtime](https://github.com/arm/remoteproc-runtime) let you orchestrate the whole device as one Docker Compose project.
+**You have a heterogeneous device and want to use all of it.** Your board has remote processors like a Cortex-M that normally need separate toolchains and manual firmware loading. Topo and [Remoteproc Runtime](https://github.com/arm/remoteproc-runtime) let you orchestrate the whole device as one Docker Compose project.
+
+Not sure what these terms mean? The [glossary](docs/glossary.md) defines Topo's core concepts.
 
 ## What does it look like?
 
@@ -22,7 +24,7 @@ Already have a Compose project? Topo gives you a fast, incremental build-deploy 
 # Check your target is ready
 topo health --target pi@raspberrypi
 
-# See which templates match your hardware
+# See which Templates match your hardware
 topo templates --target pi@raspberrypi
 
 # Clone one and configure it for your target
@@ -34,9 +36,9 @@ topo deploy --target pi@raspberrypi
 ## Highlights
 
 - **Fast, incremental deploys** over SSH, with layer caching to keep rebuilds quick
-- **Hardware-aware template discovery** that matches your target's actual capabilities
+- **Hardware-aware Template discovery** that matches your target's actual capabilities
 - **Standard tooling throughout**: Docker Compose, container images, and OCI registries
-- **Whole-device orchestration** of Linux services and coprocessor firmware in a single Compose project
+- **Whole-device orchestration** of Linux services and remote processor firmware in a single Compose project
 
 ## Installation
 
@@ -77,21 +79,21 @@ Alternatively, manually add the appropriate binary from [GitHub Releases](https:
 topo health --target [user@]host
 ```
 
-### 2. Find a template
+### 2. Find a Template
 
 ```sh
 topo templates --target [user@]host
 ```
 
-### 3. Clone your chosen template
+### 3. Clone your chosen Template
 
-Choose a template you wish to try, then clone it:
+Choose a Template you wish to try, then clone it:
 
 ```sh
 topo clone https://github.com/Arm-Examples/topo-welcome.git
 ```
 
-If the template requires build arguments, Topo will prompt you for them.
+If the Template requires build arguments, Topo will prompt you for them.
 
 ### 4. Deploy to your target
 
@@ -104,7 +106,7 @@ Topo builds the container images on your host, transfers them to the target over
 
 ### 5. Review the deployment
 
-Your project is now running on your target. See the template README for details.
+Your project is now running on your target. See the Template README for details.
 
 ### 6. Stop the deployment
 
--- a/cmd/topo/deploy.go
+++ b/cmd/topo/deploy.go
@@ -30,13 +30,13 @@ var deployOpts deploy.DeployOptions
 var deployCmd = &cobra.Command{
 	Use:   "deploy",
 	Short: "Deploy services using the compose file",
-	Long: `Deploy services to the target host using definitions in the compose file.
+	Long: `Deploy services to the target using definitions in the compose file.
 
 This command performs the following operations in sequence:
-  1. Build - Builds container images defined in the compose file on the local host
-  2. Pull - Pulls any required images from registries to the local host
-  3. Transfer - Transfers built and pulled images and compose file to the target host
-  4. Run - Runs docker compose up on the target host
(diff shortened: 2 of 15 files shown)


ZachMaso:
A few terms we may need to adjust more widely across the docs due to this:

1. target host - should this be referred to as 'target system processor'?
2. local host - should this be referred to as just 'host'?
3. x-Topo - should we always refer to this as 'x-topo' or 'x-Topo'?
4. service templates - should these always be referred to as 'templates' or 'topo templates' or is this a different thing?

ZachMaso:
Suggested reviewers: @Luke-Parkin and @th3james.

docs/glossary.md
```diff
@@ -0,0 +1,63 @@
...
+
+A containerised sample project that provides a starting point for development. It includes application code, build configuration, and an **x-topo** metadata block describing required hardware **features** and configurable parameters. **Templates** are configured when cloned. Topo allows the discovery and **deployment** of **templates** for validation.
+
+## x-topo
+
+A metadata extension defined within a Compose file that describes a **template**'s purpose, hardware requirements, **features**, and configurable arguments. Enables tooling (such as Topo) to provide validation, filtering, and interactive setup.
+
+## feature
+
+A declared hardware or software capability required or utilised by a **template** (e.g., NEON support, GPU compute capability, specific **accelerator** support). Used for matching **templates** to compatible **targets** and mapping containers to **processing domains**.
```
federicobozzini: I'm not sure if `software` is correct here. Features are a bit nebulous, but I think we generally mean HW features.

Also I don't think we use them to map containers in any way. That's the subsystems.

ZachMaso: > A declared hardware capability required or utilised by a **template** (e.g., NEON support, GPU compute capability, specific **accelerator** support). Used for matching **templates** to compatible **targets**.

Omitting software seems fine, although it would be useful to explain what impact features have in determining where containers run on the target if any? Would it be correct to say:

> Whether a container can run in a processing domain is decided by the domain's features.

If not is there a similar statement explaining how features determine how a template is deployed across a target?

README.md
```diff
@@ -16,6 +16,8 @@ Already have a Compose project? Topo gives you a fast, incremental build-deploy
 
 **You have a heterogeneous device and want to use all of it.** Your board has coprocessors like a Cortex-M that normally need separate toolchains and manual firmware loading. Topo and [remoteproc-runtime](https://github.com/arm/remoteproc-runtime) let you orchestrate the whole device as one Docker Compose project.
 
+Not sure what these terms mean? The [glossary](docs/glossary.md) defines Topo's core concepts.
```
muchzill4, line 19: Which terms are we referring to here? I skimmed through it, and unless I've missed something (likely!), glossary doesn't seem to cover that big of a surface of the text preceding this sentence.

- Is it worth explaining "heterogeneous device"?
- Perhaps worth clarifying "coprocessors" (named auxiliary in glossary)?

ZachMaso: - Great spot - I've added in a term for heterogeneous device which I think is accurate.

- I also adjusted some terms throughout the repo to align with the glossary for consistency (such as always using the broader term auxiliary processor rather than coprocessor). Would be great to sanity check those, especially 'Service Template' which I believe should be replaced with 'Topo template' but not certain.

docs/glossary.md
```diff
@@ -0,0 +1,67 @@
...
+## deploy
+
+The process of building container images on the **host**, transferring them to the **target**, and starting services, typically via Docker Compose. If a service is configured to run on an **auxiliary processor**, a runtime such as **remoteproc runtime** may be used.
+
+## deployment
+
+The outcome of a successful **deploy** operation. A **deployment** includes all active services, containers, and (where applicable) firmware executing across one or more **processing domains**, and reflects the system as it is currently provisioned and operating.
+
+## incremental build/deploy
+
```
Luke-Parkin, line 62:
```suggestion
A **build** approach where only changed layers are rebuilt, reducing iteration time.
And a **Deploy** approach where only rebuilt layers are transferred to the target.
```
Might make sense to split this up a bit?

docs/glossary.md
```diff
@@ -0,0 +1,67 @@
+# Topo Glossary
+
+This glossary defines the core terms used in Topo documentation and command output.
+
+## host
```
Luke-Parkin: Imo these should all be capitalised

cmd/topo/setup_keys.go
```diff
@@ -16,7 +16,7 @@ var (
 
 var setupKeysCmd = &cobra.Command{
 	Use:   "setup-keys",
-	Short: "Generate SSH keys for the target and install the public key on the target host",
+	Short: "Generate SSH keys for the target and install the public key on the target system processor",
```
Luke-Parkin: This feels wrong here..

```
DESCRIPTION
     ssh (SSH client) is a program for logging into a remote machine and for executing commands on a remote machine.  It is intended to provide secure encrypted communications between two untrusted hosts over an insecure network.  X11 connections, arbitrary TCP ports and UNIX-domain sockets can also be forwarded
     over the secure channel.
```

Target host, target remote machine, make a bit of sense; but `target system processor` implies we're installing a public key.. onto a system processor. Which is not what we're doing.

cmd/topo/vscode.go
```diff
@@ -26,8 +26,8 @@ var getProjectCmd = &cobra.Command{
 
 var describeCmd = &cobra.Command{
 	Use:    "describe",
-	Short:  "Describe the hardware characteristics of the target host",
-	Long:   "Print a description of the hardware characteristics of the target host including CPU ISA features and remoteproc capabilities.",
+	Short:  "Describe the hardware characteristics of the target system processor",
```
Luke-Parkin: This isn't true, we're not **just** describing the hardware characteristics of the target `system processor`.

Again, I think 'system processor' makes it sound like we're describing just the processor- if that were true here that's fine, but we also describe
```
remoteProcessors:
    - name: cix-dsp-rproc
totalMemoryKb: 65037292
```
One of which (memory) is arguably part of the 'system processor' (I would definitely argue against it being so,) and the other is definitely not part of the system processor by the glossary definitions

cmd/topo/setup_keys.go
awphi: Was typing out the exact same comment. I don't like 'host' in this context either but it was accurate, if confusing with our other meaning of 'host'.

Think we need a different term for the OS running on the system processor?

docs/glossary.md
```diff
@@ -0,0 +1,67 @@
...
+
+## heterogeneous device
+
+A device that contains multiple kinds of processing capabilities, such as a **system processor**, **auxiliary processors**, and **accelerators**, across one or more **processing domains**.
+
+## system processor
+
+The main general-purpose **processing domain** running the Linux operating system. It manages system resources, runs user-space applications and containers, and coordinates **auxiliary processors** via frameworks such as **remoteproc runtime**.
+
+## auxiliary processor
```
Luke-Parkin: The definitions for auxilary processor and accelerator are.. a bit muddy here.

These definitions get a bit confused between hardware type, and role/usage.

A DSP is a bad example of this, as it can be either an auxiliary processor, or an accelerator.
And GPUs have their own processing domain, they have their own memory scheduler and execution units.

Maybe something like.. (but not exactly..)

Auxiliary processor: a peer execution environment with independent firmware lifecycle, managed via remoteproc

Accelerator: a workload-specific offload target with no independent firmware lifecycle, accessed via driver or AP

docs/glossary.md
ZachMaso: I think its pretty normal for a glossary to not capitalise its terms unless they're proper nouns. I went back on forth on this, but its more standard than capitalising to me:

https://docs.github.com/en/education/manage-coursework-with-github-classroom/get-started-with-github-classroom/glossary

https://docs.github.com/en/get-started/learning-about-github/github-glossary

docs/glossary.md
ZachMaso: I think this your suggestion is clearer, i'll opt for:

A **build** approach where only changed layers are rebuilt, and a **deploy** approach where only rebuilt layers are transferred to the **target**, reducing iteration time.

cmd/topo/setup_keys.go
ZachMaso: Generate SSH keys for the target and install the public key on the target's primary OS

cmd/topo/deploy.go
```diff
@@ -30,13 +30,13 @@ var deployOpts deploy.DeployOptions
 var deployCmd = &cobra.Command{
 	Use:   "deploy",
 	Short: "Deploy services using the compose file",
-	Long: `Deploy services to the target host using definitions in the compose file.
+	Long: `Deploy services to the target system processor using definitions in the compose file.
```
ZachMaso: @Luke-Parkin is system processor wrong here too?

Luke-Parkin: It does feel a little wrong too, for the reasons we discussed in the sprint planning

docs/glossary.md
federicobozzini: > Whether a container can run in a processing domain is decided by the domain's features.

I don't think so. We don't using the term processing domain and the *subsystem* doesn't depend on the features.

> If not is there a similar statement explaining how features determine how a template is deployed across a target?

Features are not using at all to decide how a template is deployed across a target!

ZachMaso: Fair enough, I've removed both these mistakes from the features definition.

yejseo01 (approved):
I think this is a big improvement! Thanks Zach!

ZachMaso (commented):
Have addressed all comments and made changes accordingly.

awphi (approved):
lgtm!

federicobozzini (approved):

PR #270, opened 2026-05-20
feat: Success message

Luke-Parkin:
## Changes

Add support for the success_message x-topo attribute. This prints a string (defined by compose.yaml) after a successful deployment.

I've used the deployment operations sequence to leverage the operation heading.
With no success_message
```
┌─ Deploy Success ──────────────────────────────────────
Run `topo ps` to see deployed containers
```
with a success_message:
```
┌─ Deploy Success ──────────────────────────────────────
You did it!!
```

### DIFF
--- a/cmd/topo/deploy.go
+++ b/cmd/topo/deploy.go
@@ -102,6 +102,7 @@ The compose file (compose.yaml) must be in the current working directory, as thi
 		if err != nil {
 			return fmt.Errorf("deployment failed; ensure topo health is passing: %w", err)
 		}
+
 		return nil
 	},
 }
--- a/cmd/topo/deploy.go
+++ b/cmd/topo/deploy.go
@@ -102,7 +102,6 @@ The compose file (compose.yaml) must be in the current working directory, as thi
 		if err != nil {
 			return fmt.Errorf("deployment failed; ensure topo health is passing: %w", err)
 		}
-
 		return nil
 	},
 }
--- a/internal/deploy/post_deploy/post_deploy.go
+++ b/internal/deploy/post_deploy/post_deploy.go
@@ -9,20 +9,22 @@ import (
 	"github.com/arm/topo/internal/template"
 )
 
-type PostDeployMessage struct {
-	composeFile string
-	host        command.Host
+type DeploySuccess struct {
+	composeFile    string
+	host           command.Host
+	defaultMessage string
 }
 
-func NewPostDeployMessage(composeFile string, h command.Host) *PostDeployMessage {
-	return &PostDeployMessage{
-		composeFile: composeFile,
-		host:        h,
+func NewDeploySuccess(composeFile string, h command.Host, defaultMessage string) *DeploySuccess {
+	return &DeploySuccess{
+		composeFile:    composeFile,
+		host:           h,
+		defaultMessage: defaultMessage,
 	}
 }
 
-func (t *PostDeployMessage) Description() string {
-	return "Deploy Success"
+func (t *DeploySuccess) Description() string {
+	return "Deployment Success"
 }
 
 func getSuccessMessage(composeFile string) (string, error) {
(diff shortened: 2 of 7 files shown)


cmd/topo/deploy.go
```diff
@@ -102,6 +102,7 @@ The compose file (compose.yaml) must be in the current working directory, as thi
 		if err != nil {
 			return fmt.Errorf("deployment failed; ensure topo health is passing: %w", err)
 		}
+
```
Luke-Parkin:
```suggestion
```

muchzill4:
Needs a mirror PR in https://github.com/arm/topo-template-format.

I wonder if `success_message` is explicit enough. Is it after clone? After extend? Maybe we need `post_deploy_message` or something like that.

Luke-Parkin:
https://github.com/arm/topo-template-format/pull/5

internal/template/definition.go
```diff
@@ -75,10 +76,11 @@ func FromDir(destDir string) (Template, error) {
 }
 
 type rawMetadata struct {
-	Name        string            `yaml:"name"`
-	Description string            `yaml:"description"`
-	Features    []string          `yaml:"features,omitempty"`
-	Args        map[string]rawArg `yaml:"args,omitempty"`
+	Name           string            `yaml:"name"`
+	Description    string            `yaml:"description"`
+	SuccessMessage string            `yaml:"deploy_success_message"`
```
muchzill4: Let's keep naming consistent with other fields: `Name` is `name:`, `Features` are `features:`.

internal/deploy/post_deploy/post_deploy.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/arm/topo/internal/deploy/command"
+	"github.com/arm/topo/internal/template"
+)
+
+type PostDeployMessage struct {
```
muchzill4: Should we add `DeploySuccess`, so that it's nice and consistent? Also, maybe `Print` instead of `Post`, so that it's verb orientated? Unless we're posting a message. 😉 

Should this live in `operations` if it's an operation?

internal/deploy/post_deploy/post_deploy.go
```diff
@@ -0,0 +1,57 @@
...
+	return tpl.Metadata.SuccessMessage, nil
+}
+
+func (p *PostDeployMessage) Run(w io.Writer) error {
+	successMessage, err := getSuccessMessage(p.composeFile)
+	if err != nil {
+		return err
+	}
+	if successMessage == "" {
+		successMessage = "Run `topo ps` to see deployed containers"
```
muchzill4: What if `NewPostDeployMessage(composeFile, h, defaultMessage)`, i.e.: make this an argument instead of hardcoding?

Luke-Parkin: https://github.com/arm/topo/pull/270/commits/5779adddd357f7acffb53b39fade774d190dc989

internal/deploy/post_deploy/post_deploy.go
Luke-Parkin: It's not a Post a deploy message, it's a message post deploy!

internal/template/definition.go
Luke-Parkin: https://github.com/arm/topo/pull/270/commits/25d8647a385710141306aecfb1fb1f1e754ef0c5

internal/deploy/post_deploy/post_deploy.go
Luke-Parkin: Should be better now, not in operation and not 'postdeploymessage'

internal/deploy/post_deploy/post_deploy.go
```diff
@@ -0,0 +1,59 @@
...
+		host:           h,
+		defaultMessage: defaultMessage,
+	}
+}
+
+func (t *DeploySuccess) Description() string {
+	return "Deployment Success"
+}
+
+func getSuccessMessage(composeFile string) (string, error) {
```
yejseo01, line 30: why do we have to open the composeFile every time we run `getSuccessMessage`? Can't we open it once in `NewDeploySuccess`, and set the message in DeploySuccess? We can set the message to default given in `NewDeploySuccess` if the composeFile doesn't have one

yejseo01 (commented):
Looks nice except for one question raised

internal/deploy/post_deploy/post_deploy.go
Luke-Parkin: Not sure I understand, we only call it once, in getSuccessMessage? We only read the compose file once per deploy?

yejseo01 (approved):
LGTM

PR #273, opened 2026-05-21
feat: Experimental support for `--source` in `topo templates`

muchzill4:
## Changes

- Add support for `--source` in `topo templates` when `TOPO_EXPERIMENTAL_FEATURES=1`
- Test with: 
```shell
TOPO_EXPERIMENTAL_FEATURES=1 \
    topo templates \
        --source https://raw.githubusercontent.com/arm/topo/refs/heads/main/internal/catalog/data/templates.json
```
- Also supports `file://` scheme for easier testing locally


cmd/topo/root.go
```diff
@@ -94,3 +94,8 @@ func resolveOutput(cmd *cobra.Command) term.Format {
 	}
 	return term.Plain
 }
+
+func experimentalFeaturesEnabled() bool {
+	const experimentalFeaturesEnvVar = "TOPO_EXPERIMENTAL_FEATURES"
+	return strings.TrimSpace(os.Getenv(experimentalFeaturesEnvVar)) == "1"
+}
```
yejseo01, line 102: should DEVELOPTMENT.md be mentioning this?
Some might say documenting is more maintenance but I think we should document this because of following reasons:
1. I don't think people will use it at the top of their memory very well
2. Unless we use it, this is not nice for the code base
3. it would be good to understand the commitment on using `experimentalFeaturesEnabled` by documenting it. Lets say an experimental feature was later made an official feature. There would be extra work on testing and cleaning. Would be good to provide a good use case on when `experimentalFeaturesEnabled`  should be introduced when we already have a PR system with e2e testing.

yejseo01: Maybe the `experimentalFeatureEnabled` part of the PR should be a separate PR with the DEVELOPMENT.md change?

cmd/topo/templates.go
```diff
@@ -44,5 +56,19 @@ var templatesCmd = &cobra.Command{
...
 	addTimeoutFlag(templatesCmd, defaultTimeout)
+	if experimentalFeaturesEnabled() {
+		templatesCmd.Flags().StringP(sourceFlag, "s", "", "where to source templates' data from")
+	}
 	rootCmd.AddCommand(templatesCmd)
 }
+
+func getSource(cmd *cobra.Command) string {
+	if experimentalFeaturesEnabled() {
+		flagValue, err := cmd.Flags().GetString("source")
```
th3james:
```suggestion
		flagValue, err := cmd.Flags().GetString(sourceFlag)
```

cmd/topo/templates.go
```diff
@@ -44,5 +56,19 @@ var templatesCmd = &cobra.Command{
...
+		if err != nil {
+			panic(fmt.Sprintf("internal error: source flag not registered: %v", err))
```
th3james:
```suggestion
			panic(fmt.Sprintf("internal error: %s flag not registered: %v", sourceFlag, err))
```

cmd/topo/root.go
muchzill4: > should DEVELOPTMENT.md be mentioning this?

Probably not, as I'm still exploring the shape of this toggle. I'm yet to decide if it needs to support gating multiple experimental features for example. I'd treat it as _very_ internal for now. I'd get some mileage out of the current shape to see how this works while continuing to build the support for sourcing templates from remote urls.

cmd/topo/root.go
```diff
@@ -94,3 +94,8 @@ func resolveOutput(cmd *cobra.Command) term.Format {
 	}
 	return term.Plain
 }
+
+func experimentalFeaturesEnabled() bool {
+	const experimentalFeaturesEnvVar = "TOPO_EXPERIMENTAL_FEATURES"
+	return strings.TrimSpace(os.Getenv(experimentalFeaturesEnvVar)) == "1"
```
awphi: We also have a `TOPO_SKIP_VERSION_CHECKS` env var which checks truthiness by whether its set or not, rather than comparing to `"1"`. Ideally the behaviour between the two (and any others) would be consistent imo. Could make a breaking change or change how this one works idm, probably the former is nicer tbh 😬

cmd/topo/templates.go
```diff
@@ -12,23 +13,34 @@ import (
...
+
+		var repos []catalog.Repo
+		var err error
+		source := getSource(cmd)
+		switch source {
+		case "":
+			repos, err = catalog.ListBuiltinTemplates()
+		default:
+			repos, err = catalog.ListTemplatesFromURL(ctx, source)
+		}
```
awphi, line 36: Special case for the empty string meaning default templates is fine but confused me for a brief second - WDYT about changing the signature `getSource` -> `getCustomSource(...) (string, bool)`

then this becomes
```suggestion
		source, ok := getCustomSource(cmd)
		if ok {
	        repos, err = catalog.ListTemplatesFromURL(ctx, source)
		} else {
		    repos, err = catalog.ListBuiltinTemplates()
		}
```

(not linted/fmted)

cmd/topo/templates.go
```diff
@@ -44,5 +56,19 @@ var templatesCmd = &cobra.Command{
...
+	}
 	rootCmd.AddCommand(templatesCmd)
 }
+
+func getSource(cmd *cobra.Command) string {
+	if experimentalFeaturesEnabled() {
+		flagValue, err := cmd.Flags().GetString("source")
+		if err != nil {
+			panic(fmt.Sprintf("internal error: source flag not registered: %v", err))
+		}
```
awphi:
```suggestion
		flagValue, err := cmd.Flags().GetString(sourceFlag)
		if err != nil {
			panic(fmt.Sprintf("internal error: %s flag not registered: %v", sourceFlag, err))
		}
```

internal/catalog/catalog.go
```diff
@@ -28,3 +46,54 @@ func ParseRepos(b []byte) ([]Repo, error) {
...
+		http.MethodGet,
+		parsedURL.String(),
+		nil,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	client := http.Client{}
+	resp, err := client.Do(req) // #nosec G704 -- URL is explicitly provided by the CLI user and scheme-validated above.
```
awphi: Could just use the default client instead of making a new one? No strong opinions here if you disagree though

internal/catalog/catalog.go
```diff
@@ -28,3 +46,54 @@ func ParseRepos(b []byte) ([]Repo, error) {
 	}
 	return templates, nil
 }
+
+func FetchTemplatesJSON(ctx context.Context, url string) ([]byte, error) {
```
th3james: Does this make sense as part of the public interface?

th3james (approved):

cmd/topo/root.go
th3james: I'd also note that this isn't really a developer feature, it's a user feature, so IMHO DEVELOPMENT.md would not be the right home

awphi (approved):
Styley bits but very nice change!

cmd/topo/root.go
muchzill4: https://github.com/arm/topo/pull/278

cmd/topo/templates.go
muchzill4: 28e5c5f0174c5944e1784061a76196896dd4e6d5

cmd/topo/templates.go
muchzill4: 28e5c5f0174c5944e1784061a76196896dd4e6d5

cmd/topo/templates.go
muchzill4: 28e5c5f0174c5944e1784061a76196896dd4e6d5

internal/catalog/catalog.go
muchzill4: 09defd791ae732072b48df2a093dd5dada4147c8

internal/catalog/catalog.go
muchzill4: Good shout, reducing the surface meant tests are now a bit further from parsing, but perhaps that's a good thing: 5aea007d93df48f883bd1e4635a02970ea7249cb

cmd/topo/templates.go
muchzill4: 317c68f7540435eb72e884d69e4fdbbcd5f1ff2e

PR #274, opened 2026-05-26
feat: Add support for topo ps returning processing domains

Luke-Parkin:
## Changes

- Topo has many examples which deploy to heterogenous devices using remoteproc runtime
- This PR adds support to topo ps for displaying which processing domain each container is deployed to.
-

### DIFF
--- a/internal/output/templates/ps.go
+++ b/internal/output/templates/ps.go
@@ -14,9 +14,9 @@ type PrintablePSReport struct {
 	Containers []deploy.Container `json:"containers"`
 }
 
-const PSTemplate = `{{if .}}Image	Status	Address
+const PSTemplate = `{{if .}}Image	Status	Processing Domain	Address
 {{- range .}}
-{{.Image}}	{{.Status}}	{{.Address}}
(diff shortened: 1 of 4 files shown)


internal/output/templates/ps.go
```diff
@@ -14,9 +14,9 @@ type PrintablePSReport struct {
 	Containers []deploy.Container `json:"containers"`
 }
 
-const PSTemplate = `{{if .}}Image	Status	Address
+const PSTemplate = `{{if .}}Image	Status	Processing Domain	Address
```
yejseo01, line 17: what does PS stand for? Maybe not the best to abbreviate it?

yejseo01 (commented):
minor comment

internal/output/templates/ps.go
Luke-Parkin: I think it is best to abbreviate it as that's what everyone calls it 🤷 - I think it's recognisable enough tbh

Luke-Parkin: it means process status

yejseo01 (approved):
LGTM

PR #275, opened 2026-05-27
fix: Fix incorrect matching of container IDs with topo PS

Luke-Parkin:
## Changes

- Docker inspect returns full container IDs, compose ps --format json can return short IDs. 
- This can lead to a mismatch, where the remoteproc services would fall back and incorrectly report as Linux Host.

This PR adds a regression test and a fix- using the short ID only.


federicobozzini (approved):
I don't mind this, but just in case you are not aware you can force a different docker behaviour either with `docker inspect --format '{{truncate .Id 12}}` or the `--no-trunc` option.

PR #277, opened 2026-05-29
feat!: show actionable connectivity health fixes

federicobozzini:
## Changes

- expose actionally connectivity health fixes as already done for the dependencies

For example:

```js
    "connectivity": {
      "name": "Connectivity",
      "status": "error",
      "value": "ssh authentication failed",
      "fix": {
        "description": "Configure SSH keys on remote target",
        "command": "topo setup-keys --target ssh://pi5-rod"
      }
    }
```

### DIFF
--- a/internal/health/health.go
+++ b/internal/health/health.go
@@ -14,13 +14,6 @@ import (
 
 type CheckStatus string
 
-func NewCheckStatusFromError(err error) CheckStatus {
-	if err != nil {
-		return CheckStatusError
-	}
-	return CheckStatusOK
-}
-
 const (
 	CheckStatusOK      CheckStatus = "ok"
 	CheckStatusWarning CheckStatus = "warning"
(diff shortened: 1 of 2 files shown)


internal/health/health.go
```diff
@@ -149,55 +146,50 @@ func GenerateTargetReport(targetStatus Status) TargetReport {
...
-			hc.Fix = ds.Fix
-		}
-		res = append(res, hc)
+		check := HealthCheck{Name: ds.Dependency.Label, Value: ds.Dependency.Binary}
+		res = append(res, healthCheckWithStatusFromError(check, ds.Error, ds.Fix))
 	}
 	return res
 }
+
+func healthCheckWithStatusFromError(check HealthCheck, err error, fix *Fix) HealthCheck {
```
federicobozzini, line 179: I'm not crazy about this name but it's the best I could come up with 🤷

internal/health/health_test.go
```diff
@@ -178,6 +178,34 @@ func TestTargetReport(t *testing.T) {
 			require.NoError(t, json.Unmarshal(b, &result))
 			assert.JSONEq(t, `[]`, string(result["dependencies"]))
 		})
+
+		t.Run("includes connectivity fix", func(t *testing.T) {
```
muchzill4, line 182: 1. What additional coverage does this test add over what we have already? 
2. If there's something we want to test, testing it via json representation of the structs is an odd. Are we trying to cover how the data is converted to json here? Am I missing something?

muchzill4 (commented):
What makes this a breaking change?

muchzill4:
This does not add any new functionality nor is a breaking change. 🤷

PR #278, opened 2026-06-01
fix!: Skip version checks argument precedence and truthiness

muchzill4:
## Changes

- `SKIP_VERSION_CHECKS` must now be set to truthy value to be considered enabled (`1`, `yes`, `on`, `true`, `enabled`)
- Fixed a bug where explicit false value provided in cli could be overridden by `SKIP_VERSION_CHECKS` env variable being non-empty

### DIFF
--- a/cmd/topo/health.go
+++ b/cmd/topo/health.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/arm/topo/internal/env"
 	"github.com/arm/topo/internal/health"
 	"github.com/arm/topo/internal/output/printable"
 	"github.com/arm/topo/internal/output/templates"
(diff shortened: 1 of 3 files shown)


cmd/topo/health.go
```diff
@@ -76,14 +77,20 @@ func init() {
...
+		return false, false
 	}
 
-	return skipVersionCheck
+	skipVersionChecks, err := cmd.Flags().GetBool(skipVersionChecksFlag)
+	if err != nil {
+		panic(fmt.Sprintf("internal error: %s flag not registered: %v", skipVersionChecksFlag, err))
+	}
+	return skipVersionChecks, true
 }
```
awphi, line 89:
```go
func resolveSkipVersionChecks(cmd *cobra.Command) bool {
	if !cmd.Flags().Changed(skipVersionChecksFlag) {
		return env.IsVarTruthy(skipVersionChecksEnvVar)
	}

	skipVersionChecks, err := cmd.Flags().GetBool(skipVersionChecksFlag)
	if err != nil {
		panic(fmt.Sprintf("internal error: %s flag not registered: %v", skipVersionChecksFlag, err))
	}
	return skipVersionChecks
}
```

thoughts? Too much behaviour in one function? Just trying to avoid the `(bool, bool)` return 😅

awphi (approved):
Nice! Good spot on the bug too. Some stylistic food for thought but not blocking :)

cmd/topo/health.go
muchzill4: Much nicer, thanks. https://github.com/arm/topo/pull/278/commits/4ad2f7cc3ef5e318815c008ff8c42799023f086c

awphi (approved):

PR #282, opened 2026-06-02
refactor: Views not templates

muchzill4:
## Changes

Goal: Reduce the proliferation of the noun template in the codebase as it's a reserved word in `topo` land.

- Merge `internal/output/templates` and `internal/output/printable` into a single `internal/output/views` package                                                                                             
- Rename the output rendering interface from `Printable` to `View`                                                                                                                                          
- Remove redundant `Printable*` prefixes from view types
- Rename view files to match the primary types they contain


awphi (approved):
Nice tidy-up! 🚀

PR #284, opened 2026-06-03
fix: Address GO-2026-5037 by updating go to 1.26.4

muchzill4:


awphi (approved):
PR title needs updating to reference correct version for changelog purposes but otherwise nice!

PR #285, opened 2026-06-04
feat: Add template catalog schema

muchzill4:
## Changes

- Incremental change moving us towards externally source-able template list
- Add JSON schema based on existing `templates.schema.json`
- Reference schema in the `templates.json`
- Reject remote template listing if it doesn't match the schema


internal/catalog/data/templates.json
```diff
@@ -1,38 +1,41 @@
...
-    "features": [
-      "NEON",
-      "SVE"
-    ],
-    "url": "https://github.com/Arm-Examples/topo-simd-visual-benchmark.git",
-    "ref": "main"
-  }
-]
+{
+  "$schema": "https://topo.arm.com/schemas/templates/1/schema.json",
```
th3james: How are we going to make this happen?

internal/catalog/data/templates.schema.json
```diff
@@ -0,0 +1,69 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://topo.arm.com/schemas/templates/1/schema.json",
+  "title": "Topo Templates Catalog",
+  "description": "A catalog of Topo template repositories.",
```
th3james:
```suggestion
  "description": "A catalog of Topo Template repositories.",
```

th3james (approved):
Looks good other than uncertainty about whether we should reference the https://topo.arm.com/schemas/templates/1/schema.json prior to actually publishing it. Probably makes sense to stick with the GitHub references until that time

internal/catalog/data/templates.json
muchzill4: https://github.com/arm/topo/pull/285/commits/d686ecb7ed3252f1a9d7aad8383856000e81b28b

PR #287, opened 2026-06-05
fix: make `update_templates` script work with new catalog format

muchzill4:
## Changes

- Update `scripts/update_templates` and regenerate `internal/catalog/data/catalog.json`    
- Rename built-in catalog data files from `templates.*` to `catalog.*`


internal/catalog/catalog.go
```diff
@@ -61,10 +61,10 @@ func parseTemplates(b []byte) ([]Template, error) {
 }
 
 func validateAgainstSchema(b []byte) error {
-	const templatesSchemaURL = "https://topo.arm.com/schemas/templates/1/schema.json"
+	const templatesSchemaURL = "https://raw.githubusercontent.com/arm/topo/main/internal/catalog/data/catalog.schema.json"
```
muchzill4, line 64: It seems that it's not strictly necessary to match that with reality, but a good idea to reference schemas by their id none the less.

th3james (approved):
Good hedge against any future renaming, too 😆

PR #288, opened 2026-06-05
docs: mention homebrew installation method

muchzill4:
## Changes

- Documents a one-liner to install `topo` via https://github.com/arm/homebrew-topo

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,14 @@ The host and target can be the same system. If you're working directly on an Arm
 
 ### Linux and macOS
 
+Using [Homebrew](https://github.com/arm/homebrew-topo):
+
+```sh
+brew install arm/topo/topo
+```
+
+Or use the install script:
+
 ```sh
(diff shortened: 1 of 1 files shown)


th3james (commented):
I think we have a usability bug we need to fix before we merge this. As I understand it, https://github.com/arm/homebrew-topo lags the current version published to artifactory until we manually approve merges with new versions. During this window, `topo health` will report that topo is out of date and tell the users to `brew upgrade topo`, but that won't fix the issue.
I think we need to resolve that before merging this PR

muchzill4:
> I think we have a usability bug we need to fix before we merge this. As I understand it, https://github.com/arm/homebrew-topo lags the current version published to artifactory until we manually approve merges with new versions. During this window, `topo health` will report that topo is out of date and tell the users to `brew upgrade topo`, but that won't fix the issue. I think we need to resolve that before merging this PR

https://github.com/arm/topo/pull/289

th3james (approved):

PR #289, opened 2026-06-08
fix: When managed by `homebrew` check for latest version in the tap

muchzill4:
## Changes

- Fixes an issue where user can be presented with `brew upgrade` based on artifactory release, which is not yet reflected in the homebrew tap.


yejseo01 (approved):
Looks good to me!

PR #292, opened 2026-06-09
refactor: Catalog update script

muchzill4:
## Changes

- Sever the update scripts coupling with topo `internal`
- Refactor template update scripts to separate GitHub source metadata, GitHub fetching, parsing, and writing concerns
- Make GITHUB_TOKEN optional, warning about possible rate limiting instead of failing immediately
- Resolve the catalog output path relative to the repo root so the script works from subdirectories                                                                                                  
 
### Going forward / known issues

- Decide what to do when a single template fails to update (right now it will be dropped)
- Validate the written catalog is schema compatible (we used to check `x-topo.name`, but we should be stricter)
- Organise `catalog.*.json` so that it's not internal to topo
- Source template list not from a `const`, but a file sibling to above
- Rename `ref` to `sha` everywhere and actually store sha's instead of `main`
- Setup CI job to run updater periodically


scripts/update_templates/github_source.go
```diff
@@ -0,0 +1,36 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+)
+
+// TODO: sha should become a hash
```
awphi: Is there a reason not to address this TODO today?

scripts/update_templates/main.go
```diff
@@ -1,80 +1,67 @@
...
 	var templates []Template
-
-	seenNames := make(map[string]struct{})
-
-	for _, spec := range repoList {
-		repo, ref := parseRepoSpec(spec)
-
-		composeBytes, err := fetchComposeFile(client, token, spec)
+	for _, source := range ListGitHubSources(strings.NewReader(sourcesJSON)) {
+		template, err := FetchTemplate(githubClient, source)
```
awphi, line 20: Fun opportunity for concurrency here but fine for now as we've got so few templates in the list

awphi (approved):
One last thing I was thinking about not changed here though - should the generated catalog JSON's schema link point to the schema at the ref from which it was generated?

scripts/update_templates/main.go
muchzill4: Agreed, I'm also excited about this. 😄

scripts/update_templates/github_source.go
muchzill4: f6d5bac62836dc519dd835a2f6651d5aa6c82905

muchzill4:
> One last thing I was thinking about not changed here though - should the generated catalog JSON's schema link point to the schema at the ref from which it was generated?

The plan is to use versioned schema eventually, I think pinning to a concrete ref will happen then.

PR #293, opened 2026-06-09
feat!: rename Subsystem -> Processing Domain

tgonzalezorlandoarm:
Does what it says on the tin, to align with the Topo glossary.

### DIFF
--- a/internal/health/health.go
+++ b/internal/health/health.go
@@ -48,11 +48,11 @@ func (r HostReport) MarshalJSON() ([]byte, error) {
 }
 
 type TargetReport struct {
-	Destination     string        `json:"destination"`
-	IsLocalhost     bool          `json:"isLocalhost"`
-	Connectivity    HealthCheck   `json:"connectivity"`
-	Dependencies    []HealthCheck `json:"dependencies"`
-	SubsystemDriver HealthCheck   `json:"subsystemDriver"`
+	Destination            string        `json:"destination"`
+	IsLocalhost            bool          `json:"isLocalhost"`
+	Connectivity           HealthCheck   `json:"connectivity"`
+	Dependencies           []HealthCheck `json:"dependencies"`
+	ProcessingDomainDriver HealthCheck   `json:"processingDomainDriver"`
 }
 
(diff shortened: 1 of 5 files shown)


internal/health/health.go
```diff
@@ -125,7 +125,7 @@ func GenerateTargetReport(targetStatus Status) TargetReport {
 	report.IsLocalhost = targetStatus.Connection.IsPlainLocalhost()
 	report.Connectivity = connectivityCheck(targetStatus.Connection)
 
-	report.SubsystemDriver.Name = "Subsystem Driver (remoteproc)"
+	report.SubsystemDriver.Name = "Processing Domain Driver (remoteproc)"
```
federicobozzini: If we want to make this change (no strong opinion here) we would also need to change the code (EG: `report.ProcessingDomainDriver`)

tgonzalezorlandoarm: done!

internal/health/health.go
```diff
@@ -48,11 +48,11 @@ func (r HostReport) MarshalJSON() ([]byte, error) {
...
-	Destination     string        `json:"destination"`
-	IsLocalhost     bool          `json:"isLocalhost"`
-	Connectivity    HealthCheck   `json:"connectivity"`
-	Dependencies    []HealthCheck `json:"dependencies"`
-	SubsystemDriver HealthCheck   `json:"subsystemDriver"`
+	Destination            string        `json:"destination"`
+	IsLocalhost            bool          `json:"isLocalhost"`
+	Connectivity           HealthCheck   `json:"connectivity"`
+	Dependencies           []HealthCheck `json:"dependencies"`
+	ProcessingDomainDriver HealthCheck   `json:"processingDomainDriver"`
```
awphi, line 55: The json schema change makes this change breaking so probably should be `feat!:`?

awphi (commented):
One semver comment

Also 'processing domain driver' feels really weird but consistently weird > inconsistent and confusing

internal/health/health.go
tgonzalezorlandoarm: good point, done!

awphi (approved):

PR #294, opened 2026-06-10
refactor: Template updater file handling

muchzill4:
## Changes

- Move hard-coded template source entries into `scripts/update_templates/sources.json`
- Extract repository-relative file path helpers into `files.go`
- Change `WriteTemplates` to write to an `io.Writer` instead of creating files directly

### DIFF
--- a/scripts/update_templates/files.go
+++ b/scripts/update_templates/files.go
@@ -1,0 +1,48 @@
+package main
+
+import (
+	"os"
+	"path/filepath"
+)
+
+// TODO: Move files out of internal when we extract this to a separate repo
+const relativeCatalogOutputPath = "internal/catalog/data/catalog.json"
+
+const relativeSourcesPath = "scripts/update_templates/sources.json"
+
+func catalogOutputPath() (string, error) {
+	repoRoot, err := findRepoRoot()
+	if err != nil {
+		return "", err
+	}
+
+	return filepath.Join(repoRoot, filepath.FromSlash(relativeCatalogOutputPath)), nil
+}
+
+func openGitHubSources() (*os.File, error) {
+	repoRoot, err := findRepoRoot()
+	if err != nil {
+		return nil, err
+	}
+
+	return os.Open(filepath.Join(repoRoot, filepath.FromSlash(relativeSourcesPath)))
+}
+
+func findRepoRoot() (string, error) {
+	dir, err := os.Getwd()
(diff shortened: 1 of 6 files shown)


scripts/update_templates/main.go
```diff
@@ -26,42 +30,20 @@ func main() {
...
+	outputFile, outputPath, err := createCatalogOutput()
 	if err != nil {
-		log.Fatalf("failed to find catalog output path: %v\n", err)
+		log.Fatalf("failed to create catalog output: %v\n", err)
 	}
 
-	if err := WriteTemplates(outputPath, templates); err != nil {
-		log.Printf("failed to write templates: %v\n", err)
+	writeErr := WriteTemplates(outputFile, templates)
+	closeErr := outputFile.Close()
```
muchzill4, line 39: Nit to self: why is this handled and other `Close` isn't.

scripts/update_templates/sources.json
```diff
@@ -0,0 +1,6 @@
+[
+	{"repo": "Arm-Examples/topo-welcome", "sha": "8303e66db59a7a11e64877121f3db1b688d2011f"},
```
th3james, line 2: Separate PR, but `sha` is a bit "how, not what". IMHO this would be better.
```suggestion
	{"repo": "Arm-Examples/topo-welcome", "commit_hash": "8303e66db59a7a11e64877121f3db1b688d2011f"},
```

scripts/update_templates/files.go
```diff
@@ -0,0 +1,62 @@
...
+func openGitHubSources() (*os.File, error) {
+	repoRoot, err := findRepoRoot()
+	if err != nil {
+		return nil, err
+	}
+
+	return os.Open(filepath.Join(repoRoot, filepath.FromSlash(relativeSourcesPath)))
+}
+
+func findRepoRoot() (string, error) {
```
th3james: nit: I'd intuit that this function as named would find the `.git` repo root, rather than the go module root
```suggestion
func findModuleRoot() (string, error) {
```

th3james (approved):

scripts/update_templates/files.go
muchzill4: https://github.com/arm/topo/pull/294/commits/50b2ad1f32a93bfcf39c3fbbd6314af5fb9b70d8

PR #295, opened 2026-06-10
feat: Validate fetched templates before writing

muchzill4:
## Changes

- Validate fetched templates against the catalog JSON schema before adding them to the generated catalog


scripts/update_templates/files.go
```diff
@@ -10,6 +10,8 @@ const relativeCatalogOutputPath = "internal/catalog/data/catalog.json"
 
 const relativeSourcesPath = "scripts/update_templates/sources.json"
 
+const relativeCatalogSchemaPath = "internal/catalog/data/catalog.schema.json"
```
Luke-Parkin: Hm; Didn't know we had this in here. Do we need to start versioning the topo template schema and making sure we pin to one?

PR #299, opened 2026-06-12
feat: Incremental catalog update plan

muchzill4:
## Changes

- Update template catalog generation to read the existing catalog and build an incremental update plan before fetching templates
- Only fetch templates that need to be added or updated, remove stale entries, and exit early when the catalog is already up to date
- Preserve configured template source order and use repository URLs as stable source identifiers
- Improve update logging with add/update/remove/unchanged plan details

### DIFF
--- a/scripts/update_templates/template_sort.go
+++ b/scripts/update_templates/template_sort.go
@@ -1,0 +1,19 @@
+package main
+
+func TemplatesInSourceOrder(sources []GitHubSource, templates []Template) []Template {
+	templateByID := make(map[TemplateSourceID]Template, len(templates))
+	for _, template := range templates {
+		templateByID[template.SourceID()] = template
+	}
+
+	ordered := make([]Template, 0, len(sources))
+	for _, source := range sources {
+		template, exists := templateByID[source.ID()]
+		if !exists {
+			continue
+		}
+		ordered = append(ordered, template)
+	}
+
+	return ordered
+}
--- a/scripts/update_templates/template_sort.go
+++ b/scripts/update_templates/template_sort.go
@@ -1,5 +1,7 @@
 package main
 
+import "fmt"
+
 func TemplatesInSourceOrder(sources []GitHubSource, templates []Template) []Template {
 	templateByID := make(map[TemplateSourceID]Template, len(templates))
 	for _, template := range templates {
@@ -10,7 +12,7 @@ func TemplatesInSourceOrder(sources []GitHubSource, templates []Template) []Temp
 	for _, source := range sources {
 		template, exists := templateByID[source.ID()]
 		if !exists {
-			continue
+			panic(fmt.Sprintf("missing template for source %s", source))
 		}
 		ordered = append(ordered, template)
 	}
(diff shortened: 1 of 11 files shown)


scripts/update_templates/update_plan.go
```diff
@@ -0,0 +1,79 @@
...
+func PlanUpdate(sources []GitHubSource, current []Template) UpdatePlan {
+	currentByID := make(map[TemplateSourceID]Template, len(current))
+	for _, template := range current {
+		currentByID[template.SourceID()] = template
+	}
+
+	sourceByID := make(map[TemplateSourceID]GitHubSource, len(sources))
+	for _, source := range sources {
+		sourceByID[source.ID()] = source
+	}
```
th3james, line 49: extreme nitpick, but it'd swap the order of these two loops to be consistent with the argument order. Might lead the reader to think there's an order dependency
```suggestion
	sourceByID := make(map[TemplateSourceID]GitHubSource, len(sources))
	for _, source := range sources {
		sourceByID[source.ID()] = source
	}

	currentByID := make(map[TemplateSourceID]Template, len(current))
	for _, template := range current {
		currentByID[template.SourceID()] = template
	}
```

scripts/update_templates/template_sort.go
```diff
@@ -0,0 +1,19 @@
...
+	templateByID := make(map[TemplateSourceID]Template, len(templates))
+	for _, template := range templates {
+		templateByID[template.SourceID()] = template
+	}
+
+	ordered := make([]Template, 0, len(sources))
+	for _, source := range sources {
+		template, exists := templateByID[source.ID()]
+		if !exists {
+			continue
```
th3james: This is unreachable and would represent a bug

```suggestion
	panic(fmt.Sprintf("missing template for source %s", source))
```

scripts/update_templates/template_sort_test.go
```diff
@@ -0,0 +1,32 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestTemplatesInSourceOrder(t *testing.T) {
+	t.Run("returns templates ordered to match sources", func(t *testing.T) {
```
th3james, line 10: It's permissible to have two identical source IDs with different refs, right? I could see wanting to have two entries for the same ID but different branches. Might be worth asserting against that case if we don't support it, because right now it would silently drop, I believe?

th3james (approved):

scripts/update_templates/update_plan.go
muchzill4: https://github.com/arm/topo/pull/299/commits/d105dacc1d3f71fe4b390ee7e108d2888ff18f00

scripts/update_templates/template_sort.go
muchzill4: https://github.com/arm/topo/pull/299/commits/dc83134ee524418f11945b2dac937edfa205df38

scripts/update_templates/template_sort_test.go
muchzill4: https://github.com/arm/topo/pull/299/commits/43578e87e23fda284d7fa27da2db33f9c6806ec7

PR #300, opened 2026-06-15
feat: Validate catalog template updates

muchzill4:
## Changes

- Add catalog schema validation for fetched templates and generated catalog output
  - When fetching new templates, validate they conform before adding them to collection
  - When writing the catalog, validate it conforms

### DIFF
--- a/scripts/update_templates/main.go
+++ b/scripts/update_templates/main.go
@@ -24,7 +24,12 @@ func main() {
 		log.Fatalf("failed to list sources: %v\n", err)
 	}
 
-	currentTemplates, err := ReadTemplates()
+	catalogFilePath, err := CatalogFilePath()
+	if err != nil {
+		log.Fatalf("failed to find catalog file: %v\n", err)
+	}
+
+	currentTemplates, err := ReadTemplates(catalogFilePath)
 	if err != nil {
 		log.Fatalf("failed to read catalog file: %v\n", err)
 	}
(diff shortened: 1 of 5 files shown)


scripts/update_templates/main.go
```diff
@@ -36,22 +41,34 @@ func main() {
...
 		log.Printf("fetched %s\n", source)
 		templates = append(templates, template)
 	}
 	templates = TemplatesInSourceOrder(sources, templates)
 
-	filePath, err := WriteTemplates(templates)
-	if err != nil {
+	if err := WriteTemplates(catalogFilePath, templates, validator); err != nil {
 		log.Fatalf("failed to write catalog file: %v\n", err)
 	}
```
awphi: nit: perhaps a little confusing that when updating templates fetch/validate are two different steps ran in sequence but here we embed the validation inside the write function. maybe catalog creation + validation should be hoisted in here?

awphi (approved):
Nice!

scripts/update_templates/main.go
muchzill4: Do you mean have two calls here, such that:

```go
catalogToWrite := Catalog{...}
validator.Validate(catalogToWrite)
WriteCatalog(catalogFilePath, catalogToWrite)
```

awphi: Yes! Potentially too much logic for the main func I guess but just food for thought

muchzill4: https://github.com/arm/topo/pull/300/commits/691b271c1dd7b09d88cd6616597a9b49f24bae18

PR #303, opened 2026-06-16
feat: add `topo configure` for re-configuration of project parameters

awphi:
## Changes

- Adds a `topo configure` command which allows for the complete re-configuration of a Topo Project's parameters 
  - Note this command introduces a third place where a parameter provider chain is constructed from CLI args + TTY input (`clone`, `extend` and `configure`) so potentially worth commonising these though I've not done that here for simplicity's sake.
- Replaces 'Input args' in cloning project configuration with 'Configure project' to define a neater link between the two concepts

Note this PR explicitly does not update the README, glossary or clone/extend help strings as it feels like that should be done as part of a wider language update that spans across all "templates" as they're known currently.

### DIFF
--- a/cmd/topo/configure.go
+++ b/cmd/topo/configure.go
@@ -1,0 +1,55 @@
+package main
+
+import (
+	"os"
+
+	"github.com/arm/topo/internal/arguments"
+	"github.com/arm/topo/internal/output/term"
+	"github.com/arm/topo/internal/project"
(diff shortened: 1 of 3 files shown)


Luke-Parkin (approved):
LGTM, I wonder if we need some clearer messaging around this also re-configuring, or if people will naturally assume it can reconfigure.

PR #306, opened 2026-06-18
fix: Support cloning Git object IDs

muchzill4:
## Changes

- Support Git template sources pinned to full object IDs by fetching and checking out the commit directly
  - ~Avoid detached head state by creating a branch at the commit called `topo-<short-id>`~
- Keep normal branch/tag refs on the existing shallow clone path

### DIFF
--- a/internal/template/source.go
+++ b/internal/template/source.go
@@ -126,8 +126,7 @@ func cloneCommitTo(url, commit, destDir string) error {
 	if err := runGit(destDir, "fetch", "--depth", "1", "origin", commit); err != nil {
 		return fmt.Errorf("failed to fetch git commit: %w", err)
 	}
-	branchName := "topo-" + shortGitObjectID(commit)
-	if err := runGit(destDir, "checkout", "--quiet", "-B", branchName, "FETCH_HEAD"); err != nil {
+	if err := runGit(destDir, "checkout", "--quiet", "--detach", "FETCH_HEAD"); err != nil {
 		return fmt.Errorf("failed to check out git commit: %w", err)
 	}
 
@@ -158,14 +157,6 @@ func isFullGitObjectID(ref string) bool {
 		}
 	}
 	return true
-}
-
-func shortGitObjectID(ref string) string {
-	const shortObjectIDLength = 7
-	if len(ref) <= shortObjectIDLength {
-		return ref
-	}
-	return ref[:shortObjectIDLength]
 }
 
 func isHexDigit(r rune) bool {
--- a/internal/template/source.go
+++ b/internal/template/source.go
@@ -88,16 +88,90 @@ func (g GitSource) CopyTo(destDir string) error {
 		return DestDirExistsError{Dir: destDir}
 	}
 
-	args := []string{"clone", "--depth", "1"}
-	if g.Ref != "" {
-		args = append(args, "--branch", g.Ref)
+	if isFullGitObjectID(g.Ref) {
+		return cloneCommitTo(g.URL, g.Ref, destDir)
 	}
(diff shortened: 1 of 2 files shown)


awphi:
Seems relatively modern git version support `git clone <repo> --revision <sha>` - wondering if we should mirror this with a `topo clone --revision`, giving us a clearer path to deprecating `topo clone` entirely as we've been discussing offline.

internal/template/source.go
```diff
@@ -88,18 +88,92 @@ func (g GitSource) CopyTo(destDir string) error {
...
+		return fmt.Errorf("failed to add git remote: %w", err)
+	}
+	if err := runGit(destDir, "fetch", "--depth", "1", "origin", commit); err != nil {
+		return fmt.Errorf("failed to fetch git commit: %w", err)
 	}
-	args = append(args, g.URL, destDir)
+	branchName := "topo-" + shortGitObjectID(commit)
+	if err := runGit(destDir, "checkout", "--quiet", "-B", branchName, "FETCH_HEAD"); err != nil {
+		return fmt.Errorf("failed to check out git commit: %w", err)
+	}
```
awphi: Is it worth bother to avoid a detached head? `git clone --revision` doesn't bother so wondering if we should align with its behaviour

muchzill4:
> Seems relatively modern git version support `git clone <repo> --revision <sha>` - wondering if we should mirror this with a `topo clone --revision`, giving us a clearer path to deprecating `topo clone` entirely as we've been discussing offline.

That would mean breaking change to `topo clone`. I'd rather fix the immediate bug we have right now.

Now, I wonder if the internal implementation should use the flag you mention, however I do like the fact we don't end up in a detached head state. Also, the argument seems to have landed in 2.49 released in March 2025, so I'm not sure how much we can rely on it being the _default_ version across distros and OSes. 🤔 Thoughts?

awphi (approved):
LGTM! Small question about avoiding detached HEAD but doesn't block either way.

Also curious on your thoughts about using `git clone --revision` over rolling our own? Seems it was released in 2.49 which came out in March 2025 so perhaps too new?

awphi:
> > Seems relatively modern git version support `git clone <repo> --revision <sha>` - wondering if we should mirror this with a `topo clone --revision`, giving us a clearer path to deprecating `topo clone` entirely as we've been discussing offline.
> 
> That would mean breaking change to `topo clone`. I'd rather fix the immediate bug we have right now.
> 
> Now, I wonder if the internal implementation should use the flag you mention, however I do like the fact we don't end up in a detached head state. Also, the argument seems to have landed in 2.49 released in March 2025, so I'm not sure how much we can rely on it being the _default_ version across distros and OSes. 🤔 Thoughts?

Lol, asked the same thing at the same time. Imo we should fix the problem today without changing git version requirements and re-consider in the future. Realistically it's probably a bit too new but it would be so perfect if we could use it as it could render `topo clone` mostly obsolete.

internal/template/source.go
muchzill4: In all honesty, I was debating doing `rm -rf .git` after cloning, because now that we want to checkout an object, you won't be pulling any updates (easily). 🤔 Also, it'd be a clean slate to "start" a new project. However.. it is called `topo clone`, so it has weird semantics.

I'm happy either way. Removing it means less code to maintain and as you say it aligns with `git clone --revision` semantics.

muchzill4: https://github.com/arm/topo/pull/306/commits/b4b80a8ca29087f97e4f35c9afe80488bfc1a9bf

awphi (approved):

PR #307, opened 2026-06-18
feat: add `-a`/`--all` flag to `topo ps`

awphi:
## Changes

- Adds an `-a`/`--all` flag to `topo ps` which is just passed through to `docker compose ps` directly
   - Just adds an e2e test for this since the seam it lives on atm is hard to unit test as-is and refactoring didn't feel worth it
- Removes the special case human message we displayed if no containers were running when `topo ps` was invoked - it felt weird to me to deviate from the behaviour `docker ps`/`docker compose ps`. Happy to revert this change specifically if it's disagreeable though it would need re-wording depending on the presence of the `-a` flag which is a bit annoying.

The medium-term goal here is to making `topo ps` usable for determining/rendering the state of a project's containers without extra docker commands for things like the VS Code extension project view.

### DIFF
--- a/cmd/topo/ps.go
+++ b/cmd/topo/ps.go
@@ -12,8 +12,8 @@ import (
 
 var topoPsCmd = &cobra.Command{
 	Use:   "ps",
-	Short: "List running containers on the target for the current Compose project.",
-	Long: `List running containers on the target for the current Compose project.
+	Short: "List containers on the target for the current Compose project.",
+	Long: `List containers on the target for the current Compose project.
 
 The compose.yaml must be in the current working directory, as this is used to select containers to be viewed.
 `,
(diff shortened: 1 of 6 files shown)


cmd/topo/ps.go
```diff
@@ -12,8 +12,8 @@ import (
 
 var topoPsCmd = &cobra.Command{
 	Use:   "ps",
-	Short: "List running containers on the target for the current Compose project.",
-	Long: `List running containers on the target for the current Compose project.
```
th3james, line 16: 👏

cmd/topo/ps.go
```diff
@@ -12,8 +12,8 @@ import (
...
+	Short: "List containers on the target for the current Compose project.",
```
Luke-Parkin, line 15:
```suggestion
	Short: "List running containers on the target for the current Compose project. (use -a to show all)",
```
or something like that.

The point being that it's not clear it only shows running without -a.

Luke-Parkin (approved):
LGTM, just a minor wording change

cmd/topo/ps.go
awphi: I think detailing options is a bit much for short descriptions and "use -a to show all" also could be misleading since users may think it means show all containers on the target, even outside their current project.

WDYT about adding more detail to the desc. of the `-a` flag? e.g.

```
❯ docker ps -h                    
Flag shorthand -h has been deprecated, use --help
Usage:  docker ps [OPTIONS]

List containers

Aliases:
  docker container ls, docker container list, docker container ps, docker ps

Options:
  -a, --all             Show all containers (default shows just running)
```

```
❯ docker compose ps -h
Flag shorthand -h has been deprecated, use --help
Usage:  docker compose ps [OPTIONS] [SERVICE...]

List containers

Options:
  -a, --all                  Show all stopped containers (including those created by the run command)
```

th3james (approved):
Looks good, thanks for ditching that superfluous "Running"

Luke-Parkin (approved):

PR #323, opened 2026-07-07
feat: Support selecting compose files with `-f`

muchzill4:
## Changes

- Add `-f`/`--file` compose file selection to `topo deploy`, `configure`, `ps`, and `stop`.
- Centralise compose file lookup and validation in `internal/compose`.
- Preserve default lookup order of `compose.yaml`, then `compose.yml`, with a warning when both exist (similar to how `docker compose` behaves).

### DIFF
--- a/cmd/topo/root.go
+++ b/cmd/topo/root.go
@@ -27,8 +27,6 @@ var rootCmd = &cobra.Command{
 	},
 }
 
-const composeFileFlag = "file"
-
 func init() {
 	rootCmd.PersistentFlags().StringP(
 		"output",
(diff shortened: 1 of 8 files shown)


cmd/topo/root.go
```diff
@@ -35,6 +38,32 @@ func init() {
...
+	)
+}
+
+func getComposeFileName(cmd *cobra.Command) (string, error) {
+	flag := cmd.Flag(composeFileFlag)
+	if flag == nil {
+		panic(fmt.Sprintf("internal error: compose file flag not registered: %s", composeFileFlag))
+	}
+
+	if flag.Changed {
```
Luke-Parkin, line 56: TIL.

flag.Changed is a bool; where if it's set (to anything other than it's default) it's true?

Luke-Parkin (approved):
LGTM! Thanks :)

PR #324, opened 2026-07-08
fix: show full compose file path in deploy message

federicobozzini:
## Changes

- when the compose file has no default deployment message the `topo ps` command suggested always points to compose.yaml or compose.yml. This PR uses the full compose file when the `-f` option is used for deployment.

### DIFF
--- a/internal/deploy/deploy.go
+++ b/internal/deploy/deploy.go
@@ -1,6 +1,9 @@
 package deploy
 
 import (
+	"fmt"
+	"strings"
+
 	"github.com/arm/topo/internal/deploy/command"
 	"github.com/arm/topo/internal/deploy/operation"
 	"github.com/arm/topo/internal/deploy/post_deploy"
@@ -54,6 +57,26 @@ func NewDeployment(composeFile string, opts DeployOptions) (goperation.Sequence,
 		}
 	}
 	ops = append(ops, operation.NewDockerComposeUp(composeFile, targetHost, opts.RecreateMode))
-	ops = append(ops, post_deploy.NewDeploySuccess(composeFile, targetHost, "Run `topo ps` to see deployed containers"))
+	ops = append(ops, post_deploy.NewDeploySuccess(composeFile, targetHost, defaultDeploySuccessMessage(composeFile)))
 	return goperation.NewSequence(ops...), cleanup
+}
+
+func defaultDeploySuccessMessage(composeFile string) string {
+	if composeFile == "compose.yaml" {
+		return "Run `topo ps` to see deployed containers"
+	}
+
+	return fmt.Sprintf("Run `topo ps -f %s` to see deployed containers", shellQuote(composeFile))
+}
+
+func shellQuote(arg string) string {
+	if arg == "" {
+		return "''"
+	}
+
+	if strings.ContainsAny(arg, " \t\n\"'\\$`") {
+		return "'" + strings.ReplaceAll(arg, "'", `'"'"'`) + "'"
+	}
+
+	return arg
 }
--- a/internal/deploy/deploy.go
+++ b/internal/deploy/deploy.go
@@ -1,9 +1,6 @@
 package deploy
 
 import (
-	"fmt"
-	"strings"
-
 	"github.com/arm/topo/internal/deploy/command"
 	"github.com/arm/topo/internal/deploy/operation"
 	"github.com/arm/topo/internal/deploy/post_deploy"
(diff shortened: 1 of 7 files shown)


internal/deploy/deploy_test.go
```diff
@@ -38,6 +38,22 @@ func TestNewDeployment(t *testing.T) {
 		assert.Equal(t, want, got)
 	})
 
+	t.Run("includes compose file flag in default success message for non-default compose file", func(t *testing.T) {
```
muchzill4: We should test this via `post_deploy.NewDeploySuccess`, not via `deploy.NewDeployment`. The `NewDeploySuccess` could contain the logic about messaging.

Alternatively, have `post_deploy.DefaultMessage(composeFile)`.

internal/deploy/deploy.go
```diff
@@ -54,6 +57,26 @@ func NewDeployment(composeFile string, opts DeployOptions) (goperation.Sequence,
...
+
+func defaultDeploySuccessMessage(composeFile string) string {
+	if composeFile == "compose.yaml" {
+		return "Run `topo ps` to see deployed containers"
+	}
+
+	return fmt.Sprintf("Run `topo ps -f %s` to see deployed containers", shellQuote(composeFile))
+}
+
+func shellQuote(arg string) string {
```
muchzill4: Is this necessary? Where is it tested?

internal/deploy/deploy.go
```diff
@@ -54,6 +57,26 @@ func NewDeployment(composeFile string, opts DeployOptions) (goperation.Sequence,
 		}
 	}
 	ops = append(ops, operation.NewDockerComposeUp(composeFile, targetHost, opts.RecreateMode))
-	ops = append(ops, post_deploy.NewDeploySuccess(composeFile, targetHost, "Run `topo ps` to see deployed containers"))
+	ops = append(ops, post_deploy.NewDeploySuccess(composeFile, targetHost, defaultDeploySuccessMessage(composeFile)))
 	return goperation.NewSequence(ops...), cleanup
 }
+
+func defaultDeploySuccessMessage(composeFile string) string {
+	if composeFile == "compose.yaml" {
```
muchzill4: What about `compose.yml`? Please use/expose relevant functionality from `internal/compose/file.go` (suggestion: `compose.IsDefaultFile()`)

muchzill4 (changes requested):

internal/deploy/deploy.go
federicobozzini: If you have both a `compose.yaml` and a ` compose.yml` file in the workspace and you deploy the `compose.yml`,  the default `topo ps` would still use the `compose.yaml`. Now the message will (unnecessarily) show `topo ps -f compose.yml` in the case where a `compose.yaml` is not present, but I think it's still better than adding too much complexity to check if it's really necessary.

internal/deploy/deploy.go
federicobozzini: This is in case a file name contains a space, eg: `my compose.yaml`. I'm afraid that's a valid case :cry: I can add a test for it.

internal/deploy/deploy.go
muchzill4: It's trickier than that, because if you have `compose.yml`, but no `compose.yaml` topo will use that (remember that it looks for `compose.yaml` then `compose.yml` if it can't find the former).

But also, if there's both `compose.yaml` and `compose.yml`, you can do `topo ... -f compose.yml`, in which case the `topo ps` also have to be explicit about the file.

internal/deploy/deploy_test.go
federicobozzini: Good point,  moved. https://github.com/arm/topo/pull/324/commits/cce56f1cf7075ce571b921c67c89bcffa06ee7b9

internal/deploy/post_deploy/post_deploy.go
```diff
@@ -23,6 +24,26 @@ func NewDeploySuccess(composeFile string, h command.Host, defaultMessage string)
...
+		return "Run `topo ps` to see deployed containers"
+	}
+
+	return fmt.Sprintf("Run `topo ps -f %s` to see deployed containers", shellQuote(composeFile))
+}
+
+func shellQuote(arg string) string {
+	if arg == "" {
+		return "''"
+	}
```
federicobozzini: I'm personally not a fan of this but I can't find a better way to address it.

internal/deploy/deploy.go
federicobozzini: ...That's what I wrote :sweat_smile:

internal/deploy/deploy.go
federicobozzini: Done here https://github.com/arm/topo/commit/cce56f1cf7075ce571b921c67c89bcffa06ee7b9

internal/deploy/deploy.go
muchzill4: Welp, so you did. I'd still like the `"compose.yaml"` being _the default_ to come from `internal/compose/file.go` somehow. Hardcoding a string in deployment implicitly couples compose file resolution and deployment. Latter shouldn't know about how compose file is resolved.

internal/deploy/post_deploy/post_deploy.go
```diff
@@ -23,6 +24,26 @@ func NewDeploySuccess(composeFile string, h command.Host, defaultMessage string)
...
 
+func DefaultMessage(composeFile string) string {
+	if composeFile == "compose.yaml" {
+		return "Run `topo ps` to see deployed containers"
+	}
+
+	return fmt.Sprintf("Run `topo ps -f %s` to see deployed containers", shellQuote(composeFile))
+}
+
+func shellQuote(arg string) string {
```
muchzill4: `internal/command/command.go` sounds like a better home for this.

Perhaps `func QuoteArg(s string) string {`?

muchzill4:
Should this be `fix:`?

internal/deploy/post_deploy/post_deploy.go
federicobozzini: https://github.com/arm/topo/pull/324/commits/e4c4ca6f7da57587d8ecc10720bfdd09d164edc2

internal/deploy/deploy.go
federicobozzini: https://github.com/arm/topo/pull/324/commits/70655604999165cdcd086e5fb3bbacc4c20e8fa5

muchzill4 (approved):
Nice!

PR #327, opened 2026-07-09
chore: upgrade go to fix govuln GO-2026-5856

awphi:
## Changes

- Does as says on tin - see govulncheck failure: https://github.com/arm/topo/actions/runs/28958321823/job/86073420531?pr=326

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/arm/topo
 
-go 1.26.4
+go 1.26.5
 
 require (
 	github.com/compose-spec/compose-go/v2 v2.13.0


federicobozzini (approved):
Should we also update the development docs to make it clear we want version 1.26.5?

awphi:
> Should we also update the development docs to make it clear we want version 1.26.5?

Iirc `go.sum` is authoritative and will refuse to compile the project with the wrong version so fine as is imo. Can adjust separately if we feel strongly. I like one source of truth personally so would rather just reference `go.sum` from development.md.

PR #328, opened 2026-07-09
feat: rename templates to projects

muchzill4:
## Summary

Renames the user-facing Topo “Templates” concept to “Topo Projects” across the CLI, README, glossary, catalog, output views, tests, and catalog update tooling.

This keeps `topo templates` as a compatibility alias for `topo projects`, but moves the primary command, documentation, schema, and public catalog/output types to project terminology.

## Changes

- Rename `topo templates` to `topo projects`
- Update clone, extend, README, glossary, command output, and test wording to use Topo Project/project terminology
- Rename catalog schema/catalog JSON from `templates` to `projects`
- Rename catalog and output view types from template-oriented names to project-oriented names where practical
- Rename catalog update tooling from `scripts/update_templates` to `scripts/update_projects`
- Update required-parameter error text away from “build arguments”

## Follow-up

Some internal code names intentionally remain for a later cleanup PR to keep this change from ballooning:

- Consider merging `internal/template` with `topo/project` and use project-oriented names internally
- Consider renaming `arguments.Arg` / `ResolvedArg` to clearer parameter-oriented names, likely `Parameter` / `ResolvedParameter`


internal/template/definition.go
th3james: What's the rationale behind not updating all of these internal values at the same time?
```suggestion
type Project struct {
```

th3james: even if we're keeping the internal struct names, this should be updated, right?
```suggestion
		return Template{}, fmt.Errorf("failed to decode project: %w", err)
```

muchzill4: I went ahead and did the balooning rename. Turns out it wasn’t so bad.

th3james (approved):

PR #329, opened 2026-07-10
docs: clarify dependency on OpenSSH

federicobozzini:
## Changes

- `topo` actually depends on OpenSSH, not any generic ssh implementation. Let's make this explicit.


th3james:
This is a bit misleading in its current form. It suggests that we've validated that OpenSSH is installed, but we've actually only validated that `ssh` is on the PATH. If we're doing this, we should probably test for OpenSSH, not just SSH

federicobozzini:
> This is a bit misleading in its current form. It suggests that we've validated that OpenSSH is installed, but we've actually only validated that `ssh` is on the PATH. If we're doing this, we should probably test for OpenSSH, not just SSH

I agree but this would misleading in the same way saying docker is OK is misleading when its version is not supported. At the moment we only check for binaries and `ssh` is the right binary for OpenSSH. We can replace the check with something better but I would treat it as a separate issue.

th3james:
@federicobozzini That's a false equivalence. We show:
```
Container Engine: ✅ (docker)
```
That's a true statement.

This change strengthens the health check claim without actually strengthening the evidence.
```
OpenSSH: ✅ (ssh)
```
I think we risk misleading users by making a stronger claim than what we're actually checking ("don't bother checking if OpenSSH is the issue, you've got it").

I think only this change is reasonable if coupled with an upgrade in the check.

federicobozzini:
@th3james Change made here 297303359f7dd8209bbe49aa8543e2c9246c62f5

This makes this PR way bigger because OpenSSH prints its output to stderr. At the moment stdout and stderr are put together, this PR keeps them separate which I think it's more appropriate.

internal/install/install_test.go
```diff
@@ -0,0 +1,31 @@
...
+	binaries := map[string][]byte{"topo": []byte("binary")}
+	r := &runner.Fake{Commands: map[string]runner.FakeResult{
+		"install -D -m 0755 /dev/stdin /unwritable/topo": {
+			Stderr: "install: cannot create regular file '/unwritable/topo': Permission denied",
+			Err:    errors.New("exit status 1"),
+		},
+		"install -D -m 0755 /dev/stdin /writable/topo": {},
+	}}
+
+	gotPath, gotBinaries, err := installToFirstWriteableDir(paths, r, binaries)
```
th3james: This tests a private function, which is against the conventions in the codebase. Looks like the module under test is a bit of an iceberg, and is probably due a refactor. I'd suggest dropping this test from this PR and fixing the tests in a separate PR.

internal/health/dependencies.go
```diff
@@ -41,6 +42,19 @@ type Dependency struct {
...
 }
 
+type OpenSSHAvailable struct{}
+
+func (o OpenSSHAvailable) Run(ctx context.Context, r runner.Runner, dep Dependency) (*Fix, error) {
+	stdout, stderr, err := r.Run(ctx, "ssh -V")
+	if err != nil {
+		return nil, err
+	}
+	if !strings.Contains(stdout, "OpenSSH_") && !strings.Contains(stderr, "OpenSSH_") {
```
th3james: Do we actually need to check stdout?

internal/health/dependencies.go
```diff
@@ -41,6 +42,19 @@ type Dependency struct {
...
+		return nil, fmt.Errorf("%q is not OpenSSH", dep.Binary)
```
th3james: We should print what actually resolved, and return a fix
```suggestion
	version := strings.TrimSpace(stdout + stderr)
	if !strings.Contains(version, "OpenSSH_") {
		return &Fix{
			Description: "Install OpenSSH and ensure its ssh executable is first on PATH",
		}, fmt.Errorf("%q does not resolve to OpenSSH: %s", dep.Binary, version)
```

federicobozzini: Changed.

internal/health/dependencies.go
federicobozzini: Not really, removed.

internal/install/install_test.go
federicobozzini: Removed.

th3james (approved):

PR #330, opened 2026-07-10
chore: move all CLI concerns under a top-level `cli/` dir

awphi:
## Background

As agreed offline, plan is to migrate the project specification and a new docs site into this repository. As part of this I propose moving the cli into its own subdirectory `cli/` to keep the root clean and avoid favouring it over the other parts of the repo. i.e. I propose a root like:

```
cli/
docs/
project-specification/
README.md
LICENSE
CONTRIBUTING.md
```

I'm open to this being the wrong approach and we could instead just chuck the docs and spec into the existing file structure:

```
internal/
cmd/
docs/
project-specification/
go.sum
go.mod
// etc...
README.md
LICENSE
CONTRIBUTING.md
```

Lmk what you think!

## Changes

- Moves all CLI concerns under `cli/`
  - Updates all go github import paths to account for this
  - Updates github catalog schema URLs to account for this
  - Does not move the install scripts as, despite being CLI-specific, they are references heavily in public-facing docs like learning paths
  - Updates CI to account for this
- Generifies the language in CONTRIBUTING.md to avoid calling this project 'Topo CLI'
- Generifies the existing `docs/index.md`. Although doesn't matter so much since it'll soon go the way of the dodo.

I would recommend reviewing this commit-by-commit. They're all pretty atomic and the only big one is the import path update which I did last.

## Next steps

- [x] Move all current cli concerns under `cli/`, taking care not to blow up external consumers (homebrew, internal sites, release process etc)
- [ ] Update `arm/homebrew-topo` to build from the new location of the cli (PR here: https://github.com/arm/homebrew-topo/pull/6)
- [ ] Bring in a copy of `arm/topo-project-specification` to this repo under `project-specification/`. Update link in `arm/topo` to point to that.
- [ ] Nuke `arm/topo-project-specification`. Leave it up as a redirect prompt only for a while as stop-gap if we miss a link.
- [ ] Update all existing example projects (previously templates) links to point to new home for the spec under `arm/topo`
- [ ] Scaffold a docs site that can be previewed locally. Do not wire up any deployment quite yet.
- [ ] Wire up internal deployment of the docs. Deployed at the same time of CLI release **after signing**
- [ ] Write some proper docs!


muchzill4:
Should `project-specification` be part of docs? If so, perhaps we could retain the existing directory structure and _just_ have:

```
cmd/*
internal/*
docs/*.md
docs/project-specification/*.md
```

Having said that, I’m not opposed to the change you’re proposing. 👍

awphi:
> Should `project-specification` be part of docs? If so, perhaps we could retain the existing directory structure and _just_ have:
> 
> ```
> cmd/*
> internal/*
> docs/*.md
> docs/project-specification/*.md
> ```
> 
> Having said that, I’m not opposed to the change you’re proposing. 👍

Mm interesting thought - how do you think we handle the other bits of the schema repo, namely the skills and JSON schema?

muchzill4:
> Mm interesting thought - how do you think we handle the other bits of the schema repo, namely the skills and JSON schema?

What’s the difference between them living in the `<root>/project-specification` and `<root>/docs/project-specification`? 🤔

awphi:
> > Mm interesting thought - how do you think we handle the other bits of the schema repo, namely the skills and JSON schema?
> 
> What’s the difference between them living in the `<root>/project-specification` and `<root>/docs/project-specification`? 🤔

The only thing I can think that potentially matters is the skills. If we go with `<root>/docs/project-specification` then installing them with `npx skills` would be `npx skills add arm/topo/docs/project-specification` which is a tad odd. 

WDYT about ditching this PR and going for:

```
cmd/*
internal/*
docs/*.md
docs/project-specification/*.md
skills/**/*.md
```

I also think this gives a clear path to adding skill installation from the CLI which is a nice bonus.

PR #333, opened 2026-07-10
feat: support legacy `x-topo.args` with a warning

muchzill4:
## Changes

- Add backwards-compatible parsing for deprecated `x-topo.args` metadata
- Prefer `x-topo.parameters` when present, including when it aliases legacy `args`
- Emit a deprecation warning only when falling back to `x-topo.args`


th3james (approved):
Very nice

PR #334, opened 2026-07-10
docs: Clarify required value is an SSH Destination

th3james:
## Changes

- Make it clearer what `target` _is_, rather than just yelling that it hasn't been provided

### DIFF
--- a/cmd/topo/root.go
+++ b/cmd/topo/root.go
@@ -94,7 +94,7 @@ func lookupTarget(cmd *cobra.Command) (string, bool) {
 func requireTarget(cmd *cobra.Command) (string, error) {
 	t, exists := lookupTarget(cmd)
 	if !exists {
-		return "", fmt.Errorf("target not specified: provide an SSH destination (for example, user@host) using --target or TOPO_TARGET")
+		return "", fmt.Errorf("target not specified: set --target or %s to the SSH destination Topo should connect to (e.g. user@example.local)", targetEnvVar)
 	}
 	return t, nil
 }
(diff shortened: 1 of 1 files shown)


cmd/topo/root.go
```diff
@@ -94,7 +94,7 @@ func lookupTarget(cmd *cobra.Command) (string, bool) {
 func requireTarget(cmd *cobra.Command) (string, error) {
 	t, exists := lookupTarget(cmd)
 	if !exists {
-		return "", fmt.Errorf("target not specified: provide --target or set TOPO_TARGET env var")
+		return "", fmt.Errorf("target not specified: set --target or %s to the SSH destination Topo should connect to (e.g. user@example.local)", targetEnvVar)
```
federicobozzini: Are we happy to use `connect` here? It might suggest some statefulness.

federicobozzini (approved):
Minor question but looks good to me.

cmd/topo/root.go
muchzill4: Not a fan of this. I think the correct place to explain the purpose of arguments is `--help`/usage. We shouldn’t duplicate that information whenever a command fails because an argument wasn't supplied.
Try `git clone` w/o repo to clone. Or bare `docker run`.

If we keep it, perhaps make it more _conversational_: `”where to deploy to? set —target or TARGET_ENV”`

The description of `—target` in usage could definitely be improved though w/ the `e.g.` you’ve added here.

muchzill4: Actually, I rescind my comment. I think this is amazing for agentic stuff. We should look for other opportunities to apply this pattern. 👍

th3james:
@muchzill4 @federicobozzini Realised we were missing the obvious answer here - a short, concise error message, and a revamp of our rather neglected root command description.

muchzill4 (approved):

PR #335, opened 2026-07-13
feat: improve remote port errors

yejseo01:
## Changes

  - Replaced curl with native TCP checks.
  - Added specific DNS, timeout, network, and exposure errors.
  - Added --skip-remote-port-check.
  - Suggests skipping only when the check is inconclusive.
  - Added coverage for the new behavior.
  - Support skipping of the port check through ENV variable


cmd/topo/deploy.go
```diff
@@ -129,11 +145,23 @@ func resolvePort(cmd *cobra.Command, flagValue string) (string, error) {
...
+	if !cmd.Flags().Changed("skip-remote-port-check") {
+		return env.IsVarTruthy(skipRemotePortCheckEnvVar)
+	}
+
+	value, err := cmd.Flags().GetBool("skip-remote-port-check")
+	if err != nil {
+		panic("internal error: skip-remote-port-check flag not registered: " + err.Error())
+	}
+	return value
+}
```
awphi, line 155: nit: Could remove the panic here if you pass the flag value in as we do for `resolvePort`

```go
func resolveSkipRemotePortCheck(cmd *cobra.Command, flagValue string) bool {
	if cmd.Flags().Changed("skip-remote-port-check") {
		return flagValue
	}

	return env.IsVarTruthy(skipRemotePortCheckEnvVar)
}
```

either that or use `.Bool` instead of `.BoolVar` when registering the flag like we do in `health`

internal/deploy/deploy_test.go
```diff
@@ -148,6 +148,38 @@ func TestNewDeployment(t *testing.T) {
...
+		}
+		want = append(want, operation.NewRunRegistry(port)...)
+		want = append(want,
+			wantTunnelStart,
+			operation.NewRegistryTransfer(composeFile, localHost, remoteHost, port),
+			wantTunnelStop,
+			operation.NewDockerComposeUp(composeFile, remoteHost, operation.RecreateModeDefault),
+			post_deploy.NewDeploySuccess(composeFile, remoteHost, "Run `topo ps` to see deployed containers"),
+		)
+		assert.Equal(t, want, got)
```
awphi: Can we just assert the operations sequence doesn't contain the port check instead of asserting it does contain everything else?

internal/deploy/deploy_test.go
```diff
@@ -148,6 +148,38 @@ func TestNewDeployment(t *testing.T) {
...
+
+	t.Run("excludes the remote port check when skipped", func(t *testing.T) {
+		remoteDest := ssh.NewDestination("user@remote")
+		port := operation.DefaultRegistryPort
+		opts := deploy.DeployOptions{
+			TargetHost: remoteDest,
+			Registry: &deploy.RegistryConfig{
+				Port:                port,
+				SkipRemotePortCheck: true,
+				UseControlSockets:   true,
```
awphi:
```suggestion
				SkipRemotePortCheck: true,
```

Would avoid specifying anything that isn't relevant to the dest and letting it fall back to defaults

internal/ssh/config.go
```diff
@@ -22,6 +22,19 @@ func NewConfig(dest Destination) Config {
 	return NewConfigFromBytes(output)
 }
 
+func ResolveHostName(dest Destination) (string, error) {
```
awphi:
```suggestion
func resolveHostName(dest Destination) (string, error) {
```

nit: not used outside package

internal/ssh/config.go
```diff
@@ -22,6 +22,19 @@ func NewConfig(dest Destination) Config {
...
+	output, err := readConfig(dest)
+	if err != nil {
+		return "", fmt.Errorf("could not resolve SSH hostname for %q: %w", dest.String(), err)
+	}
```
awphi, line 29:
```suggestion
	output, err := readConfig(dest)
	if err != nil {
		return "", fmt.Errorf("could not resolve SSH configuration for %q: %w", dest.String(), err)
	}
```

nit: would make this error more relevant to what we actually tried to do which was call `ssh -G`

internal/ssh/ssh_tunnel.go
```diff
@@ -90,6 +94,18 @@ type CheckRemoteForwardNotExposed struct {
...
+	err error
+}
+
+func (e *inconclusiveRemotePortCheckError) Error() string {
+	return e.err.Error()
+}
+
+func (e *inconclusiveRemotePortCheckError) Unwrap() error {
+	return e.err
+}
```
awphi: This struct seems redundant - `Error` just forwards to `err.Error()` and `Unwrap` seems unused. Can we just handle the `error` value directly instead of wrapping it?

internal/ssh/ssh_tunnel.go
```diff
@@ -106,27 +122,64 @@ func (ct *CheckRemoteForwardNotExposed) Run(w io.Writer) error {
...
+	if err := checkRemotePortNotListening(host, ct.Port); err != nil {
+		return remotePortCheckErrorWithSuggestion(err, ct.Port)
+	}
+	_, _ = fmt.Fprintf(w, "Port %s is bound to remote loopback only\n", ct.Port)
+	return nil
+}
+
+func checkRemotePortNotListening(host, port string) error {
+	address := net.JoinHostPort(host, port)
+	connection, err := net.DialTimeout("tcp", address, 5*time.Second)
```
awphi: I was under the impression we couldn't use this because it doesn't resolve hostnames in all circumstance properly? If we can then great and we should update the README/docs to drop the `curl` dep!

My robotic friends tell me something that seems to ring a bell on this
>  - Release binaries use CGO_ENABLED=0 (.goreleaser.yaml:13), so Unix builds use Go’s resolver rather than the native resolver used by
   OpenSSH/curl. This can differ for mDNS/.local, custom NSS sources, and some split-DNS/VPN configurations.

cc: @muchzill4

internal/ssh/ssh_tunnel.go
```diff
@@ -106,27 +122,64 @@ func (ct *CheckRemoteForwardNotExposed) Run(w io.Writer) error {
...
+	if errors.As(err, &networkError) && networkError.Timeout() {
+		return &inconclusiveRemotePortCheckError{err: fmt.Errorf("timed out while checking whether remote port %s is exposed: %w", address, err)}
+	}
+
+	var dnsError *net.DNSError
+	if errors.As(err, &dnsError) {
+		return &inconclusiveRemotePortCheckError{err: fmt.Errorf("could not resolve remote host %q while checking tunnel exposure: %w", host, err)}
+	}
+
+	return &inconclusiveRemotePortCheckError{err: fmt.Errorf("could not verify whether remote port %s is exposed: %w", address, err)}
```
awphi: Should we provide guidance on what to do in these cases too? Feels like these should come with similar wrapping text along the lines of "cannot conclusively rule out network access to registry port..." with the specific error of _why_ we couldn't conclusively rule out network access embedded

internal/ssh/ssh_tunnel.go
```diff
@@ -106,27 +122,64 @@ func (ct *CheckRemoteForwardNotExposed) Run(w io.Writer) error {
...
-	cmd := exec.Command("curl", fmt.Sprintf("%s:%s", host, port), "--max-time", "5")
-	cmd.Stdout = io.Discard
-	cmd.Stderr = io.Discard
-	_ = cmd.Run()
+func isConnectionRefused(err error) bool {
+	if runtime.GOOS == "windows" {
+		const windowsConnectionRefused syscall.Errno = 10061
+		return errors.Is(err, windowsConnectionRefused)
+	}
+	return errors.Is(err, syscall.ECONNREFUSED)
```
awphi: `syscall.ECONNREFUSED` should be defined on windows too - is that not the right errno? Would like to avoid hard-coding platform-specific syscall err numbers in our codebase and leave that up to the stdlib

internal/ssh/ssh_tunnel.go
```diff
@@ -106,27 +122,64 @@ func (ct *CheckRemoteForwardNotExposed) Run(w io.Writer) error {
...
+	return nil
+}
+
+func checkRemotePortNotListening(host, port string) error {
+	address := net.JoinHostPort(host, port)
+	connection, err := net.DialTimeout("tcp", address, 5*time.Second)
+	if err == nil {
+		if closeErr := connection.Close(); closeErr != nil {
+			return fmt.Errorf("remote port %s accepted a TCP connection, but closing the connection failed: %w", address, closeErr)
+		}
```
awphi: Is this worth bailing out over? Perhaps should just be a warning? We tend to ignore most resource acquisition `Close` errors

internal/ssh/ssh_tunnel.go
```diff
@@ -106,27 +122,64 @@ func (ct *CheckRemoteForwardNotExposed) Run(w io.Writer) error {
...
+		return fmt.Errorf("remote sshd might be exposing the forwarded port %s on its network (likely GatewayPorts=yes); the local registry may be reachable without SSH auth", port)
```
awphi: Two nits on this:
- It might not be `sshd`, could be dropbear or whatever
- We might still want to tell them how to fix this i.e. pass the flag or fix the ssh server config

internal/ssh/ssh_tunnel.go
yejseo01: I think we resolve it using line 127 at internal/ssh/config.go which use exec to fetch the config. So I suspect it would be fine.

I think the original curl was used because back then, before the big refactoring, it was a big pain to resolve the host from alias?

yejseo01: Ah you're right. we can't move away from curl. Will make it again with curl

internal/ssh/ssh_tunnel.go
yejseo01: this error will be wrapped with `remotePortCheckErrorWithSuggestion` up the stack.

cmd/topo/deploy.go
yejseo01: done - [c37a7f3](https://github.com/arm/topo/pull/335/commits/c37a7f390a181ccf4f0e3141cde4e326fd3fd909)

internal/deploy/deploy_test.go
yejseo01: done - [3cd98ed](https://github.com/arm/topo/pull/335/commits/3cd98ed115187131e2bfec23bc5f6491b7741d95)

internal/deploy/deploy_test.go
yejseo01: done - [d8fc012](https://github.com/arm/topo/pull/335/commits/d8fc01283ec284232cc9e994ad2e59593412e5f4)

internal/ssh/config.go
yejseo01: done - [d8fc012](https://github.com/arm/topo/pull/335/commits/d8fc01283ec284232cc9e994ad2e59593412e5f4)

internal/ssh/config.go
yejseo01: done - [d8fc012](https://github.com/arm/topo/pull/335/commits/d8fc01283ec284232cc9e994ad2e59593412e5f4)

internal/ssh/ssh_tunnel.go
yejseo01: done - [17349c5](https://github.com/arm/topo/pull/335/commits/17349c5b8d0d0b9c099898e2ba94a789762244ea)

internal/ssh/ssh_tunnel.go
yejseo01: this bit of the code is removed in this change - [17349c5](https://github.com/arm/topo/pull/335/commits/17349c5b8d0d0b9c099898e2ba94a789762244ea)

internal/ssh/ssh_tunnel.go
yejseo01: this bit of the code is removed - [17349c5](https://github.com/arm/topo/pull/335/commits/17349c5b8d0d0b9c099898e2ba94a789762244ea)

internal/ssh/ssh_tunnel.go
yejseo01: updated the messages here - [5149407](https://github.com/arm/topo/pull/335/commits/51494073bc41a070a88600f1f8e2036faf14c622)

cmd/topo/deploy.go
awphi: As discussed offline, the new code doesn't seem to check the flag value

internal/ssh/ssh_tunnel.go
```diff
@@ -106,27 +109,48 @@ func (ct *CheckRemoteForwardNotExposed) Run(w io.Writer) error {
...
 	}
-	return fmt.Errorf("remote sshd might be exposing the forwarded port %s on its network (likely GatewayPorts=yes); the local registry may be reachable without SSH auth", ct.Port)
+	_, _ = fmt.Fprintf(w, "Port %s is bound to remote loopback only\n", ct.Port)
+	return nil
 }
 
-func RemotePortRefusedConnection(host, port string) bool {
-	cmd := exec.Command("curl", fmt.Sprintf("%s:%s", host, port), "--max-time", "5")
-	cmd.Stdout = io.Discard
+func checkRemotePortNotListening(host, port string) (error, bool) {
```
awphi: The `bool` return seems useless, only seems be false when error is nil and true and when error is non-nil. This will make the error wrapping that happens on call site clearer too since atm it looks like
```go
	err, inconclusive := checkRemotePortNotListening(host, ct.Port)
	if inconclusive {
		return remotePortCheckErrorWithSuggestion(err, ct.Port)
	}
	if err != nil {
		return err
	}
```
which is a bit misleading since we always return `inconclusive=true` when returning an error so it can just be
```go
	err := checkRemotePortNotListening(host, ct.Port)
	if err != nil {
		return remotePortCheckErrorWithSuggestion(err, ct.Port)
	}
```
---
bonus idea: rename to `ensureRemotePortNotListening` to make it clear that error = we couldn't tell, nil = it's definitely not listening

internal/ssh/ssh_tunnel.go
```diff
@@ -106,27 +109,48 @@ func (ct *CheckRemoteForwardNotExposed) Run(w io.Writer) error {
...
+	if errors.As(err, &exitError) && exitError.ExitCode() == 7 {
+		return nil, false
 	}
-	return false
+	if errors.As(err, &exitError) && exitError.ExitCode() == 28 {
+		return fmt.Errorf("timed out while checking whether remote port %s is exposed: %w", address, err), true
+	}
+	if errors.As(err, &exitError) && exitError.ExitCode() == 6 {
+		return fmt.Errorf("could not resolve remote host %q while checking tunnel exposure: %w", host, err), true
+	}
```
awphi: nit/style: could avoid repeated casting. not opinionated on this so feel free to ignore if you prefer the current style


```suggestion
	if exitError, ok := errors.AsType[*exec.ExitError](err); ok {
		switch exitError.ExitCode() {
		case 7:
			return nil, false
		case 28:
			return fmt.Errorf("timed out while checking whether remote port %s is exposed: %w", address, err), true
		case 6:
			return fmt.Errorf("could not resolve remote host %q while checking tunnel exposure: %w", host, err), true
		}
	}
```

internal/ssh/ssh_tunnel.go
```diff
@@ -106,27 +109,48 @@ func (ct *CheckRemoteForwardNotExposed) Run(w io.Writer) error {
...
 }
 
-func RemotePortRefusedConnection(host, port string) bool {
-	cmd := exec.Command("curl", fmt.Sprintf("%s:%s", host, port), "--max-time", "5")
-	cmd.Stdout = io.Discard
+func checkRemotePortNotListening(host, port string) (error, bool) {
+	address := net.JoinHostPort(host, port)
+	var remoteIP strings.Builder
+	cmd := exec.Command("curl", "http://"+address, "--max-time", "5", "--noproxy", "*", "--output", os.DevNull, "--silent", "--write-out", "%{remote_ip}")
+	cmd.Stdout = &remoteIP
```
awphi, line 135: Think we could do with a comment that covers a couple questions future us (and current me!) may have with this:
- Why we're using curl
- Why we're using some of the more non-standard args like `--write-out ...` and `--noproxy *`

Also is `--silent` needed? We discard stderr at the Go cmd wrapper level

cmd/topo/deploy.go
yejseo01: sorry! done here now! - [8a986e1](https://github.com/arm/topo/pull/335/commits/8a986e1da46632b54d41b3daead8e1aad6a34998)

internal/ssh/ssh_tunnel.go
yejseo01: I'll do something in between

internal/ssh/ssh_tunnel.go
yejseo01: done - [0b413ab](https://github.com/arm/topo/pull/335/commits/0b413abeb3aeb5d69631c97105c948bc3b946fb5)

internal/ssh/ssh_tunnel.go
yejseo01: I did this - [1ab175e](https://github.com/arm/topo/pull/335/commits/1ab175e75977cd11edda55a7278dde918be5fe1f)****

internal/ssh/ssh_tunnel.go
yejseo01: good point - [d4ad836](https://github.com/arm/topo/pull/335/commits/d4ad8367f41e0c254d8fa3fe170314683a9d1260)

internal/ssh/ssh_tunnel.go
```diff
@@ -106,27 +109,64 @@ func (ct *CheckRemoteForwardNotExposed) Run(w io.Writer) error {
...
 
-	// Only curl exit 7 ("Failed to connect to host") proves no TCP listener
-	// answered. Anything else — connection succeeded, DNS failure, timeout,
-	// curl missing — leaves us unable to certify the port is not not listening.
-	if cmd.ProcessState != nil && cmd.ProcessState.ExitCode() == 7 {
-		return true
+	var exitError *exec.ExitError
+	if !errors.As(err, &exitError) {
+		return fmt.Errorf("could not verify whether remote port %s is exposed: %w", address, err)
+	}
```
awphi: nit: dislike this error being duplicated here in this guard clause and in the `return` at the bottom of the function, would rather split the curl `exec.ExitError` error code classification out and gate that with an `errors.AsType` but stylistic and minor so 🤷

internal/ssh/ssh_tunnel.go
```diff
@@ -106,27 +109,64 @@ func (ct *CheckRemoteForwardNotExposed) Run(w io.Writer) error {
...
 	}
-	return fmt.Errorf("remote sshd might be exposing the forwarded port %s on its network (likely GatewayPorts=yes); the local registry may be reachable without SSH auth", ct.Port)
+	_, _ = fmt.Fprintf(w, "Port %s is bound to remote loopback only\n", ct.Port)
+	return nil
 }
 
-func RemotePortRefusedConnection(host, port string) bool {
-	cmd := exec.Command("curl", fmt.Sprintf("%s:%s", host, port), "--max-time", "5")
-	cmd.Stdout = io.Discard
+func ensureRemotePortNotListening(host, port string) error {
```
awphi: I now realize I completely missed the main part of this function in my earlier review. So afaict we have 3 return states:
- Definitely not listening (curl exit code 7)
- Definitely listening (curl didn't error or len(stdout) > 0)
- Inconclusive (curl errored with non-7 known or unknown exit code)

I think capturing these three states with nil, a special type of error and a regular error respectively isn't super clear (and I know this was my suggestion so apologies!).

WDYT about reverting to two return values that represent the states more explicitly?

```go
func isRemotePortListening(host, port string) (listening bool, error) {
  // ...
}
```

bool answers the question "could I connect to the port?" error answers the question "did anything go wrong while attempting to do so?". Then the call site becomes:

```go
	listening, err := isRemotePortListening(host, ct.Port)
	if listening {
		return fmt.Sprintf("the remote SSH server is exposing forwarded registry port %s beyond remote loopback; configure the SSH server to bind remote forwards to loopback only, or use `--skip-remote-port-check` if you understand that the registry may be reachable without SSH authentication", err.port)
	}
	if err != nil {
		return remotePortCheckErrorWithSuggestion(err, ct.Port)
	}
	_, _ = fmt.Fprintf(w, "Port %s is bound to remote loopback only\n", ct.Port)
```

avoids the special error struct that callers have to know to check which might be easier? Anyway I've bikeshedded this function for long enough so will approve and leave up to you!

awphi (approved):
Food for thought but not blocking!

internal/ssh/ssh_tunnel.go
yejseo01: [f1a501f](https://github.com/arm/topo/pull/335/commits/f1a501fc20eabe3fdfba3450870ac540b1dc53e8)

internal/ssh/ssh_tunnel.go
yejseo01: [f1a501f](https://github.com/arm/topo/pull/335/commits/f1a501fc20eabe3fdfba3450870ac540b1dc53e8)

PR #336, opened 2026-07-14
docs: structure docs for public docs site, add basic introduction section

awphi:
## Changes

- Creates a new `introduction/` subdir for intro ducmentation. Overviews, glossary, core concepts, getting started guides etc. 
  - Adds a `README.md` to this file to give an overview of 'Topo' overall. Not specific to the CLI, extension or product specification but lists all 3 of those as next steps to check out. 
    - I'm favouring `README.md` over `index.md` as it allows GH to render it when looking at the the directory it lives in. Not sure if this is the right approach longer term if we opt for more MDX but works for now.
    - This page also includes a diagram of what Topo is and the workflow we expect it to be used in. I'm no designer but this is a first pass.
  - Adds a `getting-started.mdx` which is mostly a re-hash of the CLI README get started guide with a tiny bit of visual flair
- Fixes some broken links

## Next steps
- [ ] Proper homepage
- [ ] Add a 'Quick start' to the 'Overview' page(?)
- [ ] Core concepts page(?) - this might be covered by the overview and glossary
- [ ] Troubleshooting page
- [ ] Dedicated 'Integrations' page to point to VS Code extension(?)
- [ ] CLI reference page
- [ ] Tutorials + learning path links

---

Previewing the docs site is internal only for now so reach out if you'd like to so and I can point you in the right direction. In the next couple of weeks we should be able to publish instructions on previewing changes locally in the repo.


yejseo01:
> ## Changes
> * Creates a new `introduction/` subdir for intro **ducmentation**. Overviews, glossary, core concepts, getting started guides etc.

🦆

docs/introduction/getting-started.mdx
```diff
@@ -0,0 +1,147 @@
...
+# Getting started
+
+Use this guide to deploy the [Hello World Topo Project](https://github.com/Arm-Examples/topo-welcome) to a target that runs Linux on AArch64 (`linux/arm64`). The project runs a web page with a configurable greeting.
+
+## Prerequisites
+
+Install the following tools on the host where you run `topo`:
+
+- [Docker](https://docs.docker.com/get-docker/)
+- [`curl`](https://curl.se/download.html)
```
yejseo01: why is only this code font?

docs/introduction/getting-started.mdx
```diff
@@ -0,0 +1,147 @@
...
+- [Git](https://git-scm.com/downloads)
+- OpenSSH client
```
yejseo01: Should we link this one to anything like others?

docs/introduction/getting-started.mdx
```diff
@@ -0,0 +1,147 @@
...
+
+If the target host key is unknown, confirm the target identity. Then run the suggested command to accept the key. If public-key authentication is not configured, run the following command after you accept the host key:
+
+```sh
+topo setup-keys --target pi@raspberrypi
+```
+
+This command creates a dedicated key, copies the public key to the target, and updates your SSH configuration. The target must permit password authentication for the initial key transfer. If it does not, configure a public key through your existing device management process.
+
+Run `topo health` again and confirm that no checks report an error. You can ignore a Remoteproc Runtime warning because the Hello World project does not require this runtime.
```
yejseo01: At this point of the document, the reader doesn't know if we are running Hello World project or not. it seems out of context to mention Hello World project here.

Can we just rephrase it to "You can ignore a Remoteproc Runtime warning if you are not planning to run a project that uses a remoteproc-runtime"?

awphi: The start of the guide says 

> Use this guide to deploy the [Hello World Topo Project](https://github.com/Arm-Examples/topo-welcome) to a target that runs Linux on AArch64 (`linux/arm64`). The project runs a web page with a configurable greeting.

Happy to generify the guide to not suggest a specific template if you think that's better?

docs/introduction/getting-started.mdx
awphi: Wanted to capture that it's the binary not `libcurl` or anything else. Have made this section match the README for consistency - https://github.com/arm/topo/pull/336/commits/75dff3319547b59ee2c9f02b92977f838fc472d4

docs/introduction/getting-started.mdx
awphi: Tbh the curl, git and an OpenSSH download link are kind of pointless - your host system almost certainly has all of them. If you don't then [OpenSSH is installed differently on each platform since MS re-implemented it so that's why there's no link there](https://github.com/PowerShell/Win32-OpenSSH/wiki/Install-Win32-OpenSSH) :/

Have simplified the list to match the top-level README to avoid the argument https://github.com/arm/topo/pull/336/commits/75dff3319547b59ee2c9f02b92977f838fc472d4. The simpler list hasn't let us down so far so should be fine!

docs/introduction/getting-started.mdx
yejseo01: Ah I see it now

yejseo01 (approved):
LGTM!

awphi:
Resetting approval as got a good suggestion from @Luke-Parkin to split out the installation instructions into their own doc so other project that use Topo as their deployment mechanism have a good doc to link to.

Have also generified the language in the getting started guide to avoid pinning it to `topo-welcome`.

Luke-Parkin (approved):
Thanks Adam, this is a great start for us to write eveeeen more docs yay!

Maybe the next step would be some sort of CI or maybe just a tickbox to make people remember to update

awphi:
Merging as-is for now to give us a base to iterate on

PR #339, opened 2026-07-15
fix: make `upgrade` unavailable, not hidden when appropriate

muchzill4:
## Changes

For installs managed by homebrew, we hide the `upgrade` command, however it’s still possible to invoke it. This PR makes it actually unavailable instead of hidden.

### DIFF
--- a/cmd/topo/upgrade.go
+++ b/cmd/topo/upgrade.go
@@ -47,8 +47,8 @@ var upgradeCmd = &cobra.Command{
 
 func init() {
 	binPath, err := upgrade.CurrentBinaryPath()
-	if err == nil {
-		upgradeCmd.Hidden = !upgrade.IsBinaryManagedByUs(binPath)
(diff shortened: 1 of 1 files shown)


th3james (approved):

cmd/topo/upgrade.go
```diff
@@ -47,8 +47,8 @@ var upgradeCmd = &cobra.Command{
 
 func init() {
 	binPath, err := upgrade.CurrentBinaryPath()
-	if err == nil {
-		upgradeCmd.Hidden = !upgrade.IsBinaryManagedByUs(binPath)
+	if err == nil && !upgrade.IsBinaryManagedByUs(binPath) {
```
th3james, line 50: Outside of the scope of this change, but shouldn't we handle `err` being non nil with a panic or something?

muchzill4: Probably. 🤔

PR #340, opened 2026-07-16
ci: deploy docs to staging on push to main

awphi:
## Changes

- Does as it says on the tin. Just following instructions from https://github.com/ARM-software/docs-action


yejseo01 (approved):
as we discussed online

PR #341, opened 2026-07-16
feat: control ability to self-upgrade with `TOPO_DISABLE_SELF_UPGRADE`

muchzill4:
## Background

First party integrations that ship with their own `topo` installation want to disable self-upgrade facility. One example of that is `vscode-topo`, where user can inadvertently upgrade topo version from vscode’s terminal. One way to prevent that would be to use similar mechanism to the one we have already - detect install path and disable self upgrade if it’s not managed by topo cli itself. For vscode however, this would mean leaking vscode-specific knowledge into topo, which is not ideal. This PR disables `upgrade` sub-command based on the environment variable.

## Changes

- Disable `topo upgrade` when `TOPO_DISABLE_SELF_UPGRADE` is set to truthy value

### DIFF
--- a/cmd/topo/upgrade.go
+++ b/cmd/topo/upgrade.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/arm/topo/internal/env"
 	"github.com/arm/topo/internal/output/term"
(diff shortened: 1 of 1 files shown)


yejseo01 (approved):
LGTM

PR #347, opened 2026-07-17
docs: add troubleshooting page

awphi:
## Changes

- Adds a troubleshooting docs page for common issues I could find by trawling jira/confluence + from my memory
- Links to said page from other parts of the introduction section, namely the overview and getting started guide

## Checklist

### DIFF
--- a/docs/introduction/troubleshooting.mdx
+++ b/docs/introduction/troubleshooting.mdx
@@ -1,0 +1,330 @@
+import Tabs from "@theme/Tabs";
+import TabItem from "@theme/TabItem";
+
+# Troubleshooting
+
+:::note
+
+The commands on this page use `<user@host>` as the target. Replace this value with the SSH destination for your target.
+
+:::
+
+## Start with a health report
+
+Run the health check before you deploy:
+
+```sh
+topo health --target <user@host>
+```
+
+Use `localhost` only when the host is also the deployment target:
+
+```sh
+topo health --target localhost
+```
+
+For logs or automation, capture the JSON report:
+
+```sh
+topo health --target <user@host> --output json
+```
+
+Check every reported `status`. `topo health` can return exit status `0` when a health check reports an error. Do not rely only on the process exit status.
+
+## Fix installation problems
+
+### `topo: command not found`
+
+Choose the method that you used to install Topo.
+
+<Tabs groupId="install-method">
+  <TabItem value="brew" label="Homebrew">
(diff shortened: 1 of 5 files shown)


docs/introduction/troubleshooting.mdx
```diff
@@ -0,0 +1,318 @@
...
+
+```sh
+topo health --target <user@host>
+```
+
+Use `localhost` only when the host is also the deployment target:
+
+```sh
+topo health --target localhost
+```
```
muchzill4, line 24: Is it worth saying this? Perhaps we can tweak the `:::note` block to say something along the lines of:

> Replace this value with a valid SSH destination for your target. It also be `localhost`.

docs/introduction/troubleshooting.mdx
```diff
@@ -0,0 +1,318 @@
...
+
+For logs or automation, capture the JSON report:
+
+```sh
+topo health --target <user@host> --output json
+```
+
+Check every reported `status`. `topo health` can return exit status `0` when a health check reports an error. Do not rely only on the process exit status.
+
+## Fix installation problems
```
muchzill4, line 34: Not convinced of the value of this entire section. Seems outside of our responsibility.

docs/introduction/troubleshooting.mdx
```diff
@@ -0,0 +1,318 @@
...
+
+The command creates a key, transfers the public key, and writes a Topo-managed SSH configuration with `IdentitiesOnly yes`.
+
+If Topo reports legacy SSH configuration entries, migrate them first:
+
+```sh
+topo migrate-ssh
+```
+
+If an SSH alias is already associated with another user, create a separate alias for each user in `~/.ssh/config`.
```
muchzill4, line 191: All of this would ideally come naturally when using `topo` commands. Errors returned in the cli should guide the user, if they don’t we should fix that instead of documenting. WDYT?

muchzill4 (commented):
I wonder if we shouldn’t be more selective with what issues we’re documenting. Ideally a lot of this would be covered by `topo` guiding users to fix the issues. Only the things that are hard to CLI-ify should be here.

awphi:
As discussed offline, think most of what is covered here is outside our remit or covered by error messages in the CLI. Can re-visit a troubleshooting doc later when we are aware of more issues that cannot be easily covered by guided error messages in the tool itself.

PR #348, opened 2026-07-17
docs: breaking changes policy

awphi:
## Changes

- Adds a breaking changes policy document, links to it from the general development guide.
  - The content of this policy is relatively opinionated and is informed by my ideas of where to draw the line on breaking changes. Please throw many stones at it so we can get something we're all happy with!
- Fixes the casing of 'Development guide' to use sentence case as encouraged by the Arm technical writing style guide as I was editing that file.
- Puts both these files under a 'Contributing' section

Screenshot 2026-07-17 at 13 31 44

## Checklist

### DIFF
--- a/docs/development/breaking-changes.md
+++ b/docs/development/breaking-changes.md
@@ -1,0 +1,121 @@
+# Breaking change policy
+
+This policy defines which changes to the Topo are breaking changes for end users. It helps reviewers classify changes consistently before a release.
+
+This policy explicitly does not extend to:
+
+- Hidden commands and flags
+- Experimental features
+- Human-readable output, help text, warnings, and logs
+- Source code APIs and packages
+- Build tools and developer scripts
+- Undocumented or malformed input that the binary happened to accept
+
+A hidden or experimental interface becomes part of the compatibility contract when it is promoted to a documented, supported interface.
+
+## CLI compatibility
+
+The following CLI changes are breaking:
+
+- Removing or renaming of a supported command, flag, positional argument, or environment variable consitutes a breaking change.
+- Making an optional flag or environment variable required
+
+The following CLI changes are not breaking:
+
+- Adding a command, flag, positional argument, or environment variable without changing existing behavior
+- Renaming an interface while retaining the old form as a functional alias
+- Changing help text or completion descriptions
+
+## JSON compatibility
+
+JSON structure is contractual. JSON values are not contractual unless their allowed values are explicitly documented as a closed set.
+
+The following JSON changes are breaking:
+
+- Removing or renaming a field
+- Changing a field type, nullability/optionality or nesting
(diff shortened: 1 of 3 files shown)


docs/development/breaking-changes.md
```diff
@@ -0,0 +1,121 @@
...
+
+The following CLI changes are not breaking:
+
+- Adding a command, flag, positional argument, or environment variable without changing existing behavior
+- Renaming an interface while retaining the old form as a functional alias
+- Changing help text or completion descriptions
+
+## JSON compatibility
+
+JSON structure is contractual. JSON values are not contractual unless their allowed values are explicitly documented as a closed set.
```
awphi, line 31: I talk about 'closed sets' here as we have at least one JSON output value users should reasonably depend on atm which is the `status` prop of `topo health` dependency checks.

```
      {
        "name": "OpenSSH",
        "status": "ok", // <-- this guy
        "value": "ssh"
      },
```

Plan is to document this in the long help text of `topo health` in a separate PR so it's beholden to this part of the policy

muchzill4 (approved):
Nice!

docs/development/breaking-changes.md
```diff
@@ -0,0 +1,121 @@
+# Breaking change policy
+
+This policy defines which changes to the Topo are breaking changes for end users. It helps reviewers classify changes consistently before a release.
```
federicobozzini: Out of curiosity what does `for end users` mean here? I mean, as opposed to what?

docs/development/breaking-changes.md
```diff
@@ -0,0 +1,121 @@
...
+- Changing help text or completion descriptions
+
+## JSON compatibility
+
+JSON structure is contractual. JSON values are not contractual unless their allowed values are explicitly documented as a closed set.
+
+The following JSON changes are breaking:
+
+- Removing or renaming a field
+- Changing a field type, nullability/optionality or nesting
```
federicobozzini: Should a conversion from nullable to non-nullable be breaking? Technically it's a valid covariant refinement.

federicobozzini (commented):
Nice!

docs/development/breaking-changes.md
awphi: Basically just getting at what the paragraph below clarifies which is we only changes that affect the end-user as potentially breaking, not internal scripts, source code etc.

I'll remove this since I think it adds nothing https://github.com/arm/topo/pull/348/commits/7f7a0d2fe5137752fbedea2b02e37ec700d5c288

docs/development/breaking-changes.md
awphi: https://github.com/arm/topo/pull/348/commits/20f28d13f5ff60b92c9ec89306b3b7f9f22224eb good point!

PR #350, opened 2026-07-17
refactor: model SSH tunnels as resources

muchzill4:
## Changes

- Introduce a resource-based SSH tunnel with explicit open and close lifecycle methods.
- Keep process ownership, cleanup, (future) context handling, and command output together in the resource.
- Move the remaining deployment-specific tunnel operation adapters into `internal/deploy/operation`, preserving registry-specific error guidance.
- Remove brittle SSH operation value tests and deployment cleanup-operation assertions.

This is a first step toward removing Topo’s operation abstraction. Deployment orchestration has pushed that abstraction beyond what it models cleanly: passing state between operations is awkward, conditional control flow and resource lifecycles are difficult to reason about, and comparing operation sequences by value provides limited confidence in deployment behavior.

## Checklist


internal/ssh/ssh_tunnel_unix_test.go
```diff
@@ -0,0 +1,141 @@
...
+		secondCloseErr := tunnel.Close(context.Background(), nil)
+
+		require.NoError(t, firstCloseErr)
+		require.NoError(t, secondCloseErr)
+		invocations := readSSHInvocations(t, logPath, 2)
+		assert.Len(t, invocations, 2)
+	})
+}
+
+func installFakeSSH(t *testing.T) string {
```
awphi, line 87: Not sure about this as a mean to test `ssh` invocations because it excludes windows and that's mostly where we expect `useControlSockets` to be set to false and is worth testing explicitly since it's a different OpenSSH impl. that no one on the team uses daily.

WDYT about injecting something to capture the calls into `OpenSSHTunnel` in Go land? Might also need an alternate solution for pid capturing.

muchzill4: Discussed offline and (hopefully) agreed that this coverage is a bit better than what we had.

- Uses native go’s way to kill a process (win)
- Keeps testing command strings (meh, but coverage)
- Doesn’t cover windows (it’s the same execution path, so covered by unix bit?)

awphi (approved):
As discussed offline, coverage is better with new tests if anything and actually testing the tunnel start/stop is covered by e2e tests on all platforms

PR #354, opened 2026-07-21
ci: tag deleted on fail

yejseo01:
## Changes

- remove the release tag from repo if the release failed
- use `gh run watch` instead of polling to keep an eye of publisher's failure

## Checklist

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -164,45 +164,16 @@ jobs:
         env:
           GH_TOKEN: ${{ steps.publisher-token.outputs.token }}
           PUBLISHER_RUN_ID: ${{ steps.trigger_sign_and_release.outputs.publisher_run_id }}
-          ARTIFACT_NAME: release-links
         run: |
-          echo "Waiting for $ARTIFACT_NAME artifact from topo-publisher run $PUBLISHER_RUN_ID"
+          gh run watch "$PUBLISHER_RUN_ID" \
+            --repo Arm-Debug/topo-publisher \
+            --compact \
+            --exit-status
 
-          for attempt in {1..30}; do
-            artifact_id="$(gh api \
-              "repos/Arm-Debug/topo-publisher/actions/runs/$PUBLISHER_RUN_ID/artifacts" \
-              --jq ".artifacts | map(select(.name == \"$ARTIFACT_NAME\")) | .[0].id // empty")"
-
-            if [ -n "$artifact_id" ]; then
-              echo "Found $ARTIFACT_NAME artifact: $artifact_id"
-              gh run download "$PUBLISHER_RUN_ID" \
-                --repo Arm-Debug/topo-publisher \
-                --name "$ARTIFACT_NAME" \
-                --dir publisher-artifacts
-              exit $?
-            fi
-
-            status="$(gh run view "$PUBLISHER_RUN_ID" \
-              --repo Arm-Debug/topo-publisher \
-              --json status \
-              --jq .status)"
-
-            conclusion="$(gh run view "$PUBLISHER_RUN_ID" \
-              --repo Arm-Debug/topo-publisher \
-              --json conclusion \
-              --jq '.conclusion // ""')"
-
-            if [ "$status" = "completed" ]; then
-              echo "::error::topo-publisher run completed with conclusion '$conclusion' before producing $ARTIFACT_NAME"
-              exit 1
-            fi
-
-            echo "$ARTIFACT_NAME is not available yet; publisher run status is $status; retrying in 60 seconds ($attempt/30)"
-            sleep 60
-          done
-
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -164,45 +165,16 @@ jobs:
...
-
-            if [ "$status" = "completed" ]; then
-              echo "::error::topo-publisher run completed with conclusion '$conclusion' before producing $ARTIFACT_NAME"
-              exit 1
-            fi
-
-            echo "$ARTIFACT_NAME is not available yet; publisher run status is $status; retrying in 60 seconds ($attempt/30)"
-            sleep 60
-          done
+          gh run watch "$PUBLISHER_RUN_ID" \
```
muchzill4, line 172: `gh run watch` has 3s default interval, we previously slept for 60s. Do we want to keep the long sleep or not?

.github/workflows/release.yaml
```diff
@@ -226,3 +198,9 @@ jobs:
           else
             gh release create "$RELEASE_TAG" --title "$RELEASE_TAG" --notes-file combined_notes.md
           fi
+
+      - name: Delete tag after failure
+        if: ${{ (failure() || cancelled()) && steps.create_tag.outcome == 'success' }}
```
muchzill4: Is this going to run if CI times out? I.e.: `gh watch` never succeeds, as it seems to wait indefinitely.

muchzill4 (commented):
Looks like a nice simplification. 👍

.github/workflows/release.yaml
```diff
@@ -226,3 +198,9 @@ jobs:
           else
             gh release create "$RELEASE_TAG" --title "$RELEASE_TAG" --notes-file combined_notes.md
           fi
+
+      - name: Delete tag after failure
```
awphi: Do we have to create the tag before running the publisher? 

Imo it would be nice, if possible, to only create the tag after the flakiest step(s) to avoid creating a tag at all when it does fail

.github/workflows/release.yaml
yejseo01: it's probably better to keep the interval

.github/workflows/release.yaml
yejseo01: I'm afraid goreleaser requires you to make release on generated tags?
```
Run goreleaser/goreleaser-action@f06c13b6b1a9625abc9e6e439d9c05a8f2190e94
  with:
    distribution: goreleaser
    version: ~> v2
    args: release --clean --skip=publish
    workdir: .
    install-only: false
  env:
    GOTOOLCHAIN: local
    GITHUB_TOKEN: ***
    GORELEASER_CURRENT_TAG: v8.0.1-pre.2
  
Downloading https://github.com/goreleaser/goreleaser/releases/download/v2.17.0/goreleaser_Linux_arm64.tar.gz
Downloading https://github.com/goreleaser/goreleaser/releases/download/v2.17.0/checksums.txt
Checksum verified for goreleaser_Linux_arm64.tar.gz
cosign not found in PATH, skipping signature verification
Extracting GoReleaser
/usr/bin/tar xz --warning=no-unknown-keyword --overwrite -C /home/runner/work/_temp/a9a635bd-af56-4973-b1ab-bc03b9f03a3e -f /home/runner/work/_temp/c1e7820a-9cf7-4db3-9073-3a24fc5e4513
GoReleaser ~> v2 installed successfully
/opt/hostedtoolcache/goreleaser-action/2.17.0/arm64/goreleaser release --clean --skip=publish
  • starting release
  • skipping announce and publish...
  • cleaning distribution directory
  • loading environment variables
    • using token from $GITHUB_TOKEN
  • getting and validating git state
    • couldn't find any tags before "v8.0.1-pre.2"
    • using tags                                     previous=<unknown> current=v8.0.1-pre.2
  ⨯ release failed after 0s                          error=git tag v8.0.1-pre.2 was not made against commit 78ec37d9798b503e2520fe53d06c5f87ef8791af
```

.github/workflows/release.yaml
yejseo01: it won't. good point. I made cleanup into a separate job so that it will cleanup on timeouts - [9cf4653](https://github.com/arm/topo/pull/354/commits/9cf4653bd7ba1e870247183c8c3c1016ea482494)

.github/workflows/release.yaml
yejseo01: done - [6350ca2](https://github.com/arm/topo/pull/354/commits/6350ca295c8647a9d9460f7ac078e255e71e2787)

.github/workflows/release.yaml
awphi: Argh okay, thanks for checking!

muchzill4 (approved):

PR #356, opened 2026-07-21
feat!: add docker compose health check

awphi:
## Changes

- Adds a `docker compose version` probe to the health check that asserts:
  - Docker compose is usable as a docker CLI plugin with `docker compose` syntax
  - Docker compose is of the minimum required version. I found this to be [2.21.0](https://github.com/docker/compose/releases?page=8#release-v2.21.0) which changes the `docker compose ps` output format to the one we expect

## Implementation

- Avoids use of `docker compose --version` since docker CLI ignores invalid subcommands and parses the `--version` flag first i.e.:
```
❯ docker foobar --version          
Docker version 29.6.1, build 8900f1d330
```
- Pulls semver parsing out of the `upgrade` internals, exposes a public func for the check to use. Also makes semver parsing tolerate the lackof a `v` as it was dropped from the `docker-compose version` output at some point
```
adahod01@lima-default:~$ docker compose version
Docker Compose version v2.21.0
```
```
❯ docker compose version
Docker Compose version 5.2.0
```
- Robust to a dead daemon since `docker compose version` does not rely on it. We also predicate the check on the container engine health check.

## Future podman concerns

Podman allows us to force a certain compose provider with an env var, thus never needing to assert anything about the 'Docker Compose version ...' bit of the stdout e.g.:
```
❯ PODMAN_COMPOSE_PROVIDER=docker-compose podman compose version 2>/dev/null
Docker Compose version 5.2.0
```
This usually means `docker-compose` will be on the path. This doesn't really matter to us - we will only care if `podman compose version` gives the expected output. 

## Checklist

### DIFF
--- a/internal/health/check.go
+++ b/internal/health/check.go
@@ -109,8 +109,8 @@ func (c DockerComposeCompatible) Run(ctx context.Context, r runner.Runner, _ Dep
 
 	if !version.IsAtLeastVersion(stdout, c.MinVersion) {
 		return &Fix{
-			Description: fmt.Sprintf("Upgrade Docker Compose to at least version %s", c.MinVersion),
-		}, fmt.Errorf("installed docker compose version %q is less than required version %q", stdout, c.MinVersion)
+			Description: fmt.Sprintf("Upgrade Docker Compose to version %s or later", c.MinVersion),
+		}, fmt.Errorf("installed docker compose version %q is older than required version %q", stdout, c.MinVersion)
 	}
 
 	return nil, nil
--- a/internal/health/check.go
+++ b/internal/health/check.go
@@ -2,6 +2,7 @@ package health
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"slices"
@@ -95,22 +96,28 @@ func (o OpenSSHAvailable) Run(ctx context.Context, r runner.Runner, dep Dependen
 	return nil, nil
 }
 
-type DockerComposeCompatible struct {
+type DockerComposeMinVersion struct {
 	MinVersion string
 }
 
-func (c DockerComposeCompatible) Run(ctx context.Context, r runner.Runner, _ Dependency) (*Fix, error) {
-	stdout, _, err := r.Run(ctx, "docker compose version")
+func (c DockerComposeMinVersion) Run(ctx context.Context, r runner.Runner, _ Dependency) (*Fix, error) {
+	stdout, _, err := r.Run(ctx, "docker compose version --format json")
 	if err != nil {
-		return &Fix{
-			Description: "Ensure Docker Compose is installed as a plugin for Docker",
-		}, errors.New(strings.ReplaceAll(err.Error(), "\n", " "))
+		return nil, err
 	}
 
-	if !version.IsAtLeastVersion(stdout, c.MinVersion) {
+	var output struct {
+		Version string `json:"version"`
+	}
+	err = json.Unmarshal([]byte(stdout), &output)
+	if err != nil {
+		return nil, err
+	}
+
+	if !version.IsAtLeastVersion(output.Version, c.MinVersion) {
 		return &Fix{
 			Description: fmt.Sprintf("Upgrade Docker Compose to version %s or later", c.MinVersion),
-		}, fmt.Errorf("installed docker compose version %q is older than required version %q", stdout, c.MinVersion)
+		}, fmt.Errorf("installed docker compose version %s is older than required version %s", output.Version, c.MinVersion)
 	}
 
 	return nil, nil
--- a/internal/health/check.go
+++ b/internal/health/check.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/arm/topo/internal/output/logger"
 	"github.com/arm/topo/internal/runner"
+	"github.com/arm/topo/internal/version"
 )
 
 type Check interface {
@@ -91,6 +92,27 @@ func (o OpenSSHAvailable) Run(ctx context.Context, r runner.Runner, dep Dependen
 			Description: "Install OpenSSH and ensure its ssh executable is first on PATH",
 		}, fmt.Errorf("%q does not resolve to OpenSSH: %s", dep.Binary, stderr)
 	}
+	return nil, nil
+}
+
+type DockerComposeCompatible struct {
+	MinVersion string
+}
+
+func (c DockerComposeCompatible) Run(ctx context.Context, r runner.Runner, _ Dependency) (*Fix, error) {
+	stdout, _, err := r.Run(ctx, "docker compose version")
+	if err != nil {
+		return &Fix{
+			Description: "Ensure Docker Compose is installed as a plugin for Docker",
+		}, errors.New(strings.ReplaceAll(err.Error(), "\n", " "))
+	}
+
+	if !version.IsAtLeastVersion(stdout, c.MinVersion) {
+		return &Fix{
+			Description: fmt.Sprintf("Upgrade Docker Compose to at least version %s", c.MinVersion),
+		}, fmt.Errorf("installed docker compose version %q is less than required version %q", stdout, c.MinVersion)
+	}
+
 	return nil, nil
 }
 
--- a/internal/health/check_test.go
+++ b/internal/health/check_test.go
@@ -212,7 +212,7 @@ Run 'docker --help' for more information`
 		fix, err := check.Run(ctx, runner, dep)
 
 		assert.Contains(t, err.Error(), "1.9.0")
-		assert.Contains(t, err.Error(), "less than required version")
+		assert.Contains(t, err.Error(), "older than required version")
 		assert.Contains(t, fix.Description, "Upgrade Docker Compose")
 	})
 }
--- a/internal/health/check_test.go
+++ b/internal/health/check_test.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"strings"
 	"testing"
 
 	"github.com/arm/topo/internal/health"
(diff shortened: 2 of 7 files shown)


internal/health/check.go
```diff
@@ -78,6 +80,27 @@ func (v VersionMatches) Run(ctx context.Context, _ runner.Runner, _ Dependency)
...
+type DockerComposeCompatible struct {
+	MinVersion string
+}
+
+func (c DockerComposeCompatible) Run(ctx context.Context, r runner.Runner, _ Dependency) (*Fix, error) {
+	stdout, _, err := r.Run(ctx, "docker compose version")
+	if err != nil {
+		return &Fix{
+			Description: "Ensure Docker Compose is installed as a plugin for Docker",
+		}, errors.New(strings.ReplaceAll(err.Error(), "\n", " "))
```
muchzill4: Is this `\n` replacement tested anywhere? 🤔 What is this guarding against?

internal/health/check.go
```diff
@@ -78,6 +80,27 @@ func (v VersionMatches) Run(ctx context.Context, _ runner.Runner, _ Dependency)
...
+	}
+
+	if !version.IsAtLeastVersion(stdout, c.MinVersion) {
+		return &Fix{
+			Description: fmt.Sprintf("Upgrade Docker Compose to at least version %s", c.MinVersion),
```
muchzill4: "Upgrade Docker Compose to version %s or later”

internal/health/check.go
```diff
@@ -78,6 +80,27 @@ func (v VersionMatches) Run(ctx context.Context, _ runner.Runner, _ Dependency)
...
+		}, fmt.Errorf("installed docker compose version %q is less than required version %q", stdout, c.MinVersion)
```
muchzill4:
```suggestion
		}, fmt.Errorf("installed docker compose version %q is older than required version %q", stdout, c.MinVersion)
```

internal/health/check.go
```diff
@@ -78,6 +80,27 @@ func (v VersionMatches) Run(ctx context.Context, _ runner.Runner, _ Dependency)
 	return &fix, InfoError{Err: fmt.Errorf("out of date - current: %s, latest version: %s", v.CurrentVersion, latest)}
 }
 
+type DockerComposeCompatible struct {
```
muchzill4: Why `*Compatible`? Shouldn’t it be `DockerComposeMinVersion`?

internal/health/dependencies.go
```diff
@@ -114,6 +114,16 @@ var HostRequiredDependencies = []Dependency{
 			},
 		},
 	},
+	{
+		Binary:                "docker-compose",
+		Label:                 "Docker Compose",
+		SoftwarePrerequisites: []SoftwareDependency{Docker},
+		Checks: []Check{
+			DockerComposeCompatible{
```
muchzill4: A bit sad we can’t use `VersionMatches` for this, but I think going specific instead of generic is the right call. Perhaps we should also make `VersionMatches` a bit more explicit about being Topo oriented.

internal/health/check.go
awphi: The thought was because it also checks that it's installed as a plugin and not just checking the standalone binary `docker-compose`, though arguably that's implicit or should be split into separate checks?

internal/health/dependencies.go
awphi: Agree. I kind of think the topo version check is a different thing to the other health checks and is a bit shoe-horned in at the moment. For example, with the topo version check if we fail to get the latest from artifactory we actively don't want to bail out, unlike any other version matching health check.

I'll make a note to de-generify it in name and implementation in a follow up if that's okay?

internal/health/check.go
awphi: It's not guarding per-se but trying to make the output slightly less ugly when compose isn't installed as a plugin - without it the `topo health` output looks like:

```
Host
----
Topo: ✅ (topo)
OpenSSH: ✅ (ssh)
Curl: ✅ (curl)
Container Engine: ✅ (docker)
Docker Compose: ❌ (local command failed: exit status 1 | stderr: docker: unknown command: docker compose

Run 'docker --help' for more information
)
  Fix:
    Ensure Docker Compose is installed as a plugin for Docker
```

with it:

```
Host
----
Topo: ✅ (topo)
OpenSSH: ✅ (ssh)
Curl: ✅ (curl)
Container Engine: ✅ (docker)
Docker Compose: ❌ (local command failed: exit status 1 | stderr: docker: unknown command: docker compose  Run 'docker --help' for more information )
  Fix:
    Ensure Docker Compose is installed as a plugin for Docker
```

Did also consider trying to classify the stderr and present a nice hand-written error message but not sure it's worth the effort? Added a test for current behaviour here; https://github.com/arm/topo/pull/356/commits/c307516568dddc78138859134036cc70d6d55b4f

internal/health/check.go
awphi: https://github.com/arm/topo/pull/356/commits/488abd0fbbbde9647bbaa3ae7dc30a99e141a605

internal/health/check.go
awphi: https://github.com/arm/topo/pull/356/commits/488abd0fbbbde9647bbaa3ae7dc30a99e141a605

internal/health/check.go
muchzill4: Thanks for explaining. I wonder if we should re-think how we present the stderr. While I agree that single line works better in this case, it also isn’t ideal due to its length. Ideally we’d have something akin to:

```
Host
_____
…
Docker Compose: ❌
    Executed: docker compose version
    Stderr:
        docker: unknown command: docker compose
        
        Run 'docker --help' for more information
    Fix:
        Ensure Docker Compose is installed as a plugin for Docker     
```

Of course this is likely too big for this PR.

internal/health/check.go
muchzill4: I see. Two checks mean running `docker compose version` twice, unless we introduce caching, which would be a bit sad.

internal/health/check_test.go
```diff
@@ -89,3 +91,85 @@ func TestVersionMatches(t *testing.T) {
...
+			Commands: map[string]runner.FakeResult{
+				"docker compose version": {
+					Err: errors.New(stderr),
+				},
+			},
+		}
+
+		_, err := check.Run(ctx, runner, dep)
+
+		assert.Contains(t, err.Error(), strings.ReplaceAll(stderr, "\n", " "))
```
muchzill4: In general, we should use `assert.ErrorContains` because it’ll do `require.Error(t, err)` first, in case `err` is nil.

But in this case, we could actually do:

```
assert.EqualError(t, err, strings.ReplaceAll(stderr, "\n", " "))
```

internal/version/semver_test.go
```diff
@@ -0,0 +1,31 @@
+package version
```
muchzill4:
```suggestion
package version_test
```

But also needs importing `version`.

awphi: https://github.com/arm/topo/pull/356/commits/ada021dba9cdbeafe4f6e79c85c8db746102b4cc

internal/health/check_test.go
awphi: https://github.com/arm/topo/pull/356/commits/ada021dba9cdbeafe4f6e79c85c8db746102b4cc

internal/health/check.go
awphi: Agree, posting for posterity as discussed offline:
> Thinking about whether we should show exit codes/stderr interleaved with the health report at all or if it's only useful for advanced users. Perhaps an idea is to ditch it completely and show it in logs before the health report is rendered.

Example of the idea:
```
10:24:35 ERR error executing host health probe command="docker compose version" stdout="" err="local command failed: exit status 1 | stderr: docker: unknown command: docker compose\n\nRun 'docker --help' for more information\n"stderr="docker: unknown command: docker compose\n\nRun 'docker --help' for more information\n"

Host
----
Topo: ✅ (topo)
OpenSSH: ✅ (ssh)
Curl: ✅ (curl)
Container Engine: ✅ (docker)
Docker Compose: ❌ (failed to execute: `docker compose version`)
  Fix:
    Ensure Docker Compose is installed as a plugin for Docker
...
```

Not 100% on this, there may just be a better way to render stderr 🤔 Potentially some color in the human output would go a long way here

internal/health/check_test.go
```diff
@@ -89,3 +91,85 @@ func TestVersionMatches(t *testing.T) {
...
+			Commands: map[string]runner.FakeResult{
+				"docker compose version": {
+					Err: fmt.Errorf("it blew up"),
+				},
+			},
+		}
+
+		fix, err := check.Run(ctx, runner, dep)
+
+		assert.Contains(t, err.Error(), "it blew up")
```
muchzill4:
```suggestion
		assert.EqualError(t, err, "it blew up")
```

internal/health/check_test.go
```diff
@@ -89,3 +91,85 @@ func TestVersionMatches(t *testing.T) {
...
+		runner := &runner.Fake{
+			Commands: map[string]runner.FakeResult{
+				"docker compose version": {Output: "Docker Compose version 1.9.0"},
+			},
+		}
+
+		fix, err := check.Run(ctx, runner, dep)
+
+		assert.Contains(t, err.Error(), "1.9.0")
+		assert.Contains(t, err.Error(), "older than required version")
```
muchzill4: Should we use `EqualError` here too or is this too inflexible and hardcode-y?

muchzill4 (approved):
I realised you can use `docker compose version --format json` and avoid trying to regex match the version in the default, pretty output.

awphi:
> I realised you can use docker compose version --format json and avoid trying to regex match the version in the default, pretty output.

It does give you a significantly less useful stderr if you don't have compose installed. On macOS it gives:
```
❯ docker compose version --format json                                                           
unknown flag: --format

Usage:  docker [OPTIONS] COMMAND [ARG...]

Run 'docker --help' for more information
```

On Linux it with docker 26.1.5 it seemed to blat the entire contents of `docker --help` into stderr 😢

internal/health/check_test.go
awphi: https://github.com/arm/topo/pull/356/commits/1d884b9a7b9ebae45229fc29c450eb68f30f29c8

internal/health/check_test.go
awphi: https://github.com/arm/topo/pull/356/commits/1d884b9a7b9ebae45229fc29c450eb68f30f29c8

awphi:
@muchzill4 Have separated the "is docker compose installed as a plugin?" check from the "is it of the minimum version check?" as we discussed here - https://github.com/arm/topo/pull/356/commits/1d55e2145ec7a10b01320cf40e07779329b1bbe4

Definitely feels neater. Downside is since we're now using `CommandSuccessful` to run `docker compose` we've lost the multi-line error joining unless we want fiddle with that check too. I'm personally okay with losing it given we want to make general improvements to health output - thoughts?

muchzill4 (approved):
LGTM

PR #357, opened 2026-07-21
fix: suppress login shell startup output

federicobozzini:
## Changes

- the login shell can emit startup output that corrupts command results. This preserves stdout/stderr, suppresses startup noise, then restores both streams before running the command.


federicobozzini:
Closing in favour of #362

PR #359, opened 2026-07-21
refactor: add procedural layer for port exposure check

muchzill4:
## Changes

- Extract registry tunnel exposure checking from the operation layer into the Docker procedural layer.
- Move exposure-check coverage alongside the new procedural function.
- Add context propagation to SSH hostname resolution and remote port probing.
- Keep the existing operation as a delegating compatibility layer.

## Checklist

### DIFF
--- a/internal/deploy/deploy.go
+++ b/internal/deploy/deploy.go
@@ -47,7 +47,7 @@ func NewDeployment(composeFile string, opts DeployOptions) (goperation.Sequence,
 			cleanup = stopTunnel
 			ops = append(ops, operation.NewRunRegistry(opts.Registry.Port)...)
 			ops = append(ops, startTunnel)
-			if !opts.Registry.SkipRemotePortCheck {
+			if !opts.TargetHost.IsLocalhost() && !opts.Registry.SkipRemotePortCheck {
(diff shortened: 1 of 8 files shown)


Luke-Parkin (approved):
LGTM, Just to confirm that I understand the intent; You've moved the tunnel exposure check into docker/tunnel_exposure_check so that in the future, operation/registry_tunnel_exposure can select between podman.CheckTunnelExposure or docker.CheckTunnelExposure?

(I think these might be the same, but I guess we can find out later) - (and passing through ctx to the ssh commands so they are cancellable)

muchzill4:
Nope. I’m trying to de-operationify the flow, because it’s annoyingly hard to pass data between operations right now, unless they share a reference to some mutable state (yuck). I also want to pass `context.Context` in, so that the deployment flow can be responsible for doing the cleanup. Right now, the "cleanup logic” escapes its domain and leaks into the cli layer, which isn’t great.

For podman, the initial approach will be to copy paste a lot of these procedural flows and after it works e2e think about what we can share between flows. It’s likely that podman will require additional ssh proxy setup so that socket file on the target is shared with the host.

PR #362, opened 2026-07-21
fix: make sure SSH commands don't fail because of login shell startup output

federicobozzini:
## Changes

- Ignore all output from the login shell when executing remote commands

## Checklist


internal/ssh/exec.go
```diff
@@ -6,13 +6,16 @@ import (
...
-// Pass stdin data as optional parameter, or nil for no stdin.
-func RunCommand(ctx context.Context, dest Destination, command string, stdin []byte, sshArgs ...string) (string, string, error) {
-	args := slices.Concat(sshArgs, []string{"--", dest.String(), command})
+const commandStartMarker = "__TOPO_COMMAND_START__"
+
+// RunCommand executes a command in the destination's login shell and returns
+// stdout and stderr separately, excluding output emitted before the command starts.
+// Stderr is classified into a typed error when a known failure pattern is detected.
+func RunCommand(ctx context.Context, dest Destination, cmdStr string, stdin []byte, sshArgs ...string) (string, string, error) {
+	args := slices.Concat(sshArgs, []string{"--", dest.String(), wrapInLoginShell(cmdStr)})
```
muchzill4: WDYT about doing something like:

```go
wrapper := command.NewLoginShellWrapper()

cmdStr = wrapper.Wrap(cmdStr)

…

stdout, stderr := wrapper.Unwrap(stdoutBuf.String(), stderrBuf.string())
```

I feel like the wrapper could live in `command` package, as it’s universally useful for login shells. You could also unit test it very simply and skip the complexity of testing via the SSH runner:

```go
func TestLoginShellWrapper(t *testing.T) {
	t.Run("Wrap", func(t *testing.T) {
		// The value of this test is questionable, but perhaps worth having
		t.Run("wraps given command in login shell with marker", func(t *testing.T) {
			wrapper := command.NewLoginShellWrapperWithMarker("_FOO_")
			wrappedCmd := wrapper.Wrap("echo $PATH")

			wantCmd := "..."
			assert.Equal(t, wantCmd, wrappedCmd)
	})

	t.Run("Unwrap", func(t *testing.T) {
		t.Run("strips the marker from output", func(t *testing.T) {
			wrapper := command.NewLoginShellWrapperWithMarker("_FOO_")
			output := "_FOO_\nhello, some output"

			got := wrapper.Unwrap(output)

			want := "hello, some output"
			assert.Equal(t, want, got)
		})
	})
} 
```

muchzill4 (commented):
Nice fix.

internal/command/login_shell_test.go
```diff
@@ -0,0 +1,44 @@
+package command
```
muchzill4: We tend to use black box testing. This means using `package command_test` here and exporting the `NewLoginShellWrapperWithMarker` from `command` package. This is nice, in that it forces us programatically to never test the internals of the packages. 

I think exporting `NewLoginShellWrapperWithMarker` is ok. One might think of that as an “expert mode” for otherwise convenient `NewLoginShellWrapper`.

muchzill4 (approved):
Thanks for tweaking this, tests look great and it’s much easier to add coverage for the edge cases you’ve identified. 👍

PR #366, opened 2026-07-22
refactor: replace operation-based deployment with procedural approach

muchzill4:
## Changes

- Replace deployment operation sequences with explicit procedural orchestration.
- Propagate cancellation context from CLI signals through deployment steps.
- Ensure registry SSH tunnels are closed with a bounded cleanup timeout.
- Remove obsolete deployment operation wrappers and signal cleanup handling.

## Checklist


muchzill4:
I want to confine this deployment flow in `docker` namespace, but ballooning is an issue, so will be done separately. There’s also `Stop`, which is still an operation.

internal/operation/operation.go
```diff
@@ -1,45 +1,8 @@
...
-)
+import "io"
 
 type Operation interface {
 	Description() string
 	Run(cmdOutput io.Writer) error
 }
-
-// SetupExitCleanup sets up a handler to run an operation once when the program exits due to an interrupt signal.
-func SetupExitCleanup(w io.Writer, operation Operation, exit func(int)) func() {
```
awphi, line 20: :letsgooooo:

internal/deploy/deploy.go
```diff
@@ -37,31 +46,92 @@ func NewDeploymentStop(composeFile string, dest ssh.Destination) goperation.Sequ
...
+func Deploy(ctx context.Context, output io.Writer, composeFile string, opts DeployOptions) (deploymentError error) {
 	sourceHost := command.LocalHost
-	ops := []goperation.Operation{
-		operation.NewDockerComposeBuild(composeFile, sourceHost),
-		operation.NewDockerComposePull(composeFile, sourceHost),
+	if err := runStep(output, "Build images", func() error {
+		return docker.BuildImages(ctx, output, composeFile, sourceHost)
+	}); err != nil {
+		return err
+	}
```
awphi:
```suggestion
    buildImages := func() error {
		return docker.BuildImages(ctx, output, composeFile, sourceHost)
	}
	if err := runStep(output, "Build images", buildImages); err != nil {
		return err
	}
```
nit but imo easier to read - applies all the multi-line functions in `if` statements

internal/deploy/deploy.go
```diff
@@ -37,31 +46,92 @@ func NewDeploymentStop(composeFile string, dest ssh.Destination) goperation.Sequ
...
+
+	if err := runStep(output, "Start services", func() error {
+		return docker.StartServices(ctx, output, composeFile, targetHost, opts.RecreateMode)
+	}); err != nil {
+		return err
+	}
+	return runStep(output, "Deployment Success", func() error {
+		return post_deploy.PrintDeploySuccess(output, composeFile, post_deploy.DefaultMessage(composeFile))
+	})
```
awphi: Since registry closing is deferred now, we'll print that after the success message and if fails, the error log after that too.
```
      ┌─ Deployment Success
      Run `topo ps ...`

      ┌─ Close registry SSH tunnel
      it broke

      ERR deployment failed ... failed to close SSH tunnel
```
Imo this kinda distracts from the "Run `topo ps` bit we want users to focus on

internal/deploy/deploy.go
```diff
@@ -37,31 +46,92 @@ func NewDeploymentStop(composeFile string, dest ssh.Destination) goperation.Sequ
...
-			if !opts.TargetHost.IsLocalhost() && !opts.Registry.SkipRemotePortCheck {
-				ops = append(ops, operation.NewRegistryTunnelExposureCheck(opts.TargetHost, opts.Registry.Port))
+	if !opts.TargetHost.IsPlainLocalhost() && opts.Registry == nil {
+		if err := runStep(output, "Transfer images", func() error {
+			return docker.TransferImagesViaPipe(ctx, output, composeFile, sourceHost, targetHost)
+		}); err != nil {
+			return err
+		}
+	}
+	if !opts.TargetHost.IsPlainLocalhost() && opts.Registry != nil {
```
awphi: We've lost a little high-level readability with the departure of operations imo - it's more indentation but wdyt about
```go
if !opts.TargetHost.IsPlainLocalhost() {
  if opts.Registry == nil {
    // transfer via pipe 
  } else {
    // transfer via registry
  }
}
```
could of course extract into functions as needed

internal/deploy/deploy.go
awphi: Actually what about instead of `runStep` we pass in a new instance of the `ProgressReporter` interface that we use for spinners atm, but its impl. logs headings. Then the flow could look more like
```go
func Deploy(ctx context.Context, output io.Writer, composeFile string, opts DeployOptions) (deploymentError error) {
	sourceHost := command.LocalHost
    opts.progress.Step("Build Images")
    err := docker.BuildImages(ctx, output, composeFile, sourceHost)
    if err != nil {
        return nil
    }
    // etc...
}
```
I guess the issue is we need to handle errors from the header printing but I'd also think it's okay to just ignore them or panic. If we can't print to stdout then something has gone desperately wrong.

cmd/topo/deploy.go
```diff
@@ -87,22 +90,22 @@ By default, Topo uses compose.yaml in the current working directory, then compos
...
 		defer stop()
 
-		err = deployment.Run(os.Stdout)
-		if err != nil {
+		if err := deploy.Deploy(ctx, os.Stdout, composeFile, deployOpts); err != nil {
+			if errors.Is(err, context.Canceled) {
+				return err
+			}
 			return fmt.Errorf("deployment failed; ensure topo health is passing: %w", err)
 		}
```
awphi, line 108: I think there's another error we need to capture here for the ctrl+c case - it probably shouldn't tell me to check `topo health`

```
┌─ Close registry SSH tunnel ───────────────────────────
Exit request sent.
17:16:29 ERR deployment failed; ensure topo health is passing: failed to execute docker -H ssh://spartikus tag localhost:12737/topo-welcome-app@sha256:1f4e2ad574636b6a9ba6cd3bf82fa13fc660facc322deade110abed6552b72e7 topo-welcome-app: signal: killed
exit status 1
```

internal/deploy/deploy_test.go
awphi, line 1: Maybe 1-2 more tests to check some the conditional logic in `Deploy` itself? Happy if you disagree and feel it's well covered elsewhere

awphi (commented):
Think there's some more red lines to be had - all of `conditional.go` and `conditional_test.go` are now unused.

The strength of operations/sequences as an abstraction is much weaker without it handling deploy - wonder if we should de-operationify clone/setup-keys too 🤔 Not for right now ofc!

internal/deploy/deploy.go
muchzill4: The sad reality is that writing to writer can fail for many reasons and perhaps we shouldn’t panic. You might be writing to a log file, where running out of hd space should error. Of course here we could rely on “actual work” also failing to write to output, but I think I prefer the explicit handling of write errors. It’s ugly, but does the job.

I’ve tried improving things in 38767f2 and 3ea8347. WDYT?

internal/deploy/deploy.go
muchzill4: Done @ 38767f2. Attempted function extraction, but it ends up with a mess due to tunnel handling. One blob is ironically more readable.

internal/deploy/deploy_test.go
muchzill4: d179287

cmd/topo/deploy.go
muchzill4: dcc5e52

internal/deploy/deploy.go
muchzill4: e896f7d

cmd/topo/deploy.go
```diff
@@ -3,15 +3,16 @@ package main
...
 	"os"
+	"os/signal"
 	"runtime"
 	"strconv"
 	"strings"
+	"syscall"
 
 	"github.com/arm/topo/internal/deploy"
-	"github.com/arm/topo/internal/deploy/operation"
+	"github.com/arm/topo/internal/deploy/docker"
```
yejseo01, line 13: what would this bit look like after the podman integration? we probably won't call it "docker" right?

internal/deploy/deploy_test.go
```diff
@@ -2,196 +2,106 @@ package deploy_test
...
+		testutil.RequireImageExists(t, command.LocalHost, imageName)
+		testutil.AssertContainersRunning(t, ssh.PlainLocalhost, composeFilePath)
 	})
 
-	t.Run("excludes the remote port check for a localhost SSH target", func(t *testing.T) {
-		localhostSSHDest := ssh.NewDestination("ssh://root@localhost:2222")
-		opts := deploy.DeployOptions{
-			TargetHost: localhostSSHDest,
-			Registry:   &deploy.RegistryConfig{Port: deploy.DefaultRegistryPort},
-		}
```
yejseo01, line 137: Why you are removing some of the coverage on deploy behaviour? would this be covered somwhere else now?

yejseo01 (commented):
two questions

cmd/topo/deploy.go
muchzill4: Yup, this file is likely to disappear and core logic be moved to `docker` namespace.

internal/deploy/deploy_test.go
muchzill4: I think it’s equal coverage, perhaps even better: it actually runs through three flows we support. 👍

awphi (approved):

PR #367, opened 2026-07-22
refactor: simpler logger

awphi:
## Changes

- Replaces the use of slog and tint with a simple custom logger implementation:
  - Adds a new `colors` package that wraps ANSI color codes to replace what tint did. Hope is this is a flexible enough primitive that we can start to incorporate colors into our other output.
  - Ditches slog altogether since we weren't using the main feature of it (attributes) at all so it added complexity for no reason.
  - Pulls the global `logger` instance into a separate file `global.go`. `logger.go` now contains the underlying struct that global wraps. Note the global is not tested directly as it's a thin wrapper over a well-tested struct.
- Uses `synctest` in the tests to avoid timestamp annoyances. This was a nice TIL!

Note that although the plain output has changed in a couple of ways, the JSON output is unchanged meaning this is not a breaking change.

Before:
Screenshot 2026-07-22 at 16 02 32

After:
Screenshot 2026-07-22 at 16 02 06

## Checklist


yejseo01 (approved):
LGTM

PR #368, opened 2026-07-23
feat: ssh-based registry port probe

awphi:
## Changes

- Implements a new ssh-based port probe for our registry port listening check that works by parsing `/proc/net/tcp{6}`
  - Why do we need an ssh-based probe? `curl`, along with being an extra dependency for something quite minor, does not work with non-trivial SSH setups like `ProxyJump`/`ProxyCommand`.
  - Why parse procfs? It came with the least dependencies. Ofc it can still fail on particularly esoteric kernel builds or setups where the ssh user cannot access procfs for any reason but that's also the case with `topo describe` and rarer than curl-based failures.
    - Other alternative explored: Forwarding a TCP port from the target to the host and pushing an HTTP request down its stdin. This came with two downsides: required the ssh server to support tcp forwarding (particularly annoying on dropbear where it's a compile flag), required parsing ssh stderr to differentiate between positive/negative states which is unreliable as the output is server-specific.
  - How stable is`/proc/net/tcp`? Some basic reading suggests yes - all the columns we care about are stable and haven't changed in decades as many tools rely on them too.
- Rejigs the necessary tests to spin up target containers and start loopback-only or all interface listening processes to give some confidence the check actually functions in both the positive and negative cases
- Drops dependency on curl!

## Checklist

### DIFF
--- a/docs/introduction/install.mdx
+++ b/docs/introduction/install.mdx
@@ -26,7 +26,6 @@ A target must run Linux on AArch64 (`linux/arm64`). Topo uses Docker to build im
 Install the following tools on the host:
 
 - [Docker](https://docs.docker.com/get-docker/)
-- `curl`
 - OpenSSH client
 - Git to clone [Topo Projects](glossary.md#topo-project) from Git repositories
 
--- a/internal/deploy/docker/tunnel_exposure_check_test.go
+++ b/internal/deploy/docker/tunnel_exposure_check_test.go
@@ -61,24 +61,11 @@ func openSocket(t *testing.T, dest ssh.Destination, address string, port string)
 		_ = cmd.Process.Kill()
 	})
 
-	err = waitForOpenSocket(dest, address, port)
-	require.NoError(t, err)
-}
-
-func waitForOpenSocket(dest ssh.Destination, address string, port string) error {
-	deadline := time.Now().Add(5 * time.Second)
-	var lastErr error
-
-	for time.Now().Before(deadline) {
-		// #nosec G204 -- ignore as its a test helper
+	var output []byte
+	assert.Eventually(t, func() bool {
 		cmd := exec.Command("ssh", dest.String(), fmt.Sprintf("nc -z -w 1 %s %s", address, port))
-		output, err := cmd.CombinedOutput()
-		if err == nil {
-			return nil
-		}
-		lastErr = fmt.Errorf("%w output: %s", err, strings.TrimSpace(string(output)))
-		time.Sleep(200 * time.Millisecond)
-	}
-
-	return fmt.Errorf("port %s not ready: %w", port, lastErr)
+		var err error
+		output, err = cmd.CombinedOutput()
+		return err == nil
+	}, 5*time.Second, 200*time.Millisecond, "port %s not ready on %s: %v output: %s", port, address, err, string(output))
 }
--- a/internal/deploy/docker/tunnel_exposure_check_test.go
+++ b/internal/deploy/docker/tunnel_exposure_check_test.go
@@ -7,6 +7,7 @@ import (
 	"os/exec"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/arm/topo/internal/deploy/docker"
 	"github.com/arm/topo/internal/deploy/testutil"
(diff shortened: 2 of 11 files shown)


internal/testutil/container.go
```diff
@@ -46,6 +46,20 @@ var DinDContainer = ContainerSpec{
 	},
 }
 
+var SshServerContainer = ContainerSpec{
```
muchzill4: Should we call this `PasswordlessSSHContainer` for consistency’s sake?

internal/netprobe/port_test.go
```diff
@@ -1,55 +1,114 @@
...
+	tests := []struct {
+		name       string
+		procNetTCP string
+		port       string
+		expected   bool
+	}{
+		{
+			name:       "returns false for an IPv4 loopback listener",
+			procNetTCP: procNetTCPV4,
+			port:       "11000",
```
muchzill4: It’s hard to grok where 11000 comes from right now. Do you think we could do something along the lines of:

```go
func buildProcNetTable(entries []string) string { … }

func newProcNetEntry(port string, isListening bool) string {
   return fmt.Sprintf(“0: %s … %s …”, encodePort(port), encodeIsListening(isListening))
}
```

This would also potentially mean dropping table tests.

docs/introduction/install.mdx
```diff
@@ -26,7 +26,6 @@ A target must run Linux on AArch64 (`linux/arm64`). Topo uses Docker to build im
 Install the following tools on the host:
 
 - [Docker](https://docs.docker.com/get-docker/)
-- `curl`
```
muchzill4, line 29: 🎉

internal/deploy/docker/tunnel_exposure_check_test.go
```diff
@@ -17,44 +18,67 @@ import (
...
+func openSocket(t *testing.T, dest ssh.Destination, address string, port string) {
+	t.Helper()
+	cmd := exec.Command("ssh", dest.String(), fmt.Sprintf("nc -lk -p %s -s %s -e cat", port, address))
+	err := cmd.Start()
+	require.NoError(t, err)
+	t.Cleanup(func() {
+		_ = cmd.Process.Kill()
+	})
+
+	err = waitForOpenSocket(dest, address, port)
```
muchzill4: Perhaps you can leverage `t.Eventually` here?

internal/testutil/passwordless-ssh/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+    echo "DisableForwarding no" >> /etc/ssh/sshd_config && \
+    echo "PermitOpen any" >> /etc/ssh/sshd_config && \
+    echo "PermitListen any" >> /etc/ssh/sshd_config && \
+    echo "" >> /etc/ssh/sshd_config && \
+    echo "Match all" >> /etc/ssh/sshd_config && \
+    echo "    AllowTcpForwarding yes" >> /etc/ssh/sshd_config && \
+    echo "    DisableForwarding no" >> /etc/ssh/sshd_config
+
+RUN passwd -d root
+RUN ssh-keygen -A
```
muchzill4, line 18: Why is this required?

muchzill4 (approved):
LGTM!

internal/testutil/passwordless-ssh/Dockerfile
awphi: It's probably not, just a copy-pasta from the `dind` Dockerfile which does the same. I'll make a small follow up to try and trim some fat in these.

internal/testutil/container.go
awphi: https://github.com/arm/topo/pull/368/commits/c11d00edbdae9d1c2e1b0e575174d171df618fc6

internal/deploy/docker/tunnel_exposure_check_test.go
awphi: Nice https://github.com/arm/topo/pull/368/commits/246b3c5e1b9ec86b5e059aaa1418d4ed312155a4

internal/netprobe/port_test.go
awphi: I'd prefer to keep the real output blobs as real-world test cases to give more confidence in changes to the parser. Especially because encoding IPv4/v6 to procfs-style hex (addresses, not ports - ports are easy but only half the picture) is not difficult to implement but potentially even harder to understand.

I've rejigged the tests to avoid re-using the port `11000` and instead prefer decoding the hex in the test case so the relationship is clearer; https://github.com/arm/topo/pull/368/commits/155670de4fc82583f9388b3a77cb051abf64b762 - lmk what you think!

muchzill4 (approved):

PR #369, opened 2026-07-23
feat: experimental support for Podman localhost deployment

muchzill4:
## Changes

- Add an experimental `--engine podman` deployment path for localhost targets.
- Build, pull, and start Compose services using Podman with `docker-compose` as the Compose provider.
- Reject remote targets when Podman is selected.

## Try it locally

Prerequisites:

- Podman installed and running
- `docker-compose` installed

Create `compose-podman.yaml`:

```yaml
name: topo-podman-test

services:
  app:
    platform: linux/arm64
    image: docker.io/library/alpine:latest
    command: ["sh", "-c", "while true; do echo 'Topo Podman test is running'; sleep 30; done"]
```

Run the deployment:

```shell
TOPO_EXPERIMENTAL_FEATURES=1 go run ./cmd/topo deploy \
  --engine podman \
  --target localhost \
  --file compose-podman.yaml
```

Verify the service:

```bash
podman compose -f compose-podman.yaml ps
podman logs topo-podman-test-app-1
```

Clean up:

```bash
podman compose -f compose-podman.yaml down
```

## Checklist

### DIFF
--- a/cmd/topo/deploy.go
+++ b/cmd/topo/deploy.go
@@ -22,7 +22,7 @@ import (
 )
 
 var (
-	deploymentEngine    string
+	engine              string
 	noRegistry          bool
 	registryPort        string
 	skipRemotePortCheck bool
@@ -34,8 +34,8 @@ var (
 type containerEngine string
 
 const (
-	deploymentEngineDocker containerEngine = "docker"
-	deploymentEnginePodman containerEngine = "podman"
+	containerEngineDocker containerEngine = "docker"
+	containerEnginePodman containerEngine = "podman"
 )
 
 var deployOpts deploy.DeployOptions
@@ -56,11 +56,11 @@ By default, Topo uses compose.yaml in the current working directory, then compos
 	RunE: func(cmd *cobra.Command, args []string) error {
 		cmd.SilenceUsage = true
 
-		engine, err := parseDeploymentEngine(deploymentEngine)
+		parsedEngine, err := parseContainerEngine(engine)
 		if err != nil {
 			return err
 		}
-		if engine == deploymentEnginePodman {
+		if parsedEngine == containerEnginePodman {
 			return deployWithPodman(cmd)
 		}
 		return deployWithDocker(cmd)
@@ -158,15 +158,15 @@ func executeDeployment(cmd *cobra.Command, deployment func(context.Context) erro
 	return nil
 }
 
-func parseDeploymentEngine(value string) (containerEngine, error) {
-	engine := containerEngine(value)
-	if engine == "" {
-		return deploymentEngineDocker, nil
+func parseContainerEngine(value string) (containerEngine, error) {
+	parsedEngine := containerEngine(value)
+	if parsedEngine == "" {
+		return containerEngineDocker, nil
 	}
-	if engine != deploymentEngineDocker && engine != deploymentEnginePodman {
+	if parsedEngine != containerEngineDocker && parsedEngine != containerEnginePodman {
 		return "", fmt.Errorf("invalid engine %q: must be docker or podman", value)
 	}
-	return engine, nil
+	return parsedEngine, nil
 }
 
 func validatePort(port string) error {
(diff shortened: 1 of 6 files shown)


cmd/topo/deploy.go
```diff
@@ -46,68 +56,117 @@ By default, Topo uses compose.yaml in the current working directory, then compos
...
-		switch {
-		case forceRecreate:
-			deployOpts.RecreateMode = docker.RecreateModeForce
-		case noRecreate:
-			deployOpts.RecreateMode = docker.RecreateModeNone
-		}
+	targetArg, err := requireTarget(cmd)
+	if err != nil {
+		return err
+	}
```
awphi, line 100: Could hoist this and the create of `ssh.NewDestination(targetArg)` up into main command body since it's common between the two? 

Perhaps not a problem for this early version though so feel free to ignore.

cmd/topo/deploy.go
```diff
@@ -28,6 +31,13 @@ var (
 	noRecreate          bool
 )
 
+type containerEngine string
+
+const (
+	deploymentEngineDocker containerEngine = "docker"
+	deploymentEnginePodman containerEngine = "podman"
+)
```
awphi, line 39:
```suggestion
type containerEngine string

const (
	containerEngineDocker containerEngine = "docker"
	containerEnginePodman containerEngine = "podman"
)
```

nit as feels like we'll extract this to a `containerengine` package in the near future

cmd/topo/deploy.go
```diff
@@ -46,68 +56,117 @@ By default, Topo uses compose.yaml in the current working directory, then compos
...
+	if err := deployment(ctx); err != nil {
+		if ctx.Err() != nil {
+			return ctx.Err()
+		}
+		return fmt.Errorf("deployment failed; ensure topo health is passing: %w", err)
+	}
+	return nil
+}
+
+func parseDeploymentEngine(value string) (containerEngine, error) {
```
awphi:
```suggestion
func parseContainerEngine(value string) (containerEngine, error) {
```

similar nit here - dislike the term 'deployment engine' I suppose

internal/deploy/podman/deploy_test.go
```diff
@@ -0,0 +1,122 @@
...
+	"time"
+
+	"github.com/arm/topo/internal/deploy/podman"
+	gtestutil "github.com/arm/topo/internal/testutil"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestDeploy(t *testing.T) {
+	requireLocalPodman(t)
```
awphi, line 21: Do our CI runners already have podman? I could believe the linux ones do but would be very surprised if the windows one does.

Happy for this to be done later ofc but probably should be done before we go much further with podmanning the things

internal/deploy/podman/deploy_test.go
```diff
@@ -0,0 +1,122 @@
...
+		}
+
+		var object jsonObject
+		if err := json.Unmarshal(line, &object); err != nil {
+			return objects, err
+		}
+		objects = append(objects, object)
+	}
+	return objects, nil
+}
```
awphi: I know we said duplication is fine for v1 but this func already exists in `deploy/testutil`, can we just export it?

awphi (approved):
Nice work and works fine for me! Probably worth double-checking on a Windows machine with a fresh podman desktop install too.

Few small comments!

cmd/topo/deploy.go
muchzill4: 13f80d3

cmd/topo/deploy.go
muchzill4: 13f80d3

cmd/topo/deploy.go
muchzill4: Let’s keep it as is for now. Right now, each `deployWith*` branch simply takes `*cobra.Command` as parameter and is essentially an _extension_ of the root `deploy` command.

internal/deploy/podman/deploy_test.go
muchzill4: - [Linux doesn’t have `docker-compose`](https://github.com/arm/topo/actions/runs/30104881098/job/89519754910?pr=369#step:8:1035)
- [Windows doesn’t have `podman`](https://github.com/arm/topo/actions/runs/30104881098/job/89519754916?pr=369#step:8:1005)

CI does need more work, but I want to do this separately. 👍

internal/deploy/podman/deploy_test.go
muchzill4: 7a80205

PR #370, opened 2026-07-24
refactor: move docker ps to its own namespace

muchzill4:
## Changes

- Move Docker-specific container listing and tests into the `deploy/docker` package.
- Decouple output rendering from Docker in preparation for supporting other runtimes.

## Checklist

### DIFF
--- a/internal/deploy/docker/ps.go
+++ b/internal/deploy/docker/ps.go
@@ -1,4 +1,4 @@
-package deploy
+package docker
 
 import (
 	"bytes"
--- a/internal/deploy/docker/ps_test.go
+++ b/internal/deploy/docker/ps_test.go
@@ -3,7 +3,7 @@ package docker_test
 import (
 	"testing"
 
-	docker "github.com/arm/topo/internal/deploy/docker"
+	"github.com/arm/topo/internal/deploy/docker"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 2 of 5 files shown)


internal/deploy/docker/ps_test.go
```diff
@@ -1,9 +1,9 @@
-package deploy_test
+package docker_test
 
 import (
 	"testing"
 
-	"github.com/arm/topo/internal/deploy"
+	docker "github.com/arm/topo/internal/deploy/docker"
```
awphi:
```suggestion
	"github.com/arm/topo/internal/deploy/docker"
```

internal/deploy/docker/ps.go
awphi, line 36: Is there still a need for this? Couldn't the relevant functions just return the engine-agnostic `Container` struct? I suppose it risks being less re-usable if we ever read properties that exist in one engine and not the others but not sure that's an issue we have?

My main gripe is it's a little annoying to remember to map output twice - once from container engine CLI to `<container engine>.Container`, then again to `views.Container`

awphi (approved):
One question to ponder and one nit, not blocking on either count

internal/deploy/docker/ps.go
muchzill4: I don’t know yet, so I’ve decided to keep it safe and make the “orchestrator” - cli in the short term - do the mapping. I think we will either end up with a `deploy.go`, which will handle engine details and return common structs, or as you say, each engine will return a common shape. I’m going full undry mode here, see where the pieces fall. Worst case we’ll end up with a lot of red in subsequent PRs. 👍

awphi (approved):

PR #374, opened 2026-07-27
ci: test podman in CI

muchzill4:
## Changes

- Install `podman` and `docker-compose` in CI, so that podman tests are executed
- Turns out podman <5.3 doesn’t support `PODMAN_COMPOSE_WARNING_LOGS=false`, so I had to fix nd json parsing
    - This will affect `ps` in the future where we explicitly parse the output, something to think about there and then

## Checklist

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -136,7 +136,7 @@ jobs:
         if: runner.os == 'Windows'
         shell: pwsh
         run: |
-          podman machine init --provider wsl --cpus 2 --memory 2048 --now
+          podman machine init --cpus 2 --memory 2048 --now
 
       - name: Verify Podman test prerequisites (Linux)
         if: runner.os == 'Linux'
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -48,6 +48,8 @@ jobs:
         uses: Vampire/setup-wsl@d1da7f2c0322a5ee4f24975344f67fc0f5baf364 # v7
         with:
           distribution: Ubuntu-24.04
+          # Docker and Podman require the Linux kernel and container primitives provided by WSL2.
+          wsl-version: 2
           additional-packages: |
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -48,6 +48,8 @@ jobs:
         uses: Vampire/setup-wsl@d1da7f2c0322a5ee4f24975344f67fc0f5baf364 # v7
         with:
           distribution: Ubuntu-24.04
+          # Docker and Podman require the Linux kernel and container primitives provided by WSL2.
+          wsl-version: 2
```
muchzill4, line 52: This is default, but good to call out explicitly.

.github/workflows/ci.yaml
```diff
@@ -78,6 +80,82 @@ jobs:
...
+          COMPOSE_VERSION: v2.32.4
+        run: |
+          sudo apt-get update
+          sudo apt-get install --yes podman
+
+          case "$(uname -m)" in
+            x86_64) compose_arch=x86_64 ;;
+            aarch64|arm64) compose_arch=aarch64 ;;
+            *) echo "Unsupported architecture: $(uname -m)" >&2; exit 1 ;;
+          esac
```
awphi, line 96: nit: Is this needed? We don't have any runners on x86_32 or any other arch

.github/workflows/ci.yaml
awphi, line 1: Yeesh we need to break this up into something like
```
ci.yaml
setup_windows_containers.yaml
setup_linux_containers.yaml
```
Doesn't block this though, separate refactor for  a GH yaml lover to take on later!

awphi (approved):

.github/workflows/ci.yaml
muchzill4: We don’t, but it’d be good if it exploded otherwise. 🤷

PR #377, opened 2026-07-27
feat!: Use artifactory catalog

yejseo01:
## Changes

  - Loads the catalog and schema from Artifactory.
  - Removes the embedded catalog and builtin source.
  - Makes --source always available.
  - Adds shared HTTP fetching for catalog, install, upgrade, and version checks.
  - Fixes the --source registration panic.
  - Adds fetcher tests.

## Checklist


internal/fetch/fetch.go
```diff
@@ -0,0 +1,48 @@
...
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"time"
+)
+
+const timeout = 2 * time.Minute
```
muchzill4: I think this should be configurable by the caller. I.e.: passed w/ context already.

internal/fetch/fetch.go
```diff
@@ -0,0 +1,48 @@
...
+		return nil, fmt.Errorf("creating request: %w", err)
+	}
+
+	// #nosec G704 -- callers explicitly provide the URL and its scheme is validated above.
+	response, err := http.DefaultClient.Do(request)
+	if err != nil {
+		return nil, fmt.Errorf("sending request: %w", err)
+	}
+
+	if response.StatusCode < http.StatusOK || response.StatusCode >= http.StatusMultipleChoices {
```
muchzill4: Why not `!= http.StatusOK`?

internal/catalog/catalog.go
```diff
@@ -34,20 +25,23 @@ type Project struct {
 	Ref         string   `json:"ref"`
 }
 
-func ListBuiltinProjects() ([]Project, error) {
-	return parseProjects(catalogJSON)
-}
+const (
+	DefaultURL        = "https://artifacts.tools.arm.com/devx-topo-project-catalog/" + version + "/catalog/"
```
muchzill4: Why are some of these `const`s exported and some not? Are they supposed to be used outside of those packages?

internal/fetch/fetch.go
```diff
@@ -0,0 +1,48 @@
...
+		statusErr := fmt.Errorf("request failed: HTTP %d (%s)", response.StatusCode, response.Status)
+		return nil, errors.Join(statusErr, response.Body.Close())
+	}
+
+	data, readErr := io.ReadAll(response.Body)
+	if err := errors.Join(readErr, response.Body.Close()); err != nil {
+		return nil, fmt.Errorf("reading response: %w", err)
```
muchzill4: The error messages seem to be off across this PR:
- “reading response:”
- “sending request:”

Shouldn’t this say: “failed to read response”?

cmd/topo/projects.go
```diff
@@ -56,23 +51,18 @@ var projectsCmd = &cobra.Command{
...
+	flagValue, err := cmd.Flags().GetString(sourceFlag)
+	if err != nil {
+		panic(fmt.Sprintf("internal error: %s flag not registered: %v", sourceFlag, err))
 	}
-	return builtinProjects
+	if flagValue != "" {
+		return flagValue
+	}
+
+	return catalog.DefaultCatalogURL
```
muchzill4: Can we not put this as default value of the flag when registering it?

internal/catalog/catalog.go
```diff
@@ -82,6 +72,18 @@ func validateAgainstSchema(b []byte) error {
 	return schema.Validate(jsonDoc)
 }
 
+type schemaLoader struct {
```
muchzill4: That’s pretty sleek.

muchzill4 (dismissed):

muchzill4:
Is `fix` a good category for this? It’s not fixing a bug.

internal/catalog/catalog.go
yejseo01: no, only `DefaultCatalogURL` needs to be exported. Will de-capitalize other variables

internal/fetch/fetch.go
yejseo01: well, since we are trying to make it generic, I thought we might want to accept 204 No Content and 206 Partial Content, and the caller would handle such. Can stick to `!= http.StatusOK` if you think that's better?

internal/fetch/fetch.go
yejseo01: it was based on what we had already at `latest_artifactory.go`, but I agree. It is bad in that as well.

internal/fetch/fetch.go
yejseo01: done - [8a036a4](https://github.com/arm/topo/pull/377/commits/8a036a452d5bb2f8b07daa2c559b1c1f61cd93c7)

cmd/topo/projects.go
yejseo01: done - [c0535da](https://github.com/arm/topo/pull/377/commits/c0535da45984ac2d4d3488e8f0229383dbe8eb26)

internal/fetch/fetch.go
yejseo01: done - [8a036a4](https://github.com/arm/topo/pull/377/commits/8a036a452d5bb2f8b07daa2c559b1c1f61cd93c7)

internal/catalog/catalog.go
yejseo01: done - [c0535da](https://github.com/arm/topo/pull/377/commits/c0535da45984ac2d4d3488e8f0229383dbe8eb26)

internal/catalog/catalog.go
```diff
@@ -34,20 +25,23 @@ type Project struct {
 	Ref         string   `json:"ref"`
 }
 
-func ListBuiltinProjects() ([]Project, error) {
-	return parseProjects(catalogJSON)
-}
+const (
+	DefaultURL        = "https://artifacts.tools.arm.com/devx-topo-project-catalog/" + version + "/catalog/"
+	DefaultCatalogURL = DefaultURL + "catalog.json"
+	DefaultSchemaURL  = DefaultURL + "catalog.schema.json"
```
awphi: It feels a little unfortunate to download the schema for artifactory every time for every type of source. It means even users who use `file://` sources require a network connection + slows the command down to download a static file every time.

Maybe fine for a v1 but wonder if we should consider downloading + embedding the schema version that the CLI supports into the binary? Especially since the `catalogDocument` type is a mirror of the schema.

--- 

Bonus - wonder if we should generate go types based on the schema?!

muchzill4: Come to think of it, you’re right @awphi, this is a bit wasteful. I wonder if generating types based on the schema would be better. We could also do a pre-flight check of the `$schema` field in the received catalog - if it matches our expectation.

yejseo01: yea I was thinking this on the way back home. For Bartek's suggestion, I'm not sure how that would alleviate user with `file://` source from needing network connection? I think to address that, we have no choice but to keep the schema in the same repository

yejseo01: Another thing we can try is to fetch the schema on release of topo, and embed that in the binary on release

yejseo01:
for error messages to precisely compare the versions, this PR needs to be merged first to continue - https://github.com/arm/topo-project-catalog/pull/8

internal/fetch/fetch.go
muchzill4: How would caller handle them right now?

yejseo01: right now, they just check `!= http.StatusOK`. Maybe you're right, and we won't need it. Will fix it to `!= http.StatusOK`

yejseo01: done - [09ac673](https://github.com/arm/topo/pull/377/commits/09ac67388429923d9449c789f56941fd569275c9)

internal/catalog/catalog.go
```diff
@@ -1,85 +1,35 @@
...
-
-	return catalog.Projects, nil
-}
-
-func validateAgainstSchema(b []byte) error {
-	const projectsSchemaURL = "https://raw.githubusercontent.com/arm/topo/main/internal/catalog/data/catalog.schema.json"
-
-	compiler := jsonschema.NewCompiler()
-	schemaDoc, err := jsonschema.UnmarshalJSON(bytes.NewReader(catalogSchemaJSON))
+func parseProjects(ctx context.Context, b []byte) ([]Project, error) {
```
awphi: `ctx` is unused

internal/fetch/fetch.go
awphi, line 1: ❤️

docs/development/DEVELOPMENT.md
```diff
@@ -87,3 +87,19 @@ docker compose up
...
+
+## Updating the catalog schema
+
+Topo uses the catalog version declared in `internal/catalog/catalog_schema_generated.go`. The same version selects both the catalog and its schema from Artifactory.
+
+Generating the Go types requires Node.js 20 or newer, `npx`, and access to Artifactory. Pass the catalog version to the generator:
+
+```sh
+go run ./scripts/generate_catalog_types v1
+```
```
awphi, line 99: Was `//go generate` not worth the trouble? I kind of like the idea of running all of our codebase dev tasks through standard go-supported paths if possible, not a blocker ofc

docs/development/DEVELOPMENT.md
```diff
@@ -87,3 +87,19 @@ docker compose up
 ```
 
 The documentation preview is available at `http://localhost:3000` and automatically reloads when files change.
+
+## Updating the catalog schema
+
+Topo uses the catalog version declared in `internal/catalog/catalog_schema_generated.go`. The same version selects both the catalog and its schema from Artifactory.
+
+Generating the Go types requires Node.js 20 or newer, `npx`, and access to Artifactory. Pass the catalog version to the generator:
```
awphi: Is it worth calling out access to artifactory? It's public so anyone with an internet connection should have access right?

scripts/generate_catalog_types/main.go
```diff
@@ -0,0 +1,137 @@
...
+	"strings"
+	"time"
+)
+
+var catalogSchemaVersionPattern = regexp.MustCompile(`^v(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)$`)
+
+const (
+	quicktypeVersion     = "26.0.0"
+	defaultSchemaBaseURL = "https://artifacts.tools.arm.com/devx-topo-project-catalog/"
+	generationTimeout    = 2 * time.Minute
```
awphi: Is a timeout strictly necessary? Think we can get away without it for an internal dev tool like this that does a pretty simple task

scripts/generate_catalog_types/main.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	formatted, err := addCatalogVersionConstant(generated, catalogVersion)
+	if err != nil {
+		return fmt.Errorf("failed to format quicktype output from schema %q: %w", schemaURL, err)
+	}
+	// #nosec G703 -- outputFile is derived from the generator source location.
+	if err := os.WriteFile(outputFile, formatted, 0o644); err != nil {
+		return fmt.Errorf("failed to write generated types to %q: %w", outputFile, err)
+	}
```
awphi: Is building an AST worth it here? Couldn't we just inject the const at the end of the file? e.g.

```suggestion
	formatted := string(generated) + "\n" + "const CatalogMajorVersion = \"" + majorVersion(catalogVersion) + "\"\n"
	// #nosec G703 -- outputFile is derived from the generator source location.
	if err := os.WriteFile(outputFile, []byte(formatted), 0o644); err != nil {
		return fmt.Errorf("failed to write generated types to %q: %w", outputFile, err)
	}
```

scripts/generate_catalog_types/main.go
```diff
@@ -0,0 +1,137 @@
...
+	if err := format.Node(&formatted, fileSet, file); err != nil {
+		return nil, fmt.Errorf("failed to format generated types: %w", err)
+	}
+	return formatted.Bytes(), nil
+}
+
+func majorVersion(version string) string {
+	major, _, _ := strings.Cut(version, ".")
+	return major
+}
```
awphi: It might be useful to store the full version in the `const` for us as maintainers if nothing else so we know what version of the schema a given version of the CLI uses without manually diffing our go types vs the source json schema.

internal/catalog/catalog_schema_generated.go
```diff
@@ -0,0 +1,68 @@
...
+	// parameter values. Unknown hint keys should be ignored.
+	Hints *Hints `json:"hints,omitempty"`
+	// If true, implementations must enforce input or error.
+	Required *bool `json:"required,omitempty"`
+}
+
+// Advisory metadata that implementations may use to discover, filter, or suggest suitable
+// parameter values. Unknown hint keys should be ignored.
+type Hints struct {
+}
```
awphi, line 62: Why has this been generated empty? Does quicktype not support `patternProperties`?

awphi (commented):
Nice! Have approved https://github.com/arm/topo-project-catalog/pull/8 so we can add the major version check.

internal/catalog/catalog.go
yejseo01: good catch

docs/development/DEVELOPMENT.md
yejseo01: yes. I'll get rid of this mention

docs/development/DEVELOPMENT.md
yejseo01: if you like it, no problem using it. I was thinking if someone is not familiar with go generate, it's one less thing for them to read about if we keep it this way though. Basically does the same thing, but with directives, it's more looking up in the internet than just reading the code in my perspective

scripts/generate_catalog_types/main.go
yejseo01: good catch. I agree

scripts/generate_catalog_types/main.go
yejseo01: I guess we can? it's just quite awkward to have const at the bottom in my instinct

yejseo01: yea, I think we should judging by how much code we save

scripts/generate_catalog_types/main.go
yejseo01: I agree. making this change now. Also we don't have the messge comparing the schema and the catalog version yet. making this as well

internal/catalog/catalog.go
yejseo01: done - [4088273](https://github.com/arm/topo/pull/377/commits/4088273072c6e8b5b715b0b22ec120614c79bb30)

docs/development/DEVELOPMENT.md
yejseo01: done - [1b69da0](https://github.com/arm/topo/pull/377/commits/1b69da0602732251c63003721ab17bfe31a06313)

scripts/generate_catalog_types/main.go
yejseo01: done - [e6921e1](https://github.com/arm/topo/pull/377/commits/e6921e18fdb84e798e8d543dba0cf560e74c280f)

scripts/generate_catalog_types/main.go
yejseo01: done - [ce1b9fa](https://github.com/arm/topo/pull/377/commits/ce1b9fa8a84651c0edef5203729bc58afd427eea)

scripts/generate_catalog_types/main.go
yejseo01: done - [b832259](https://github.com/arm/topo/pull/377/commits/b832259eb010824eebaf66b3ac5aedb99c395849)

docs/development/DEVELOPMENT.md
awphi: On the flipside, it's nicer since you learn what `go generate` is once, then you can use it across any Go project rather than learning project-specific idioms like `go run ./scripts/...`, Makefiles, Justfiles or whatever people choose to use. My understanding is that's the whole point of go generate - standardisation.

internal/catalog/catalog_schema_generated.go
yejseo01: oh good catch. It seems like `patternProperties` are not supported in quicktype. Does Hints need to be `patternProperties`?

docs/development/DEVELOPMENT.md
yejseo01: hmmm I guess that is true. I'll try making it a go generate then

yejseo01: done - [0a007b7](https://github.com/arm/topo/pull/377/commits/0a007b778d419ffde5cecdb8a2407cda616fb088)

internal/catalog/catalog_schema_generated.go
awphi: Good question. I don't see why it does personally! Regardless, I don't think it blocks this PR since we don't consume `hints` in the CLI at all atm.

docs/development/DEVELOPMENT.md
awphi: Nice commit SHA

internal/catalog/catalog.go
```diff
@@ -25,13 +31,32 @@ func ListProjectsFromURL(ctx context.Context, url string) ([]Project, error) {
...
+			return nil, fmt.Errorf(
+				"failed to unmarshal catalog: catalog version %q differs from generated schema version %q: %w",
+				catalogVersion,
+				CatalogSchemaVersion,
+				err,
+			)
+		}
 		return nil, fmt.Errorf("failed to unmarshal catalog: %w", err)
 	}
 	return catalog.Projects, nil
```
awphi, line 51: Is it intentional to only catch version incompatibility if unmarshalling fails? I guess technically a breaking schema change should always break unmarshalling but maybe it's better to be aggressive with the version checking at least to start with?

internal/catalog/catalog.go
```diff
@@ -11,27 +14,51 @@ import (
...
 }
 
-func parseProjects(ctx context.Context, b []byte) ([]Project, error) {
+func parseProjects(b []byte) ([]Project, error) {
+	catalogVersion, versionErr := unmarshalCatalogVersion(b)
 	catalog, err := UnmarshalCatalogDocument(b)
 	if err != nil {
+		if versionErr == nil && catalogVersion != "" && catalogVersion != CatalogSchemaVersion {
+			return nil, fmt.Errorf(
+				"failed to unmarshal catalog: catalog version %q differs from generated schema version %q: %w",
```
awphi: nit: Maybe something more like `"failed to parse catalog: requested catalog version %q is incompatible with supported schema version %q: %w"`? 'generated' and 'unmarshal' are implementation details and 'differs' is too broad since it doesn't have to be an exact match

awphi (commented):
Nice! One implementation question and one error message nit!

internal/catalog/catalog.go
yejseo01: this bit of the code is intentional because we want version 2.5.0 catalog to be parsable by version 2.3.0 schema.  The best wishes we have is for one version of schema to be able to parse many different versions of catalog even when catalog schema exceeds the schema version.

We can put the major version comparison, but I am slightly against it in case we want to make a major change in catalog schema that is backwards compatible

internal/catalog/catalog.go
yejseo01: done - [bf53d6a](https://github.com/arm/topo/pull/377/commits/bf53d6afebf654df91de4bc4bb0663e6f26233a6)

internal/catalog/catalog.go
yejseo01: done - [f30b5d4](https://github.com/arm/topo/pull/377/commits/f30b5d4840bffdb3f390adc13dc30ed1c642024d)

internal/catalog/catalog.go
```diff
@@ -47,39 +33,32 @@ func ListProjectsFromURL(ctx context.Context, url string) ([]Project, error) {
 }
 
 func parseProjects(b []byte) ([]Project, error) {
-	if err := validateAgainstSchema(b); err != nil {
-		return nil, fmt.Errorf("failed schema validation: %w", err)
+	catalogVersion, versionErr := unmarshalCatalogVersion(b)
+	catalogVersionMajor := majorVersion(catalogVersion)
+	SchemaVersionMajor := majorVersion(CatalogSchemaVersion)
```
awphi:
```suggestion
	schemaVersionMajor := majorVersion(CatalogSchemaVersion)
```

Alternatively, could add a `var majorCatalogVersion = majorVersion(CatalogSchemaVersion)` at the top and re-use in the `defaultURL` to avoid repeating. This isn't a blocker though

internal/catalog/catalog_test.go
```diff
@@ -76,15 +65,37 @@ func TestListProjectsFromURL(t *testing.T) {
 		_, err := catalog.ListProjectsFromURL(context.Background(), url)
 
 		require.Error(t, err)
-		assert.ErrorContains(t, err, "failed to unmarshal projects")
+		assert.ErrorContains(t, err, "failed to parse catalog")
+		assert.ErrorContains(t, err, `requested catalog version "" is incompatible`)
+	})
+
+	t.Run("reports catalog and schema versions when incompatible catalog fails to unmarshal", func(t *testing.T) {
```
awphi:
```suggestion
	t.Run("reports incompatible catalog", func(t *testing.T) {
```

Test name isn't overly verbose and wrong atm since it doesn't only report when unmarshaling fails

awphi (approved):
Couple more nits but nothing to block over (though please fix the variable casing!)

internal/catalog/catalog_test.go
yejseo01: done - [f0a3cc0](https://github.com/arm/topo/pull/377/commits/f0a3cc0b0a445d61e87928731753b415b005d4ba)

internal/catalog/catalog.go
yejseo01: done - [f0a3cc0](https://github.com/arm/topo/pull/377/commits/f0a3cc0b0a445d61e87928731753b415b005d4ba)

PR #378, opened 2026-07-28
docs: add versioned docs

awphi:
## Changes

- Bumps us to the latest version of ADP; 0.6.0 which comes with
  - New styling
  - Support for versioned docs
- Adds automated versioned docs releasing:
  - The main pain point of the docs versioning is it requires hard-coding a list of refs we want to host docs for in the repo. This felt like a lot of manual work so this adds a little script to generate the relevant config for the last 5 major releases.
  - Runs said script prior to publishing to staging. As part of this, the script also supports passing `--include-head` so we can preview a 'Latest' version in staging
  - ~~Adds a `compose.preview.yaml` compose project to preview the production/staging version of the docs locally. I don't foresee this being desperately useful but it does provide some level of confidence before deploying.~~

I'm keen for feedback if we think the automated `versions` generation is the right way to go + if we need the local production preview project at all. Lmk!

## Checklist


docs/scripts/generate-docs-config.sh
```diff
@@ -0,0 +1,25 @@
...
+cd "$(dirname "$0")/.."
+
+include_head=false
+if [[ ${1:-} == --include-head ]]; then
+  include_head=true
+fi
+
+release_refs=$(
+  for ref in $(gh release list --exclude-drafts --exclude-pre-releases --limit 1000 --json tagName --jq '[.[] | select(.tagName | test("^v?[0-9]+\\.0\\.0$"))][:5][].tagName'); do
+    if git cat-file -e "$ref:docs/docs-config.json" 2>/dev/null; then
```
muchzill4:
```sh
has_docs_config_at_ref() {
  git cat-file -e "$1:docs/docs-config.json" 2>/dev/null
}
```

And maybe:

```sh
last_five_major_release_tags=$(
  gh release list \
    --exclude-drafts --exclude-pre-releases \
    --limit 1000 --json tagName |
    jq '[.[] | select(.tagName | test("^v?[0-9]+\\.0\\.0$"))][:5][].tagName'
)
```

muchzill4 (commented):
We need to push latest major, not `major.0.0`. I.e.: `7.5.2` > `7.0.0`.

docs/scripts/generate-docs-config.sh
awphi: As discussed, have hard-coded a minimum major version to search from + made it fetch all latest releases from each major prefix

muchzill4 (approved):

PR #379, opened 2026-07-28
feat: experimental support for deployment to remote podman

muchzill4:
## Changes

- Deploy to remote Podman targets through an SSH tunnel to the Podman socket.
    - Use a loopback TCP tunnel because it bridges the target’s Unix socket to a local endpoint compatible with Podman and Compose across Unix and Windows hosts.
- Transfer images to remote targets using `podman save`/`load` pipes (registry coming next).

## Checklist

### DIFF
--- a/internal/deploy/podman/deploy_test.go
+++ b/internal/deploy/podman/deploy_test.go
@@ -6,25 +6,51 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"os"
 	"os/exec"
 	"path/filepath"
 	"testing"
 	"time"
 
 	"github.com/arm/topo/internal/deploy/podman"
+	"github.com/arm/topo/internal/ssh"
+	gtestutil "github.com/arm/topo/internal/testutil"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v3"
 )
 
 func TestDeploy(t *testing.T) {
 	requireLocalPodman(t)
-	composeFile, projectName := deploymentFixture(t)
-	t.Cleanup(func() { cleanupComposeProject(t, composeFile) })
 
-	err := podman.Deploy(t.Context(), io.Discard, composeFile)
+	t.Run("deploys to localhost", func(t *testing.T) {
+		composeFile, projectName := deploymentFixture(t)
+		t.Cleanup(func() { cleanupComposeProject(t, composeFile) })
+		options := podman.DeployOptions{TargetHost: ssh.PlainLocalhost}
 
-	require.NoError(t, err)
-	assertContainersRunning(t, projectName, podman.LocalSocket)
+		err := podman.Deploy(t.Context(), t.Output(), composeFile, options)
+
+		require.NoError(t, err)
+		assertContainersRunning(t, projectName, podman.LocalSocket)
+	})
+
+	t.Run("transfers images to a remote host via pipe", func(t *testing.T) {
+		target := gtestutil.StartContainer(t, gtestutil.PodmanContainer)
+		composeFile, projectName := deploymentFixture(t)
+		fixPodmanInDockerQuirk(t, composeFile)
+		targetHost := ssh.NewDestination(target.SSHDestination)
+		options := podman.DeployOptions{TargetHost: targetHost}
+
+		err := podman.Deploy(t.Context(), t.Output(), composeFile, options)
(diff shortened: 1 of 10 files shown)


internal/ssh/tcp_to_unix_socket_tunnel.go
```diff
@@ -0,0 +1,100 @@
...
+func OpenTCPToUnixSocketTunnel(ctx context.Context, output io.Writer, dest Destination, remoteSocketPath string) (*TCPToUnixSocketTunnel, error) {
+	address, err := availableLoopbackAddress()
+	if err != nil {
+		return nil, err
+	}
+
+	args := []string{
+		"-N",
+		"-o", "ExitOnForwardFailure=yes",
+		"-L", fmt.Sprintf("%s:%s", address, remoteSocketPath),
```
awphi, line 35: This puts a requirement on the target SSH server to support TCP forwarding correct? 

This will probably need documenting or a nice error message if/when it fails. Also worth noting while it's trivial to enable TCP forwarding on OpenSSH server, dropbear support for it is set at compile time which may cause a headache

internal/ssh/tcp_to_unix_socket_tunnel.go
```diff
@@ -0,0 +1,100 @@
...
+
+func availableLoopbackAddress() (string, error) {
+	listener, err := net.Listen("tcp4", "127.0.0.1:0")
+	if err != nil {
+		return "", fmt.Errorf("failed to reserve tunnel port: %w", err)
+	}
+	address := listener.Addr().String()
+	if err := listener.Close(); err != nil {
+		return "", fmt.Errorf("failed to release tunnel port %s: %w", address, err)
+	}
```
awphi, line 76: nit: is it a tunnel port at this point? would be a bit more generic

internal/deploy/podman/image_transfer_pipe.go
```diff
@@ -0,0 +1,53 @@
...
+func TransferImagesViaPipe(ctx context.Context, output io.Writer, source, destination Socket, composeFile string) error {
+	images, err := compose.ImageNames(composeFile)
+	if err != nil {
+		return err
+	}
+	for _, image := range images {
+		if err := transferImageViaPipe(ctx, output, source, destination, image); err != nil {
+			return err
+		}
+	}
```
awphi, line 23: Should these be transferred sequentially like this or in parallel via an outer `Group` like in the docker pipe transfer?

internal/deploy/podman/deploy_test.go
```diff
@@ -92,6 +118,34 @@ CMD ["tail", "-f", "/dev/null"]
...
 
+// fixPodmanInDockerQuirk avoids a Docker Desktop nested-container restriction.
+// The Podman target inherits oom_score_adj: 200, but Podman otherwise starts
+// each service with oom_score_adj: 0. Docker Desktop rejects that decrease, so
+// this adds oom_score_adj: 200 to each fixture service, for example:
+//
+//	services:
+//	  app:
+//	    oom_score_adj: 200
+func fixPodmanInDockerQuirk(t *testing.T, composeFile string) {
```
awphi: Should we just add the `oom_score_adj` to the compose content written by `deploymentFixture`?

I guess I'm wondering what happens if this is omitted - does fail silently and is therefore easy to forget?

muchzill4: 703fddd

internal/deploy/podman/image_transfer_pipe.go
muchzill4: Nice catch! 8c63996

internal/ssh/tcp_to_unix_socket_tunnel.go
muchzill4: 8b08097

internal/ssh/tcp_to_unix_socket_tunnel.go
muchzill4: Mhm. Are you ok for me to tackle this separately? I’ll put it in my backlog. Would probably like to do registry transfer first and then worry about probing/health.

FWIW, I tried it on all our targets and it seems to work (we use “standard” yocto images).

awphi: Sounds good to me!

awphi (approved):

PR #382, opened 2026-07-29
feat!: remove `topo extend`

awphi:
## Changes

- Deletes `topo extend` as we don't want to encourage use of compose file composition in Topo projects. Note that this doesn't break support for compose `extends:` usage, just removed the `topo extend` command
- Updates docs to remove references to `topo extend`, or `topo extend` specific information. There is still lots of information about how parameters interact with compose `extends:` usage as that's still relevant.

## Checklist


federicobozzini (approved):
❤️

PR #386, opened 2026-07-31
fix: Podman compose incorrectly targets docker engine w/ compose

muchzill4:
## Changes

ℹ️  This seems to affect podman 4.9.x, which is the version ubuntu 24.04 LTS ship with, my local tests with brewed podman 6.0.0 did work as expected w/o explicit `DOCKER_HOST` being set

- Configure the `docker-compose` provider with Podman’s resolved local API endpoint via `DOCKER_HOST`.
- Keep raw Podman commands on their normal local configuration; use `CONTAINER_HOST` only for explicit sockets.
- Correct deployment assertions to query containers with raw Podman rather than Compose, preventing tests from checking Docker while Podman Compose was misconfigured.
   - See the CI failure from the first commit of this PR 
- Cache the resolved local Compose endpoint

## Checklist

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -142,11 +142,14 @@ jobs:
         if: runner.os == 'Linux'
         shell: bash
         run: |
-          runtime_dir="$RUNNER_TEMP/podman-runtime"
+          runtime_dir="${XDG_RUNTIME_DIR:-$RUNNER_TEMP/podman-runtime}"
           mkdir -p "$runtime_dir/podman"
-          chmod 700 "$runtime_dir"
-          echo "XDG_RUNTIME_DIR=$runtime_dir" >> "$GITHUB_ENV"
-          export XDG_RUNTIME_DIR="$runtime_dir"
+
+          if [ -z "${XDG_RUNTIME_DIR:-}" ]; then
+            chmod 700 "$runtime_dir"
+            echo "XDG_RUNTIME_DIR=$runtime_dir" >> "$GITHUB_ENV"
+            export XDG_RUNTIME_DIR="$runtime_dir"
+          fi
 
           socket_path="$XDG_RUNTIME_DIR/podman/podman.sock"
           if [ ! -S "$socket_path" ]; then
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -138,6 +138,29 @@ jobs:
         run: |
           podman machine init --cpus 2 --memory 2048 --now
 
+      - name: Start Podman API service (Linux)
+        if: runner.os == 'Linux'
+        shell: bash
+        run: |
+          runtime_dir="$RUNNER_TEMP/podman-runtime"
+          mkdir -p "$runtime_dir/podman"
+          chmod 700 "$runtime_dir"
+          echo "XDG_RUNTIME_DIR=$runtime_dir" >> "$GITHUB_ENV"
+          export XDG_RUNTIME_DIR="$runtime_dir"
+
+          socket_path="$XDG_RUNTIME_DIR/podman/podman.sock"
+          if [ ! -S "$socket_path" ]; then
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yaml
```diff
@@ -138,6 +138,29 @@ jobs:
...
 
+      - name: Start Podman API service (Linux)
+        if: runner.os == 'Linux'
+        shell: bash
+        run: |
+          runtime_dir="$RUNNER_TEMP/podman-runtime"
+          mkdir -p "$runtime_dir/podman"
+          chmod 700 "$runtime_dir"
+          echo "XDG_RUNTIME_DIR=$runtime_dir" >> "$GITHUB_ENV"
+          export XDG_RUNTIME_DIR="$runtime_dir"
```
th3james: Overwriting this seems a touch invasive and could let to subtle misbehaviour. Possible that I'm being paranoid, but is it worth respecting the existing value?
```suggestion
          runtime_dir="${XDG_RUNTIME_DIR:-$RUNNER_TEMP/podman-runtime}"
          mkdir -p "$runtime_dir/podman"

          if [ -z "${XDG_RUNTIME_DIR:-}" ]; then
            chmod 700 "$runtime_dir"
            echo "XDG_RUNTIME_DIR=$runtime_dir" >> "$GITHUB_ENV"
            export XDG_RUNTIME_DIR="$runtime_dir"
          fi
```

internal/deploy/podman/socket.go
```diff
@@ -0,0 +1,232 @@
...
+
+	if socket.url == "" {
+		url, err := ResolveLocalComposeSocket(context.Background())
+		if err != nil {
+			return nil, err
+		}
+		socket.url = url
+	}
+	return append(environment, "DOCKER_HOST="+socket.url), nil
+}
```
th3james: Found this quite hard to follow, I think it's clearer if we separate out "get the URL from a `Socket`" from "change the ENV"


```suggestion
func (socket Socket) ConfigureComposeEnv(environment []string) ([]string, error) {
	socketURL, err := socket.GetSocketURL()
	if err != nil {
		return nil, err
	}
	return append(environment, "DOCKER_HOST="+socketURL), nil
}

func (socket Socket) GetSocketURL() (string, error) {
	if socket.localComposeSocket != nil {
		return socket.localComposeSocket.GetSocketURL()
	}
	if socket.url == "" {
		return "", errors.New("podman socket URL not configured")
	}
	return socket.url, nil
}

func (socket *localComposeSocket) GetSocketURL() (string, error) {
	socket.mutex.Lock()
	defer socket.mutex.Unlock()

	if socket.url == "" {
		url, err := ResolveLocalComposeSocket(context.Background())
		if err != nil {
			return "", err
		}
		socket.url = url
	}
	return socket.url, nil
}
```

internal/deploy/podman/socket.go
```diff
@@ -0,0 +1,232 @@
...
+	Path string
+}
+
+type machinePipe struct {
+	Path string
+}
+
+// ResolveLocalComposeSocket returns the host-accessible Podman API endpoint
+// that the external Compose provider must use for a local deployment.
+func ResolveLocalComposeSocket(ctx context.Context) (string, error) {
```
th3james, line 90: This is on the public interface and it takes a `context`, but it has a single call site inside this package , and the only context it gets passed is `context.Background()`.
Should we consider either removing the ability to pass a context, or update the function APIs up the call stack to allow passing the existing context down?

th3james (approved):
Functionally looks OK, but there are a few tweaks to consider

internal/deploy/podman/socket.go
muchzill4: eff1e80

internal/deploy/podman/socket.go
muchzill4: 248673a

PR #389, opened 2026-07-31
docs: embed light/dark styling in svg diagram

awphi:
## Changes

- The troublesome diagram SVG strikes again and has bricked staging deployment! The problem is that APDS wants the index page to be a standalone React app and does not support relative imports. Here's what's been tried so far:
  - Relative `@site` imports - works for publishing, does not work for previewing locally
  - Symlinking the svg into the index page dir - works for previewing locally, does not work for publishing
  - **This; embedding the light/dark styling in the svg itself (the whole reason we want to inline the svg at all) and just using an `` tag in the public directory - seems to work for both.** 
- Also softens the foreground `color` in light mode to match the new text styling

Successful staging deploy from this branch: https://github.com/arm/topo/actions/runs/30658690136

## Checklist

### DIFF
--- a/docs/landing-page/index.module.css
+++ b/docs/landing-page/index.module.css
@@ -163,15 +163,10 @@
 }
 
 .heroDiagram {
-  color: #000;
   display: block;
   height: auto;
   width: 100%;
   max-width: 100%;
-}
-
-:global(html[data-theme="dark"]) .heroDiagram {
-  color: #fff;
 }
 
 @media screen and (min-width: 1440px) {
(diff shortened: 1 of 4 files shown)


yejseo01 (approved):
LGTM. Would it be okay to attach the difference in screenshots from now on?

awphi:
> LGTM. Would it be okay to attach the difference in screenshots from now on?

Sure, have just been avoiding it since the docs are still unreleased! Probably doesn't matter in all honesty though.

PR #398, opened 2026-08-07
docs: Discover, Configure, Deploy in intro

th3james:
## Changes

- Align intro with our three-verb messaging

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,8 +2,7 @@
 
 Discover, configure, and deploy containerised software to Arm hardware over SSH.
 
-### Discover
-Point Topo at any Arm-based Linux device to discover [Topo Projects](docs/project-specification/README.md) that showcase its capabilities.
+*Discover* - Point Topo at any Arm-based Linux device to discover [Topo Projects](docs/project-specification/README.md) that showcase its capabilities.
 
 ```console
 $ topo projects --target user@my-linux-device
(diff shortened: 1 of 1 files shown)


federicobozzini (approved):
This change is fine, but I wonder if we should lean on it more and really make it our slogan.

What it would mean in practice?

1. capitalise the three verbs
2. Have a small paragraph to explain them a bit better

```md
## Discover
We provide examples ...

## Configure
The example can be configured for your specific hardware capabilities...

### Deploy
Topo makes deployment easy in such and such way..
```

federicobozzini (commented):
I like it!

Two comments:

1. in other places we suggest `topo health` is the first operation a user should run. Here this is lost
2. `topo clone` was going to be deprecated I think? Maybe it's better to do `git clone` and `topo configure`. Also clearer from the perspective of the terminology used.

th3james:
@federicobozzini
1. The focus here is on trying to highlight the commands which communicate the value proposition. `topo health` is something that helps you use topo, but brings no value in of itself. To mitigate the risk that people skip `topo health`, I could addend "For detailed usage instructions, see Getting Started" after the intro?
2. I don't think we can deprecate `topo clone`. It's a thin wrapper that couples two commands together, but given the likelihood that those commands are run together, it does feel like a time saving abstraction. It also plays nicely with the hooks from other websites.

federicobozzini:
> @federicobozzini
> 
> 1. The focus here is on trying to highlight the commands which communicate the value proposition. `topo health` is something that helps you use topo, but brings no value in of itself. To mitigate the risk that people skip `topo health`, I could addend "For detailed usage instructions, see Getting Started" after the intro?
> 2. I don't think we can deprecate `topo clone`. It's a thin wrapper that couples two commands together, but given the likelihood that those commands are run together, it does feel like a time saving abstraction. It also plays nicely with the hooks from other websites.

I agree with 1. I'm ont sure about 2 instead. I would argue that maybe `git clone` still belongs to the discovery phase and `topo configure` alone is a better fit for the configure stage.

PR #399, opened 2026-08-07
fix: make compose file parsing env var aware

awphi:
## Changes

- Passes env vars down from the environment or compose-sibling `.env` file to the manual file parsing we do to extract image names, build `platform`s and service names by setting the relevant options in our invocation of the compose go lib.

## Checklist


yejseo01:
does this fix C's issue yesterday?

yejseo01 (approved):
LGTM! Just one question.

awphi:
> does this fix C's issue yesterday?

Indeed it does!

PR #402, opened 2026-08-10
fix: single-quote known_hosts entries in fix commands

awphi:
## Background

> All tested on Ubuntu 25 which contains [this downstreamed patch](https://sources.debian.org/patches/openssh/1%3A7.9p1-10%2Bdeb10u2/mention-ssh-keygen-on-keychange.patch/) that adds `-f` to the `ssh-keygen` help string. It's unrelated to the problem/fix but thought it was interesting!

Currently, the suggested fix quotes the known_hosts entry to avoid shells interpreting the square brackets in [ip]:port e.g. the user sees:
```
  Fix:
    Remove the old SSH host key from known_hosts, then retry
  Command:
    ssh-keygen -R "[x.y.z.w]:2222"
```

However, when you set -o json, since the value is itself in a JSON string and we use Go's `%q` fmt string quoting, the quotes get escaped:
```
      "fix": {
        "description": "Remove the old SSH host key from known_hosts, then retry",
        "command": "ssh-keygen -R \"[x.y.z.w]:2222\""
      }
```
Thus if a user copy-pastes out of the JSON output, the command will fail and if you're not well-versed in the escaping semantics of your shell of choice or not paying too much attention, it's hard to tell why:
```
> ssh-keygen -R \"[x.y.z.w]:2222\"
Host "[x.y.z.w]:2222" not found in /home/adahod01.guest/.ssh/known_hosts
```

## Changes

- Swaps the shell-escaping in the fix command to use the single quote arg util and not double. Adjusts the existing `command.QuoteArg` to quote when square brackets are included as these gets interpreted by shells like zsh. This means the JSON output fix can still be copy-pasted!

As a result, this does mean the copy-pastable command will not work in `cmd.exe` as it does not support single-quote escaping. However, we already use this quoting elsewhere and don't support installation of Topo from `cmd.exe` so this seems like a fair compromise.

## Checklist

### DIFF
--- a/internal/command/command.go
+++ b/internal/command/command.go
@@ -24,7 +24,7 @@ func QuoteArg(s string) string {
 		return "''"
 	}
 
-	if strings.ContainsAny(s, " \t\n\"'\\$`") {
+	if strings.ContainsAny(s, " \t\n\"'\\/[/]$`") {
 		return "'" + strings.ReplaceAll(s, "'", `'"'"'`) + "'"
 	}
 
--- a/internal/command/command_test.go
+++ b/internal/command/command_test.go
@@ -55,6 +55,11 @@ func TestQuoteArg(t *testing.T) {
 			arg:  `C:\Users\runner\AppData\Local\Temp\custom-compose.yaml`,
 			want: `'C:\Users\runner\AppData\Local\Temp\custom-compose.yaml'`,
 		},
+		{
+			name: "arg with square brackets",
+			arg:  "custom[compose].yaml",
+			want: "'custom[compose].yaml'",
+		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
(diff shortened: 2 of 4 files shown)


internal/command/command.go
```diff
@@ -24,7 +24,7 @@ func QuoteArg(s string) string {
 		return "''"
 	}
 
-	if strings.ContainsAny(s, " \t\n\"'\\$`") {
+	if strings.ContainsAny(s, " \t\n\"'\\/[/]$`") {
```
th3james, line 27: Could we add a test case for this please

th3james (approved):

internal/command/command.go
awphi: https://github.com/arm/topo/pull/402/commits/3f688faf2f6cd4bc0c5c885686e66ccab939c62a

PR #404, opened 2026-08-11
feat!: allow skipping optional params at configure time

awphi:
## Changes

- Allows the omission of optional parameters when running `topo configure` - both interactively and non-interactively with CLI args
- Renames "argument" in an error message to "parameter" as that's the current terminology. Not strictly atomic so can split if contentious.

Breaking as previously any omitted optional parameters would just be set to the topo-defined default value.

Note this doesn't affect required parameters, even if already set. This is because to achieve this we need to parse the compose file and grab all the current values but currently `topo configure` (and clone) already open the file twice in different ways each with their own issues. Suggest refactoring to read once up front, inspect, mutate and write before thinking about this. It'd also be nice to display the current build arg value(s) for a given parameter when running interactive `topo configure`.

Screenshot 2026-08-11 at 16 02 45

## Checklist

### DIFF
--- a/e2e/configure_test.go
+++ b/e2e/configure_test.go
@@ -76,7 +76,7 @@ func TestConfigure(t *testing.T) {
 		out, err := cmd.CombinedOutput()
 
 		require.Error(t, err)
-		assert.Contains(t, string(out), "unknown argument: UNKNOWN")
+		assert.Contains(t, string(out), "unknown parameter: UNKNOWN")
 		got := testutil.RequireReadFile(t, composePath)
 		assert.Equal(t, original, got)
 	})
--- a/internal/arguments/cli_provider.go
+++ b/internal/arguments/cli_provider.go
@@ -36,7 +36,7 @@ func (p *CLIProvider) Provide(args []Arg) ([]ResolvedArg, error) {
 
 	for key := range p.input {
 		if !seen[key] {
-			return nil, fmt.Errorf("unknown argument: %s", key)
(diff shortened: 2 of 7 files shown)


federicobozzini (commented):
Two comments here:

1. This PR seems to break some examples. As an example just running `topo clone git@github.com:Arm-Examples/topo-welcome` and pressing enter when ask about `GREETING_NAME` will fail the clone. I just want to understand if this is intentional or if we need to update the examples too.
2. We might need to update the [project schema](https://github.com/arm/topo/blob/e3ca31b12818be583707c6db656053a996d17ff4/docs/project-specification/schema/topo-project-specification.json#L51)

awphi:
> This PR seems to break some examples. As an example just running topo clone git@github.com:Arm-Examples/topo-welcome and pressing enter when ask about GREETING_NAME will fail the clone. I just want to understand if this is intentional or if we need to update the examples too.

Mm good point - not intentional but does open a can of worms that requires some thought. Was hoping to avoid having to deal with required values here since _I think_ they will require awareness of current values. Will re-draft and consider a new approach. Thanks for catching, had only played with hand-rolled projects when testing.

> We might need to update the [project schema](https://github.com/arm/topo/blob/e3ca31b12818be583707c6db656053a996d17ff4/docs/project-specification/schema/topo-project-specification.json#L51)

Yep good point!

awphi:
Closing as think this requires a different, more involved approached that will be easier to do after #407 is merged

PR #407, opened 2026-08-12
fix: ensure topo parameters are always applied to extending services

awphi:
## Background

Currently when using `topo configure` (or clone) on a project which has services that extend other services, we create build args inconsistently e.g.

```yaml
services: 
  a-service: 
    platform: linux/arm64
    extends: 
      service: another-service
    build: 
      args: 
        SOMETHING: "test"
  a-second-service: 
    platform: linux/arm64
    extends: 
      service: another-service

x-topo: 
  parameters: 
    PARAM: 
      description: "a param"
```

Running `topo configure PARAM=a` here will yield the following result:

```yaml
services:
  a-service:
    platform: linux/arm64
    extends:
      service: another-service
    build:
      args:
        SOMETHING: "test"
  a-second-service:
    platform: linux/arm64
    extends:
      service: another-service
    build:
      args:
        PARAM: a
x-topo:
  parameters:
    PARAM:
      description: "a param"
```

The param was applied to `a-second-service` but not `a-service` because the latter already had some `build.args`! This could make `topo configure` unusable for more complex setups.

This PR ignores the fact that we indiscriminately write to services with `extends` clause regardless if the service being extended actually needs the arg because that's a can of worms that's not worth opening atm.

## Changes

- Ensures we always add topo parameters to extending services regardless of the state of their current `build.args` block with a small refactor
  - We now ensure the existence of a `build.args` block in the outer loop but actually do the addition at the same time as mutating the args of each service, making the flow of the arg application more linear.

To address the indiscriminate writing to `compose.yaml` which the change exacerbates we can consider:
- Avoiding applying args to extending services entirely which I'm not sure we should be so opinionated about
- Parsing the compose project more intelligently to work out if the arg is specified anywhere in a service's inheritance stack. I _think_ we could achieve this with the compose Go library. Also presents an opportunity to ditch some custom yaml parsing/writing stuff which is a larger refactor hence not attempted here.

## Checklist

### DIFF
--- a/internal/compose/yaml.go
+++ b/internal/compose/yaml.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"fmt"
 	"io"
-	"sort"
 	"strings"
 
 	"github.com/arm/topo/internal/output/logger"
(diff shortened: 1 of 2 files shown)


awphi:
Closing in favour of doing more thorough checks on extending services to check for only necessary overrides

PR #409, opened 2026-08-13
fix: only apply necessary args to extending services when configuring

awphi:
## Background

Take the following Topo project:

```yaml
# compose.yaml
services: 
  child-a: 
    extends: 
      service: parent-a
      file: compose.other.yaml
    build: 
      args: 
        SOMETHING: "something"

  child-b: 
    extends: 
      service: parent-b
      file: compose.other.yaml

x-topo: 
  parameters: 
    PARAM: 
      description: "Parameter"
```

```yaml
# compose.other.yaml
services: 
  parent-a: 
    platform: linux/arm64
    build: 
      args: 
        PARAM: "param"

  child-b: 
    platform: linux/arm64
    extends: 
      service: parent-b
      file: compose.other.yaml
```

Running `topo configure` and plugging in a `PARAM` will yield:

```yaml
# compose.yaml
services:
  child-a:
    extends:
      service: parent-a
      file: compose.other.yaml
    build:
      args:
        SOMETHING: "something"
  child-b:
    extends:
      service: parent-b
      file: compose.other.yaml
    build:
      args:
        PARAM: a
x-topo:
  parameters:
    PARAM:
      description: "Parameter"
```

Two issues:
- `child-a` needs a `PARAM` override and didn't get one!
- `child-b` didn't need a `PARAM` override and did get one!

## Changes

- Does a second full de-normalized `ReadProject` to collapse extending services to work out if they need a build arg overriding or not.
- Renames `yaml.go` -> `args.go` and changes its public API to just expose `ApplyArgs` to expose a thinner, deeper module with a neater signature and less file open/close ceremony required by the caller. This pushes this change to a hybrid refactor-fix but I think the net result of fewer lines of code despite a bug fix + new tests makes it worth it!

## Checklist


awphi:
Been re-thinking this and have been wondering if we should actually do this. It's quite complex and a bit magic.

Perhaps a stupider but more predictable thing would be drop the `extends` handling entirely and require you to specify build arg overrides in your primary `compose.yaml` project if you want `topo configure` to reason about them. We **never** add a `build.args` block for you. I think this is fine given we've removed `topo extend` and is more predictable and explainable.

If you want your parameter to apply to an extending service, then add a build arg override in your `compose.yaml`. 

Going to close this for now and raise a PR that does this, linking back to here as an alternative.

PR #412, opened 2026-08-17
feat!: avoid adding build args to services using `extends`

awphi:
## Background

See #409. This PR presents a simpler alternative to the solution implemented in #409 where we never add `build.args` to any service and only mutate existing ones. This is a simpler, more predictable and explainable solution.

## Changes

- Avoids adding `build.args` to `extends`ing services, instead warning the user on how this works

I'm open to this approach or something like #409 to address the root problem but this PR is my personal preference.

## Checklist

### DIFF
--- a/internal/compose/yaml.go
+++ b/internal/compose/yaml.go
@@ -49,7 +49,7 @@ func ApplyArgs(root *yaml.Node, toApply map[string]string) error {
 		if args == nil {
 			if find(svc, "extends") != nil {
 				name := services.Content[i].Value
-				logger.Warn(fmt.Sprintf("service %q uses `extends` without `build.args`; inherited build args will not be configured. Declare the required args on this service, or set `build.args: {}` to silence this warning", name))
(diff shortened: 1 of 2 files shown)


internal/compose/yaml.go
```diff
@@ -44,27 +43,15 @@ func ApplyArgs(root *yaml.Node, toApply map[string]string) error {
...
-		}
-
 		args := find(build, "args")
+
 		if args == nil {
-			if !hasExtends {
-				continue
+			if find(svc, "extends") != nil {
+				name := services.Content[i].Value
+				logger.Warn(fmt.Sprintf("service %q uses `extends` without `build.args`; inherited build args will not be configured. Declare the required args on this service, or set `build.args: {}` to silence this warning", name))
```
muchzill4: This is weird. Starts lowercase, but then we have a sentence?

muchzill4 (approved):

internal/compose/yaml.go
awphi: Good spot - have de-sentencified https://github.com/arm/topo/pull/412/commits/64392ab7a27e50f4040722cc9f4ed05b8ad6913f

PR #413, opened 2026-08-18
feat: support compose-style interpolation in x-topo attributes

awphi:
## Changes

- Adds support for [compose-style env var interpolation](https://docs.docker.com/reference/compose-file/interpolation/) inside `x-topo.deployment_success_message` 
- Injects two new env vars when running `compose.ReadProject` in the post deploy step:
   - `TOPO_TARGET`: the selected target for deploy/stop/ps
   - `TOPO_TARGET_HOSTNAME`: the `ssh -G`-resolved `HostName` of the selected target

```yaml
name: test-project
services:
  app:
    image: nginx
x-topo:
  deployment_success_message: "${COMPOSE_PROJECT_NAME} is available at http://${TOPO_TARGET_HOSTNAME}:${APP_PORT:-8080}"
```

## Checklist

### DIFF
--- a/docs/project-specification/01-authoring-projects.md
+++ b/docs/project-specification/01-authoring-projects.md
@@ -75,6 +75,20 @@ x-topo:
       example: string # Optional
 ```
 
+### Deployment success message
+
+`deployment_success_message` supports standard Compose interpolation. Topo provides these additional variables when it displays the message:
+
+- `TOPO_TARGET`: The complete SSH destination in URI form, such as `ssh://user@board.local`.
+- `TOPO_TARGET_HOSTNAME`: The Target hostname resolved from the SSH configuration, such as `board.local`.
+
+```yaml
+x-topo:
+  name: "web-app"
+  deployment_success_message: |
+    Open http://${TOPO_TARGET_HOSTNAME:-localhost}:8080
+```
+
 ---
 
 ## 4. Testing Your Project
--- a/internal/compose/compose.go
+++ b/internal/compose/compose.go
@@ -45,18 +45,18 @@ func ReadProject(targetProjectFile string) (*types.Project, error) {
 	return readProject(targetProjectFile, nil)
 }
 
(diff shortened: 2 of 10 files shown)


docs/project-specification/01-authoring-projects.md
```diff
@@ -75,6 +75,33 @@ x-topo:
...
+    ports:
+      - "${WEB_PORT:-8080}:80"
+
+x-topo:
+  name: "web-app"
+  deployment_success_message: |
+    Open http://${TOPO_TARGET_HOSTNAME:-localhost}:${WEB_PORT:-8080}
+```
+
+The `localhost` default keeps the Project usable with plain `docker compose`, where Topo does not provide `TOPO_TARGET_HOSTNAME`.
```
muchzill4: `docker compose` doesn’t read `x-topo` anyway, so no need for a default.

muchzill4:
Wonder if we should scope it to `success_message` for now. It’s nice to keep the behaviour consistent, but using `TOPO_TARGET_HOSTNAME` anywhere else sounds a bit nuts.

internal/compose/compose.go
```diff
@@ -42,12 +42,21 @@ func PullableServices(composeFilePath string) ([]string, error) {
 }
 
 func ReadProject(targetProjectFile string) (*types.Project, error) {
+	return readProject(targetProjectFile, nil)
+}
+
+func ReadProjectWithEnvironment(targetProjectFile string, environment []string) (*types.Project, error) {
```
muchzill4: Nit: should we do `WithEnv`? `Env` seems to be prevalent in go-land.

muchzill4 (approved):

internal/compose/compose.go
awphi: https://github.com/arm/topo/pull/413/commits/b943e776769f49d268311dd70b67d14c4ef1208e

PR #414, opened 2026-08-18
feat!: skip updating optional or already populated required parameter values when skipped

awphi:
### The problem
Take the following project:
```yaml
# compose.yaml
services:
  my-service:
    build:
      args: ["BAR=nondefault", "FOO=whatever"]	  

x-topo:
  parameters:
    BAR:
      default: 'default-bar'
    FOO:
      default: 'default-foo'
```
I've configured my `BAR` param value to a non-default. I want to now re-configure but adjust my `FOO` param value. This is really annoying in Topo today because when a parameter specified a `default` and you press enter on the 'enter a value for the param' prompt, it reverts to the default! Same goes for omitting a parameter in non-interactive mode.

## Changes
- Remove the `default` param field altogether. It's already mostly duplication because it *should* be specified in the build args themselves anyway to make the project `git clone`-able.

This avoids a complicated TUI (the alternative that we previously discussed) where we would have to present the user with 3 options (enter new value, use default, leave unchanged) and keeps the prompt simple:
- Press enter to leave build args of that param unchanged, **always**
- Enter a new value to update all build args of that param

This simpler flow has a couple of advantages:
- Easier to understand + document within the CLI
- Allows for manual 'configuration' of build args in place via file editing w/o risk that Topo might blat your value out of existence
- Gives Project authors one less thing to worry about and potentially get wrong
- Generally feels more predictable by virtue of not trying to be smarter than the user
- Less clash between `example` and `default` which are generally duplicates atm. `example` remains as useful way to revert params to sensible Project author defined value if the user borks their configure.

## Checklist


docs/project-specification/schema/topo-project-specification.json
```diff
@@ -44,11 +44,11 @@
                   },
                   "required": {
                     "type": "boolean",
-                    "description": "If `true`, Implementations must enforce input or error"
+                    "description": "If `true`, Implementations must enforce that all build args values are set for this parameter when configuring the Project"
                   },
                   "default": {
                     "type": "string",
-                    "description": "Value used if user skips input (only valid when not required)"
+                    "description": "Deprecated. Current Topo versions ignore this property."
```
muchzill4: Should we be specific? 🤔 `>x.x.x`

th3james (approved):
LGTM

docs/project-specification/schema/topo-project-specification.json
awphi: https://github.com/arm/topo/pull/414/commits/80d48872d405211b8eebf64936c2a6bac624c020

muchzill4 (commented):
<!-- Uploading "Screenshot 2026-08-20 at 09.59.42.png"... —>

Is this weird? Specifically the `(required, press Enter to skip)` bit.

I almost wonder if we should have something like `"press enter to keep current value"` in this case.

PR #418, opened 2026-08-20
refactor: inject Topo-generated env vars into process environment once at start-up

awphi:
## Background

#413 introduced compose-style interpolation in `deployment_success_message`. Originally this was achived by creating env vars in the CLI and injecting them into all `docker compose` subprocesses and `compose-go` calls. We downscoped this to only inject at call sites where it was needed i.e. `post_deploy.go`. 

This has had the unfortunate consequence of causing `docker compose` and `compose-go` calls to raise a bunch of warnings when your `x-topo` block contains one of these env vars without a fallback value e.g.

```yaml
# compose.yaml
services:
  welcome-service:
    platform: linux/arm64
    image: busybox:latest

x-topo:
  deployment_success_message: "Welcome service deployed successfully! Access it at http://${TOPO_TARGET_HOSTNAME}:8080"
```

```
❯ go run ../cmd/topo deploy --target localhost
WARN[0000] The "TOPO_TARGET_HOSTNAME" variable is not set. Defaulting to a blank string.
11:10:46 WARN registry transfer is not yet supported with this configuration. Falling back to direct transfer.

┌─ Build images ────────────────────────────────────────
WARN[0000] The "TOPO_TARGET_HOSTNAME" variable is not set. Defaulting to a blank string.
WARN[0000] No services to build

┌─ Pull images ─────────────────────────────────────────
WARN[0000] The "TOPO_TARGET_HOSTNAME" variable is not set. Defaulting to a blank string.
WARN[0000] The "TOPO_TARGET_HOSTNAME" variable is not set. Defaulting to a blank string.
[+] pull 1/1
 ✔ Image busybox:latest Pulled                                                                                        0.7s

┌─ Start services ──────────────────────────────────────
WARN[0000] The "TOPO_TARGET_HOSTNAME" variable is not set. Defaulting to a blank string.
[+] up 1/1
 ✔ Container test-workspace-welcome-service-1 Started                                                                 0.1s

┌─ Deployment Success ──────────────────────────────────
Welcome service deployed successfully! Access it at http://localhost:8080/
```

## Changes

- Reverts back to the original behaviour of #413 with a new implementation - we set the env vars once when parsing the `--target` flag or env var for the entire process. This means the injected env vars are passed to call `compose-go` calls (that use `WithOsEnv`) and subprocess invocations.

## Checklist


cmd/topo/root.go
```diff
@@ -112,6 +112,11 @@ func requireTarget(cmd *cobra.Command) (string, error) {
 	if !exists {
 		return "", fmt.Errorf("target not specified: use --target with an SSH destination (e.g. user@example.local) or set %s", targetEnvVar)
 	}
+	for name, value := range env.TargetEnv(t) {
```
muchzill4: I wonder if this is the right place to do this. This only affects flows that read the compose file, right? Is that only `deploy`? Perhaps we should have `env.SetTargetVars(t)` only in relevant commands after `requireTarget` is called?

muchzill4 (commented):
I’m still split between this being horrible and or the best thing ever. On one hand we have global env that gets propagated as such to compose, which is a superset of what this does. On the other, it's implicit arguments between layers and would not work if topo was a web service (it is not!).

I think it is the right call (for now at least!). Let’s see if we find this foot-gunny.

cmd/topo/root.go
awphi: > This only affects flows that read the compose file, right? Is that only deploy? Perhaps we should have env.SetTargetVars(t) only in relevant commands after requireTarget is called?

Yep, also `stop` but agree that doing it here overreaches. Moved here https://github.com/arm/topo/pull/418/commits/97ca818da9b5e44dec132454a720efd346afba64

muchzill4 (approved):

PR #419, opened 2026-08-20
docs: clarify copy in interactive configure provider

awphi:
## Changes

- 'Press Enter' isn't actually right since you press enter after typing a value which changes the current.

## Checklist

### DIFF
--- a/internal/arguments/interactive_provider.go
+++ b/internal/arguments/interactive_provider.go
@@ -52,7 +52,7 @@ func (p *InteractiveProvider) Provide(args []Arg) ([]ResolvedArg, error) {
 			label = "required"
 		}
 
-		_, err = fmt.Fprintf(p.output, "%s (%s, press Enter to skip)> ", arg.Name, label)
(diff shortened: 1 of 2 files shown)


internal/arguments/interactive_provider.go
```diff
@@ -52,7 +52,7 @@ func (p *InteractiveProvider) Provide(args []Arg) ([]ResolvedArg, error) {
 			label = "required"
 		}
 
-		_, err = fmt.Fprintf(p.output, "%s (%s, press Enter to skip)> ", arg.Name, label)
+		_, err = fmt.Fprintf(p.output, "%s (%s, leave blank to keep current)> ", arg.Name, label)
```
muchzill4: If current isn’t there, it’d be weird to mention it. Do we need something like `if len(arg.CurrentValues)` here too?

awphi: Good point! https://github.com/arm/topo/pull/419/commits/87e8ee8101668fe65a1e8672faf9d81db55fd5fd

I hate `CurrentValues` so much 😭

muchzill4 (approved):

PR #420, opened 2026-08-20
refactor: use descriptive verbs for SSH config construction

th3james:
## Changes

I noticed that `ssh.NewConfig` has a stateful side effect (exec'ing `ssh`) and discarded its errors. Because stateful calls are so critical for performance and reliability, I prefer it when function names make it clear that they are stateful, which, IMHO `NewConfig` does not. I recognise this may go slightly against the conventions for the use of `New` for constructors in this repo, so happy to debate this.

- Renames: `ssh.NewConfig` -> `ssh.LoadConfig`
- Renames: `ssh.NewConfigFromBytes` -> `ssh.ParseConfig`
- Repurposes the previously unused `ResolveHostName` function, and then uses it
- Makes `ssh.LoadConfig` return an error, forcing callers to handle it

## Checklist

### DIFF
--- a/internal/env/target.go
+++ b/internal/env/target.go
@@ -14,9 +14,17 @@ const (
 
 func SetTargetEnv(target string) error {
 	destination := ssh.NewDestination(target)
+	hostName := destination.Host
+	if !destination.IsPlainLocalhost() {
+		sshConfig, err := ssh.LoadConfig(destination)
+		if err != nil {
+			return err
+		}
+		hostName = sshConfig.HostName
+	}
 	vars := map[string]string{
 		TargetVariable:         destination.String(),
-		TargetHostnameVariable: ssh.NewConfig(destination).HostName,
+		TargetHostnameVariable: hostName,
 	}
 	for name, value := range vars {
 		if err := os.Setenv(name, value); err != nil {
--- a/internal/env/target.go
+++ b/internal/env/target.go
@@ -14,13 +14,13 @@ const (
 
 func SetTargetEnv(target string) error {
 	destination := ssh.NewDestination(target)
-	hostName, err := ssh.ResolveHostName(destination)
+	hostname, err := ssh.ResolveHostname(destination)
 	if err != nil {
 		return err
 	}
 	vars := map[string]string{
 		TargetVariable:         destination.String(),
-		TargetHostnameVariable: hostName,
+		TargetHostnameVariable: hostname,
 	}
 	for name, value := range vars {
 		if err := os.Setenv(name, value); err != nil {
--- a/internal/env/target.go
+++ b/internal/env/target.go
@@ -14,13 +14,9 @@ const (
 
 func SetTargetEnv(target string) error {
 	destination := ssh.NewDestination(target)
-	hostName := destination.Host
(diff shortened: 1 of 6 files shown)


internal/env/target_test.go
```diff
@@ -10,6 +10,18 @@ import (
...
 func TestSetTargetEnv(t *testing.T) {
+	t.Run("sets plain localhost without querying SSH config", func(t *testing.T) {
+		t.Setenv("PATH", "")
+		t.Setenv(env.TargetVariable, "ssh://stale.example")
+		t.Setenv(env.TargetHostnameVariable, "stale.example")
+
+		err := env.SetTargetEnv("localhost")
+
+		require.NoError(t, err)
+		assert.Equal(t, "ssh://localhost", os.Getenv(env.TargetVariable))
```
th3james, line 21: This codifies the existing behaviour, but does call into question whether or not this is desirable

awphi: What's undesirable? The prepending of `ssh://`? If yes I've got no strong opinions but an argument in favour of leaving it is you want to write something like:
```yaml
# compose.yaml
x-topo:
  deployment_success_message: "Check out your new deployed application on the target with `ssh ${TOPO_TARGET} 'cat /my/cool/thing'`"
```

This wouldn't work anymore if we removed the prepending since Topo happily accepts things like `--target localhost:2222`, an invalid `ssh` destination.

internal/ssh/config.go
```diff
@@ -125,10 +124,10 @@ func IsExplicitHostConfig(host string, config []byte) bool {
 	return false
 }
 
-func readConfig(dest Destination) ([]byte, error) {
-	return readConfigContext(context.Background(), dest)
+func queryConfig(dest Destination) ([]byte, error) {
```
muchzill4, line 127: NIT: I think `read` was more correct. We’re not querying for anything.

internal/env/target.go
```diff
@@ -14,9 +14,13 @@ const (
 
 func SetTargetEnv(target string) error {
 	destination := ssh.NewDestination(target)
+	hostName, err := ssh.ResolveHostName(destination)
```
muchzill4: While we’re here, would it be a big ask to make `hostname` casing consistent? I noticed we’re doing `hostname` and `hostName`. Golang mostly does `hostname`.

muchzill4 (approved):
👍

internal/env/target_test.go
th3james: This really serves to highlight how difficult it is to provide the same abstraction over both remote and local targets. I don't quite know what the right answer is yet, it's probably out of scope for this PR though.

internal/ssh/config.go
th3james: IMHO `read` suggests directly accessing a passive thing. We're asking (querying) the SSH command to read, parse, and resolve the config on disk.

internal/env/target.go
th3james: https://github.com/arm/topo/pull/420/commits/7799f1436e75c815b0f1c6713ff0b21ede74d650

PR #425, opened 2026-08-24
fix: support `runtime:` when deploying via Podman

muchzill4:
## Changes

- Use `podman-compose` instead of `docker-compose`, which does not support `runtime:` when invoked through Podman.
- Pass remote endpoints through `CONTAINER_HOST` and remove Docker-compatible local socket discovery (so much red!).
- Install and verify `podman-compose` in Linux and Windows CI.

## Checklist


awphi (approved):
So much red! `podman-compose` may not be as featureful as `docker-compose` but this is certainly an attractive diff

PR #427, opened 2026-08-24
docs: docker install guide

awphi:
## Changes

- Adds a container engine install guide. Right now very Docker scoped since Podman is not shipped.
- Links to the above from README (so we have an anchor to point to from `topo health`) and the install Topo page

## Why the choices of recommendation?

- Docker Desktop - default choice for most
- Colima - just works. Have used since day 1 on Topo and no problems. 
- Rancher Desktop - very slick to use, especially on Windows. Takes away headache of maintaining instruction on how to keep your WSL2 instance alive + how to forward a socket out of the VM to Windows-land.
- Docker Engine CE on Linux - what else?

One key bit we miss is a FOSS Windows-on-Arm story. Rancher Desktop is considering/working on this. Docker Desktop already supports it so felt we were well-covered.

## Testing

Manually Tested:
- Colima on macOS (I've only ever used Topo with this)
- Rancher Desktop on Windows x64 with WSL2 set-up but no docker
- Rancher Desktop on macOS
- Rancher Desktop on Windows x64 **without** existing WSL2
- Docker Engine on a clean Ubuntu VM AArch64
- Docker Desktop on Windows AArch64

Untested but high confidence given the above tests:
- Colima on Linux
- Rancher Desktop on Linux

## Checklist


docs/introduction/container-engines.md
```diff
@@ -0,0 +1,103 @@
...
+
+:::note
+
+Review the [Docker Desktop license terms](https://docs.docker.com/subscription/desktop-license/) before installation. Some organizations and types of commercial use require a paid subscription.
+
+:::
+
+Follow the [Docker Desktop installation instructions](https://docs.docker.com/desktop/setup/install/) for your host. On Windows, configure Docker Desktop to use Linux containers.
+
+Start Docker Desktop before you use Topo, and keep it running. Signing in to Docker Desktop is optional unless your organization requires it. Sign in to [increase Docker Hub pull limits, access private images, or apply organization security policies](https://docs.docker.com/desktop/setup/sign-in/).
```
muchzill4: Is this needed? My thinking: what’s the minimal set of cromulent instructions. Minimal = less maintenance.

docs/introduction/container-engines.md
```diff
@@ -0,0 +1,103 @@
...
+
+:::note
+
+On Windows, install [Windows Subsystem for Linux 2 (WSL 2)](https://aka.ms/wslinstall) before you install Rancher Desktop.
+
+:::
+
+Follow the [Rancher Desktop installation instructions](https://docs.rancherdesktop.io/getting-started/installation/) for your host. When Rancher Desktop starts for the first time, select **dockerd (moby)** as the container engine. Topo does not require Kubernetes. You can change these settings later in **Preferences**.
+
+On macOS and Linux, select **Automatic** to add the Rancher Desktop tools to `PATH`. If you select **Manual**, add `~/.rd/bin` to `PATH` yourself.
```
muchzill4: This also feels a bit too detailed for our docs, thought I can be convinced otherwise.

muchzill4 (approved):
Really great piece of work. Useful as is and happy to get it in.

I do have some reservations about long term maintenance and the amount of detail. I wonder if it’s worth us hand-holding people as much for various installation methods. I’d almost go to the other extreme and _just list_ the suggestions. Consider Docker work arounds “advanced” use case, and let people figure it out. Also curious if we could push `topo health` to provide guidance in case of broken setup.

docs/introduction/container-engines.md
awphi: I was trying to address the pitfall we've seen a couple users fall into where `topo deploy` fails due to Docker Hub rate limiting or SSO enforcement or whatever. Though iirc the error was pretty clear so will remove.

awphi:
> I’d almost go to the other extreme and just list the suggestions.

Also been thinking about this morning. My other gripe with this doc in its current state is its unclear which of the alternatives we recommend for each OS. 

I've refactored pretty heavily here to just a table mapping OS/Arch -> recommendation. This shifts maintenance away from detailed install instructions to picking a good container engine for Topo users which feels more in our remit. https://github.com/arm/topo/pull/427/commits/495613476e5f80792649db176fe3aa7251215907

muchzill4 (approved):
Very nice.

PR #432, opened 2026-08-27
fix: preserve line endings when cloning projects

federicobozzini:
## Changes

- invoke Git clone and checkout with `-c core.autocrlf=false`

## Checklist

### DIFF
--- a/internal/project/source_test.go
+++ b/internal/project/source_test.go
@@ -184,6 +184,34 @@ func TestGitSource(t *testing.T) {
 			require.NoError(t, err)
 			assert.Equal(t, firstCommit, gitOutput(t, dstDir, "rev-parse", "HEAD"))
 		})
+
+		t.Run("preserves line endings when Git enables autocrlf", func(t *testing.T) {
+			const script = "#!/bin/sh\nexit 0\n"
+			repoDir, commit := createGitRepoWithFile(t, "entrypoint.sh", script)
+			t.Setenv("GIT_CONFIG_COUNT", "1")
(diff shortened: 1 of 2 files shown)


internal/project/source_test.go
```diff
@@ -357,6 +382,29 @@ func createGitRepoWithTwoCommits(t *testing.T) (string, string) {
...
+
+func createGitRepo(t *testing.T) string {
+	t.Helper()
+
+	repoDir := t.TempDir()
+	runGit(t, repoDir, "init")
+	runGit(t, repoDir, "config", "user.name", "Topo Test")
+	runGit(t, repoDir, "config", "user.email", "topo@example.com")
+	runGit(t, repoDir, "config", "commit.gpgSign", "false")
+	runGit(t, repoDir, "config", "core.autocrlf", "false")
```
awphi, line 404: Explicitly setting gpgSign and autocrlf here seem unrequired/undesirable? We explicitly disable autocrlf in the relevant tests with env vars (which is very neat!) and gpg signing is off by default I think?

awphi (approved):
One nit!

internal/project/source_test.go
federicobozzini: These are here to avoid any possible conflict with different settings on the dev machine. I think gpgSign I think is especially necessary here because if set to true tests can fail if no key is available.

autocrlf can probably be removed as it gets set to true anyway in the test, but this is just a way to make sure tests always start from a known state.

PR #438, opened 2026-09-01
feat: support `topo stop` with `--engine podman`

muchzill4:
## Changes

- Add experimental `topo stop --engine podman` support.

## Checklist

### DIFF
--- a/internal/deploy/podman/stop.go
+++ b/internal/deploy/podman/stop.go
@@ -1,0 +1,32 @@
+package podman
+
+import (
+	"context"
+	"errors"
+	"io"
+
+	"github.com/arm/topo/internal/output/term"
+	"github.com/arm/topo/internal/ssh"
+)
+
+func Stop(ctx context.Context, output io.Writer, composeFile string, target ssh.Destination) (stopErr error) {
+	if err := term.PrintHeader(output, "Stop services"); err != nil {
+		return err
+	}
+
+	socket := LocalSocket
+	if target.IsPlainLocalhost() {
+		return RunComposeCommand(ctx, output, socket, composeFile, "stop")
+	}
+
+	tunnel, err := TunnelRemoteSocketPath(ctx, output, target)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		stopErr = errors.Join(stopErr, closeRemoteTunnel(tunnel))
+	}()
+
+	socket = NewSocket(tunnel.SocketURL())
+	return RunComposeCommand(ctx, output, socket, composeFile, "stop")
+}
--- a/internal/deploy/podman/stop.go
+++ b/internal/deploy/podman/stop.go
@@ -10,23 +10,25 @@ import (
 )
 
 func Stop(ctx context.Context, output io.Writer, composeFile string, target ssh.Destination) (stopErr error) {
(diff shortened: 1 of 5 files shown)


internal/deploy/podman/stop.go
```diff
@@ -0,0 +1,32 @@
...
+	if err := term.PrintHeader(output, "Stop services"); err != nil {
+		return err
+	}
+
+	socket := LocalSocket
+	if target.IsPlainLocalhost() {
+		return RunComposeCommand(ctx, output, socket, composeFile, "stop")
+	}
+
+	tunnel, err := TunnelRemoteSocketPath(ctx, output, target)
```
federicobozzini: NIT: [when deploying we tell the user about the socker SSH tunnel](https://github.com/muchzill4/topo/blob/d86d1a0f8f0e5c4a84b077d5438f9d8dc288ba52/internal/deploy/podman/deploy.go#L50). Should we do the same here?

internal/deploy/podman/stop_test.go
```diff
@@ -0,0 +1,42 @@
...
+func TestStop(t *testing.T) {
+	requireLocalPodman(t)
+
+	t.Run("stops services on localhost", func(t *testing.T) {
+		composeFile, projectName := deploymentFixture(t)
+		t.Cleanup(func() { cleanupComposeProject(t, composeFile) })
+		options := podman.DeployOptions{TargetHost: ssh.PlainLocalhost}
+		require.NoError(t, podman.Deploy(t.Context(), io.Discard, composeFile, options))
+
+		err := podman.Stop(t.Context(), io.Discard, composeFile, ssh.PlainLocalhost)
```
federicobozzini: Should we be using t.Output() here instead of io.Discard to get some clear errors?

federicobozzini (approved):
Couple of comments but nothing blocking. Looks good to me 👍

internal/deploy/podman/stop_test.go
muchzill4: Nice catch: 76e2790

internal/deploy/podman/stop.go
```diff
@@ -0,0 +1,32 @@
...
+	"io"
+
+	"github.com/arm/topo/internal/output/term"
+	"github.com/arm/topo/internal/ssh"
+)
+
+func Stop(ctx context.Context, output io.Writer, composeFile string, target ssh.Destination) (stopErr error) {
+	if err := term.PrintHeader(output, "Stop services"); err != nil {
+		return err
+	}
```
awphi: It's a little odd we print this header for a single operation command. I guess we do the same with `topo stop --engine docker` but something to consider removing in both paths?

awphi (approved):

internal/deploy/podman/stop.go
muchzill4: f0baec1

internal/deploy/podman/stop.go
muchzill4:
```
┌─ Stop services ───────────────────────────────────────
[+] stop 1/1
 ✔ Container topo-welcome-app-1 Stopped   
 ```
 
 It's kinda nice.

PR #439, opened 2026-09-02
feat: add git to topo health checks

federicobozzini:
## Changes

- what title says

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,7 @@ Not sure what these terms mean? The [glossary](docs/introduction/glossary.md) de
 **Host machine** (where you run `topo`):
 
 - [Docker](#install-a-container-engine)
+- [Git](https://git-scm.com/downloads)
 - OpenSSH Client
 
 **Target machine** (the remote Arm system):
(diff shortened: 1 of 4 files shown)


th3james:
I wonder if this is getting a bit excessive and we're creating a big scary checklist with relatively poor signal-to-noise ratio. If we are going to do this is should be `SeverityWarning` because it's only necessary for `topo clone` with git, we still support `dir:`

federicobozzini:
Very reasonable https://github.com/arm/topo/pull/439/commits/2088ba76140bb2a70f7712c217dadb1e489ecb53

muchzill4:
I'd rather we didn't do this (just yet).

`topo health` right now is tightly integrated into deployment flow. Heck, even `topo deploy` tells you to run `topo health` if the deployment fails. Reporting `git` status might be confusing unless we somehow group the software health checks into respective flow groups (clone, deploy). I believe we should think more carefully before bolting on another check.

> big scary checklist with relatively poor signal-to-noise ratio

Sums it up pretty nicely.

My preference would be to scope this change to `README.md` for now.

PR #442, opened 2026-09-04
refactor: rename args -> parameters, resolve -> configure

awphi:
## Changes

Attempts to have a stronger opinion about naming in the paramaterisation bit of the codebase as it's quite muddy atm (e.g. many uses of the term 'resolved', lots of references to 'args'/'arguments' which we don't use anymore):

 - Renamed the internal package from `arguments` to `parameter`. The singular gives its exports more natural names i.e. "parameter provider", "parameter values"
 - Replaced argument terminology with `Definition`, `Values`, and parameter-focused errors. This ditches one use of the term 'resolved'. `Values` is a simplification that avoids the conversion from map->array->map we had before.
 - Kept “args” only for CLI arguments and Docker build.args.
 - Renamed the overall workflow to project.Configure, ditching one usage of the word 'resolve'/'resolved'
 - Replaced ambiguous “resolved” checks in the strict provider chain with `hasValue` et al. Ditches the final use of 'resolved'

Big PR I know but it's almost pure renaming and split up by commit fairly well imo. Please push back if you start reviewing and disagree!

## Checklist


internal/arguments/cli_provider.go
```diff
@@ -1,44 +0,0 @@
...
-		parts := strings.SplitN(arg, "=", 2)
-		if len(parts) != 2 {
-			return nil, fmt.Errorf("invalid argument format: %s (expected ARG=VALUE)", arg)
-		}
-		parsed[parts[0]] = parts[1]
-	}
-	return &CLIProvider{input: parsed}, nil
-}
-
-func (p *CLIProvider) Provide(args []Arg) ([]ResolvedArg, error) {
```
muchzill4, line 26: I'm not sure if we should rename the package. We should definitely update the signatures though.

Providers are responsible for turning params into args:

```
func (p *CLIProvider) Provide(params []Param) ([]Arg, error) {
```

Frankly, I'd rather the package was still `arguments`, not `parameters`. I.e.: named after what it provides, not what it takes as an argument.

internal/compose/yaml_test.go
```diff
@@ -202,9 +202,9 @@ services:
       args:
         PLATFORM: old-value
 `)
-		args := map[string]string{"PLATFORM": "stm32mp257"}
+		parameters := map[string]string{"PLATFORM": "stm32mp257"}
```
muchzill4: They are args at this point, no?

internal/arguments/cli_provider.go
awphi: > Providers are responsible for turning params into args

It's a bit confusing because they don't turn params into args, just into a KV map (now represented by `parameter.Values`. I also think 'arg' is too loose a term and carries several meanings - docker CLI args, topo CLI args, docker build args.

I think the current package name becomes even more confusing as we look towards moving to .env-based configuration at which point we don't even touch build args, they're just values to be plonked in a file that may or may not be consumed in build args, deployment_success_messages, or anywhere.

Happy to walk this bit back if you still disagree though!

internal/parameter/resolver.go
```diff
@@ -1,25 +1,22 @@
...
 	"fmt"
 )
 
-type Arg struct {
+type Definition struct {
 	Name          string
 	Description   string
 	Required      bool
 	Example       string
 	CurrentValues []string
```
muchzill4, line 13: This rename makes it obvious that `CurrentValues` shouldn't be a property of `Arg|Definition`. 
How can a `Definition` have `CurrentValues`? It's a definition after all.

internal/compose/yaml_test.go
```diff
@@ -64,9 +64,9 @@ services:
       args:
         FOO: elephant
 `)
-		args := map[string]string{"FOO": "baz"}
+		parameters := map[string]string{"FOO": "baz"}
```
muchzill4: So these are values at this point? Parameters would be value-less?

internal/arguments/cli_provider.go
muchzill4: I see now. Thanks for explaining. I really like `parameter.Values`, however, `parameter.Provider` sounds like something that provides parameters, not values. Perhaps we should tighten the naming there as well.

internal/parameter/resolver.go
awphi: Yep felt that way too. This is something I'll be addressing as part of the .env file work since it renders `.CurrentValues` obsolete if that's okay?

internal/compose/yaml_test.go
awphi: Yep - the signature is wrong too. Good spot! I've changed the `ApplyParameters` signatures to `ApplyParameterValues` too. https://github.com/arm/topo/pull/442/commits/1decdf4e2b6418aa5c5486be56fe1cd86f914bb8

The whole concept of 'applying parameter values' is going the way of the dodo soon though to be replaced by 'setting parameter values (in an env file)' which is clearer imo!

internal/compose/yaml_test.go
awphi: They should be `parameter.Values`: https://github.com/arm/topo/pull/442/commits/1decdf4e2b6418aa5c5486be56fe1cd86f914bb8

internal/arguments/cli_provider.go
awphi: Hmm fair. What about `parameter.ValueProvider` https://github.com/arm/topo/pull/442/commits/e2eff707a5dbd589d7970aac79f2d6c862034112 - reads as 'provider of parameter values` to me. Little wordy but more accurate

yejseo01 (approved):
Looks harmless to me. it's good we are alleviating some meanings from args.

internal/arguments/cli_provider.go
muchzill4: I like that more. Though this also means we should have `ValueProvider` everywhere else. It's path of least confusion when you have `CLIValueProvider` (maybe `ValueProviderFromCLI`?). 

Alternatives that are perhaps meaningful w/o `Value` prefix: `Resolver`, `Filler`, `Binder`.

internal/compose/yaml_test.go
muchzill4: Sorry, I meant `parameters` variable name is probably bad, because they are resolved parameters a.k.a. values (now that I understand this is how you call resolved params).

I think `compose` namespace should be unaware of `parameters.Values`, former feels like a lower level module and the dependency direction should be: `parameters.Values -> ... -> compose`, not `parameters.Values -> ... compose -> parameters.Values`.

internal/compose/yaml_test.go
muchzill4: I like signature rename, but https://github.com/arm/topo/pull/442#discussion_r3950147635

internal/arguments/cli_provider.go
awphi: I like `Resolver` - feels quite neat: "resolver resolves definitions into values". Fairly chonky change to do so here:  https://github.com/arm/topo/pull/442/commits/427ed10fc87f25144ba0dd541b5859ff7d82cd87

internal/compose/yaml_test.go
awphi: Have reverted the `compose` boundary to just take a [simple map](https://github.com/arm/topo/pull/442/commits/69b7772f136994b129407e2f847c7f289f73e84f).

muchzill4 (approved):
👏

PR #445, opened 2026-09-07
feat: summarize health checks and add verbose output

federicobozzini:
## Changes

- `topo health` only shows errors and warnings
- a successful summary is shown when `topo health` doesn't report any issue
- a `verbose` flag is introduced to report the result of all checks

## Screenshot
```
$ topo health  --target pi5-rod
┌─ Host ────────────────────────────────────────────────
 ✓ All checks passed

┌─ Target: ssh://pi5-rod ───────────────────────────────
 ✓ All checks passed
 i Processing Domain Driver (remoteproc) (no remoteproc devices found)
```

## Checklist


internal/output/views/health_test.go
```diff
@@ -76,7 +90,7 @@ func TestHealthReport(t *testing.T) {
 			assert.Contains(t, out.String(), " ! Processing Domain Driver (remoteproc) (no remoteproc devices found)")
 		})
 
-		t.Run("it renders an info icon for info checks", func(t *testing.T) {
+		t.Run("it renders remoteproc info without suppressing the success summary", func(t *testing.T) {
```
muchzill4: This test now conflates two things: info icon and suppression of the success message. We should probably have a test evaluating all icons.

internal/output/views/health_test.go
```diff
@@ -13,16 +13,14 @@ import (
 
 func TestHealthReport(t *testing.T) {
 	t.Run("PlainFormat", func(t *testing.T) {
-		t.Run("it renders the healthy host dependencies", func(t *testing.T) {
+		t.Run("it renders a generic result for healthy host and target sections", func(t *testing.T) {
```
muchzill4: What's "a generic result"? It's not obvious to me what this is actually testing now.

We should make three cases obvious:
1. renders summaries if all checks pass in verbose=true mode
2. renders details of failing checks details in verbose=false mode
3. renders "all good" summary and no details in verbose=false mode

internal/output/views/health.go
```diff
@@ -14,6 +14,23 @@ type HealthReport struct {
 	Host       health.HostReport    `json:"host"`
 	Target     *health.TargetReport `json:"target,omitempty"`
 	TargetHint string               `json:"-"`
+	Verbose    bool                 `json:"-"`
```
muchzill4: Doesn't feel like `Verbose` belongs here. These are contents of a health report, verbose adds "how to print it", which doesn't belong here.

internal/output/views/health.go
```diff
@@ -14,6 +14,23 @@ type HealthReport struct {
...
+type healthTargetSection struct {
+	Destination string
+	Section     healthCheckSection
+}
+
+type plainHealthReport struct {
+	Host       healthCheckSection
+	Target     *healthTargetSection
+	TargetHint string
 }
```
muchzill4, line 21: Why do we need all these new types? What purpose do they serve? I find it a bit confusing. It could be the naming (what is plain?) or the purpose (why new types if all we do is display "All good" if all checks pass).

internal/output/views/health.go
```diff
@@ -39,39 +45,56 @@ const healthReportTemplate = `
...
-  {{- end }}
+{{ if .Target }}{{ targetHeading .Target.Destination }}{{ template "checkSection" (section .Target.Checks) }}
 {{- else -}}
 {{ sectionHeading "Target" }}
 {{ .TargetHint }}
 {{- end }}
 
 `
 
+func PrintHealthReport(report HealthReport, w io.Writer, format term.Format, verbose bool) error {
```
federicobozzini: I'm not a fan of this function as it replicates some of the `Print` functionality. My favourite solution would be to add a `verbose` option to `Print` and reuse it here. As a downside, when a verbose option is not available it would mean passing an extra `false` just of the sake of it.

If we agree on this, I would prefer to do it in a separate PR, that we can merge either before or after this one.

internal/output/views/health.go
federicobozzini: Agreed, removed.

internal/output/views/health.go
federicobozzini: Removed, only `healthCheckSection` remains which I think makes things clearer.

internal/output/views/health_test.go
federicobozzini: Done, all three cases are now tested separately.

internal/output/views/health_test.go
federicobozzini: Done by `it colors status labels when writing to a terminal`.

internal/output/views/health.go
federicobozzini: Removed doing this https://github.com/arm/topo/pull/445/changes/2f70758c6073387d8dc14cd5fd606c4c398b644c

internal/output/views/health.go
```diff
@@ -8,13 +8,14 @@ import (
 	"github.com/arm/topo/internal/output/term"
 )
 
-type HealthReport struct {
-	Host   health.HostReport
-	Target health.TargetReport
+type HealthReportView struct {
+	health.HealthReport
+	Verbose bool
 }
```
federicobozzini, line 16: I prefer this shape here, but I'm happy to revert it to something more similar to what we have on `main` to reduce the churn.

awphi (commented):
Hot take: I don't think we should do this. An unsummarised check provides useful information about what Topo does and how it does it. 

Obviously we've not taken that away but hiding it behind a flag says "you don't have to care about this" which I'm not sure is what we want. Abstracting the container engine in use, the protocol or anything else makes graduating off Topo or piloting it via `ssh`+`docker` commands a bit less obvious imo.

I think this had more utility when the health check was noisy/ugly but it's quite easy on the eyes now imo. Happy to be disagreed with/vetoed though!

PR #448, opened 2026-09-07
refactor: simplify health dependency checks

muchzill4:
## Changes

- Replace composable check types with dependency check functions and **explicit results**.
  - Single return value from each check (`(*Fix, error)` was confusing)
  - Remove abstract `MatchesVersionCheck` w/ topo version specific check
  - Drop `Binary` struct field, use **explicit result** to carry the passing check value

## Checklist

### DIFF
--- a/internal/health/dependencies_test.go
+++ b/internal/health/dependencies_test.go
@@ -6,7 +6,6 @@ import (
 	"slices"
 	"testing"
 
-	"github.com/arm/topo/internal/command"
 	"github.com/arm/topo/internal/health"
 	"github.com/arm/topo/internal/runner"
 	"github.com/arm/topo/internal/ssh"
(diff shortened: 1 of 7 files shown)


internal/health/dependencies_test.go
```diff
@@ -190,41 +179,41 @@ func TestFilterByHardware(t *testing.T) {
...
 
-type failingCheck struct{}
-
-func (p failingCheck) Run(_ context.Context, _ runner.Runner, _ health.Dependency) (*health.Fix, error) {
-	return &health.Fix{Description: "fix me please", Command: "echo fixed"}, errors.New("very broken")
+func failingCheck(_ context.Context, _ runner.Runner) health.DependencyCheckResult {
+	return health.DependencyCheckResult{Failure: &health.DependencyCheckFailure{
+		Severity: health.SeverityError,
+		Message:  "very broken",
+		Fix:      &health.Fix{Description: "fix me please", Command: "rm -rf /"},
```
awphi: It's back!!

awphi (approved):
Much, much, much better!

internal/health/dependencies_test.go
muchzill4: Oops.

muchzill4: 8e7d654

PR #449, opened 2026-09-07
refactor: decouple health internals from presentation layer

muchzill4:
- Decouple health-check domain models from output view and JSON serialization, so health logic can evolve independently of CLI presentation and output formats.

## Checklist

### DIFF
--- a/internal/output/views/health.go
+++ b/internal/output/views/health.go
@@ -17,11 +17,11 @@ type HealthReport struct {
 
 func NewHealthReport(host health.HostReport, target *health.TargetReport, targetHint string) HealthReport {
 	report := HealthReport{
-		Host:       hostReportFromHealth(host),
+		Host:       toViewHostReport(host),
 		TargetHint: targetHint,
 	}
 	if target != nil {
-		viewTarget := targetReportFromHealth(*target)
+		viewTarget := toViewTargetReport(*target)
 		report.Target = &viewTarget
 	}
 	return report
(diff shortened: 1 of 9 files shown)


internal/output/views/health.go
```diff
@@ -77,6 +89,14 @@ func (r HealthReport) AsPlain(isTTY bool) (string, error) {
...
 }
 
+func (r HealthReport) AsJSON() (string, error) {
+	b, err := json.MarshalIndent(r, "", "  ")
+	if err != nil {
+		return "", fmt.Errorf("encode report as json: %w", err)
+	}
+	return string(b), nil
+}
+
```
yejseo01, line 94: it seems like this is the same `AsJSON()` function logic as `AsJSON()` of container_list.go. Should we make this the default `AsJSON()` logic for `View` interfaces?

internal/output/views/health.go
```diff
@@ -103,10 +123,56 @@ func healthStatusFormatter(isTTY bool) func(health.CheckStatus) string {
...
+	return viewChecks
+}
+
+func healthCheckFromHealth(check health.HealthCheck) healthCheck {
+	viewCheck := healthCheck{Name: check.Name, Status: check.Status, Value: check.Value}
+	if check.Fix != nil {
+		viewCheck.Fix = &fix{Description: check.Fix.Description, Command: check.Fix.Command}
+	}
+	return viewCheck
 }
```
yejseo01: nit: I feel like the names of these functions are too similar

Luke-Parkin: +1 although maybe function signature is enough

Luke-Parkin (approved):
Lgtm! 

nit: Although I don't think you recharge the flux in a capacitor... I think that's just the medium to store energy...

internal/output/views/health.go
muchzill4: 7d61758

internal/output/views/health.go
muchzill4: d7c41e8

muchzill4:
> nit: Although I don't think you recharge the flux in a capacitor... I think that's just the medium to store energy...

68a17c1

PR #441, opened 2026-09-04
chore: Bump docs action version

Luke-Parkin:
## Changes

- Bump the docs action version

## Checklist

- [x] 🤖 This change is covered by tests as required.
- [x] 🤹 All required manual testing has been performed.
- [x] 📖 All documentation updates are complete.

PR #450, opened 2026-09-08
feat: update docs publish version & add rerelease flow

phawxby:
## Changes

* Extends on #441 to update release docs action version
* Adds workflow to allow rerelease of docs from main action based on previously released tag

## Checklist


.github/workflows/publish-released-docs.yaml
```diff
@@ -0,0 +1,60 @@
+name: Publish Released Docs
```
muchzill4: 1. Why is this workflow useful?
2. It looks almost exactly like the publish flow from release.yaml, can we dedupe this by having a reusable action/workflow?

phawxby: 1. This would allow you to re-run the docs publish for the current release tag but from main, so you can update the underlying docs action in main and not have to do a full topo release to update the production docs against the latest docs engine. 
2. Yes, that makes a lot of sense.

.github/workflows/republish-release-docs.yaml
```diff
@@ -0,0 +1,44 @@
...
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          if [[ ! "$RELEASE_TAG" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+            echo "Expected a stable SemVer tag, such as v12.0.0" >&2
+            exit 1
+          fi
+
+          gh api "repos/${GITHUB_REPOSITORY}/git/ref/tags/$RELEASE_TAG" > /dev/null
+          gh release view "$RELEASE_TAG" --json isDraft,isPrerelease \
```
th3james: Because you've not done a checkout this won't know which repo you're referring to
```suggestion
          gh release view "$RELEASE_TAG" --repo "$GITHUB_REPOSITORY" --json isDraft,isPrerelease \
```

.github/workflows/publish-release-docs.yaml
```diff
@@ -0,0 +1,40 @@
...
+  publish:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write
+    steps:
+      - name: Checkout docs source
+        uses: actions/checkout@v7
+        with:
+          ref: ${{ inputs.source-ref }}
```
th3james, line 24: Worth noting that this means you'll end up running `./docs/scripts/generate-docs-config.sh` as it exists in this version against `ARM-software/docs-action/publish@v0.4.0`, which may cause incompatibilities. I don't think this is a blocker, but might trip us up later

th3james (changes requested):
Republish flow appears broken, other comment is advisory

.github/workflows/publish-release-docs.yaml
phawxby: I'm hoping this is rarely used and mostly an emergency exit, but yes. I suppose the other option is embed the script in the workflow.

.github/workflows/republish-release-docs.yaml
phawxby: 🤦 I tested this locally but didn't account for actually having the repo checked out. Thanks

.github/workflows/republish-release-docs.yaml
awphi, line 1: It's unclear to me in what situation this republish flow would be useful. 

I guess it's for when we need to update the docs publish action in a pinch but given we fairly religiously try to keep an always-releasable `main`, we'd probably be okay to just do a CLI release to trigger that re-publish. Worst case scenario is it happens immediately after a CLI release and we end up release `X.Y.Z+1` for no real reason but that's not a huge deal imo.

Happy to vetoed by others if they don't think there's harm in it but less GH action complexity always good imo

th3james:
@phawxby Discussed this with the team and we concluded that [Adam's comment](https://github.com/arm/topo/pull/450#discussion_r3969263706) was correct - we're actually going to stick to only allowing releases of the docs at the same time as releases of the CLI. Given that we assume the default branch is always releasable, this shouldn't block releases of the docs in a timely fashion, with the primary risk being bumps to semver versions. This isn't a perfect solution, and we may revisit it, but we believe it's the right approach for now and reduces the amount of CI code we have to maintain.

PR #451, opened 2026-09-12
refactor: model remoteproc as a health dependency

muchzill4:
## Changes

Model the remoteproc driver check as a target dependency instead of a separate out of band hardware check. 
Once "connectivity" is also migrated in a subsequent PR, we'll have a simple collection of dependencies that can be set as prerequisites to each other and grouped independently into flows (deployment, general host/target health).

## Checklist

### DIFF
--- a/internal/health/dependencies.go
+++ b/internal/health/dependencies.go
@@ -3,7 +3,9 @@ package health
 import (
 	"context"
 	"fmt"
+	"strings"
 
+	"github.com/arm/topo/internal/probe"
 	"github.com/arm/topo/internal/runner"
 	"github.com/arm/topo/internal/ssh"
 )
@@ -23,7 +25,6 @@ type Dependency struct {
 	Label                 string
 	Check                 DependencyCheckFn
 	SoftwarePrerequisites []DependencyID
-	HardwarePrerequisites []HardwareCapability
 }
 
 type DependencyCheckFn func(ctx context.Context, r runner.Runner) DependencyCheckResult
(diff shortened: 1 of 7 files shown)


internal/health/dependencies.go
```diff
@@ -18,12 +20,13 @@ const containerEngineInstallURL = "https://github.com/arm/topo#install-a-contain
...
 type Dependency struct {
-	ID                    DependencyID
-	Label                 string
-	Check                 DependencyCheckFn
-	SoftwarePrerequisites []DependencyID
-	HardwarePrerequisites []HardwareCapability
+	ID            DependencyID
+	Label         string
+	Check         DependencyCheckFn
+	Prerequisites []DependencyID
```
yejseo01, line 29: nice

internal/output/views/health.go
```diff
@@ -147,13 +147,22 @@ func toViewHostReport(report health.HostReport) hostReport {
...
 func toViewTargetReport(report health.TargetReport) targetReport {
-	return targetReport{
+	target := targetReport{
 		Destination:            report.Destination,
 		IsLocalhost:            report.IsLocalhost,
 		Connectivity:           toViewHealthCheck(report.Connectivity),
-		Dependencies:           toViewHealthCheckList(report.Dependencies),
-		ProcessingDomainDriver: toViewHealthCheck(report.ProcessingDomainDriver),
+		Dependencies:           make([]healthCheck, 0, len(report.Dependencies)),
+		ProcessingDomainDriver: healthCheck{Name: "Processing Domain Driver (remoteproc)"},
```
yejseo01, line 155: do you think it still makes sense to report it different to its actual shape like this?

muchzill4: This is to avoid making a breaking change to JSON. 

Once everything is a dependency, I plan to change the shape of the output via a breaking change. 👍

yejseo01 (approved):
yay!

PR #453, opened 2026-09-14
refactor: hoist and unify project scope at the CLI layer

awphi:
> Note: Sorry for the big PR - no easy split for this one and a big-bang change seemed like less work to review that several smaller ones with transitional shims.

## Changes

- Introduces the concept of `project.Scope` which represents everything needed to properly load a Topo project. Atm this is just a compose file path and a set of arbitrary env vars which we prefill with `TOPO_TARGET` and `TOPO_TARGET_HOSTNAME` for use in things like `deployment_success_message`. **Why?**
  - Centralizes all the potentially expensive, or potentially error-creating reads/loads in the CLI layer then passes the result down to our functional operations _without_ global state (like we do right now for `TOPO_TARGET`/`TOPO_TARGET_HOSTNAME`)
  - Provides an easy extension point for command-wide compose inputs like `--env-file`. This lays the ground work for being able to run `topo deploy --target foo --env-file .custom.env` without us having to pass a `envFiles []string` parameter all over the place
- Moves `compose.ReadProject`, `compose.PullableServices` and `compose.ImageNames` to `project` and renames the former to `project.Read` as `compose` feels like it should be a much lower-level package and not deal with the concept of Topo Project scopes.
- Attempts to standardize var naming in this area; `scope` for `project.Scope`s and `composeProject` for the result of `project.Read(scope)` to be more specific than just "project" for both.

## Checklist


muchzill4 (approved):
LGTM.

PR #456, opened 2026-09-14
refactor: simplify plain health report templates

muchzill4:
## Changes

- Render plain health reports directly from health domain reports.
- Simplify target output to a single section and include all target dependencies.
- Preserve JSON output through dedicated JSON view mappings.
- Include the target destination as the successful connectivity value.

## Checklist


awphi (approved):
👍

PR #457, opened 2026-09-14
ci: re-enable package-level parallelism for test execution

awphi:
## Changes

Drops `-p 1` from our test execution to make it go faaaaast - rough conservative results are ~2x faster tests for all OSes. How? By addressing three contention/race issues:

- Container names for test "targets" clash between podman and docker tests with the same test names. Resolved by adding a bit of random text to test target container names.
- SSH connections to the "targets" can fail because we only wait for a listening port but this doesn't necessarily mean sshd is ready to receive connections. This caused flakiness rather than consistent failures. My theory is the increased load on the machine + potentially many concurrent targets delayed ssh start-up. Resolved by waiting for ssh banner.
- `known_hosts` contention when adding host keys for the test containers. Not an easy fix because `ssh-keygen` needs to edit the file for us and windows `ssh`/`docker` processes may take a read lock on the file outside the bounds of our lock. Resolved by:
  - Ensuring a small snippet of ssh config is added to the developer's ssh config which aliases `topo-test-*` to `localhost` and gives it a unique `known_hosts` file for that target. This prevents contention of a single `known_hosts` altogether + prevents littering your main `known_hosts` files with `localhost:PORT` entries.
  - Swapping all the targets to point to `ssh://topo-test-$CONTAINER_NAME:$PORT` to use the new ssh config snippet.
  - Note that to edit the ssh config we have to acquire a lock on it to prevent concurrent e2e tests clashing when attempting to add this ssh snippet.

Sample runs:
1. [Run 1](https://github.com/arm/topo/actions/runs/35077627409)
2. [Run 2](https://github.com/arm/topo/actions/runs/35077630093)
3. [Run 3](https://github.com/arm/topo/actions/runs/35077633271)
4. [Run 4](https://github.com/arm/topo/actions/runs/35077636275)
5. [Run 5](https://github.com/arm/topo/actions/runs/35077639402)

Other improvements we could consider are:
- Sharding across multiple runners
- Bigger hardware (needs org admin to provision one or use an external GHA runner)
- Pre-build images for Windows with WSL/docker/podman
- More package splitting or intra-package parallelism

## Checklist


muchzill4:
Very cool!

> Container names for test "targets" clash between podman and docker tests with the same test names. Resolved by adding a bit of random text to test target container names.

Why is this a problem? Different container engines? 🤔 

---

One thing I'm a bit uneasy about is the ssh config write. I think it should be opt in, like we do with golden files. 🤔

awphi:
> Why is this a problem? Different container engines? 🤔

The outer container engine on the host OS which we use to launch the "targets" is always docker afaict(?), which is where the clash happens.

> One thing I'm a bit uneasy about is the ssh config write. I think it should be opt in, like we do with golden files. 🤔

I'm similarly not in love with this but couldn't find an alternative I liked i.e.:
- Opt-in with an env var + skip the tests if no config set - hard to tell you've skipped a bunch of tests
- Opt-in with an env var + error out if not set - bit better but means you can't just "run the tests"
- Opt-in with an env var, fall back to localhost with `known_host` contention - can exhibit flakiness locally and need to maintain the old `known_hosts` addition+removal path

WDYT?

muchzill4:
> The outer container engine on the host OS which we use to launch the "targets" is always docker afaict(?), which is where the clash happens.

🤦‍♂️ 

> I'm similarly not in love with this but couldn't find an alternative I liked i.e.:
> 
> * Opt-in with an env var + skip the tests if no config set - hard to tell you've skipped a bunch of tests
> * Opt-in with an env var + error out if not set - bit better but means you can't just "run the tests"
> * Opt-in with an env var, fall back to localhost with `known_host` contention - can exhibit flakiness locally and need to maintain the old `known_hosts` addition+removal path

I'm leaning towards 1. `t.Skip("<reason>")`. It's not that dissimilar to us skipping tests because no docker or no podman is installed.

You're going to hate me: can we detect if `-p 1 || has_ssh_config`? 🤔

awphi:
> I'm leaning towards 1. `t.Skip("<reason>")`. It's not that dissimilar to us skipping tests because no docker or no podman is installed.
> 
> You're going to hate me: can we detect if `-p 1 || has_ssh_config`? 🤔

Hmm good point. Detection of `-p 1` doesn't seem feasible afaict. Could do a wrapper script but that is very icky. Went with just skipping if `!has_ssh_config` and moved the setup to a separate script so we can removing flocking entirely.

internal/testutil/ssh_config.go
```diff
@@ -0,0 +1,36 @@
...
+	knownHostsFile := filepath.ToSlash(filepath.Join(knownHostsDir, containerName))
+	if !strings.Contains(config, "hostname localhost\n") || !strings.Contains(config, "userknownhostsfile "+knownHostsFile+"\n") {
+		t.Skip("container tests require SSH setup: go run ./internal/testutil/cmd/setup-test-ssh")
+	}
+	info, err := os.Stat(knownHostsDir)
+	if os.IsNotExist(err) {
+		t.Skip("container tests require SSH setup: go run ./internal/testutil/cmd/setup-test-ssh")
+	}
+	require.NoError(t, err)
+	require.True(t, info.IsDir(), "test known-hosts path must be a directory")
```
muchzill4: Should we interpolate the `known-hosts path` here? I.e. `fmt.Sprintf("%s must be a directory", knownHostsDir)`?

internal/testutil/container.go
```diff
@@ -193,32 +183,53 @@ func acceptHostKey(c *Container) error {
 	return nil
 }
 
-func removeHostKey(c *Container) {
-	u, err := url.Parse(c.SSHDestination)
+func removeHostKey(t *testing.T, containerName string) {
```
muchzill4: Feels like `removeHostKey` belongs in `ssh_config.go`. Also, perhaps `removeKnownHost`?

internal/testutil/ssh_config.go
```diff
@@ -0,0 +1,36 @@
...
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func requireSSHConfig(t testing.TB, containerName string) {
```
muchzill4: `requireKnownHostsSetup`?

muchzill4 (approved):
Don't love the complexity this requires, but CI speedup trumps it. Nice work figuring this out.

internal/testutil/ssh_config.go
awphi: Went with a slightly different name to still capture it requires some ssh config https://github.com/arm/topo/pull/457/commits/653ed0207d3d89020243478805f302920998fc52

internal/testutil/ssh_config.go
awphi: https://github.com/arm/topo/pull/457/commits/326be3effc1c380e46e1a9945c7e90576690fc20

internal/testutil/container.go
awphi: https://github.com/arm/topo/pull/457/commits/27ebb1af69da28c952243acfcd9e40e8b969e1e6

muchzill4 (approved):

PR #461, opened 2026-09-14
fix: consistent cli output

yejseo01:
## Changes

  - Centralizes section headers, spacing, and final newlines.
  - Removes the initial blank line and hides headers for quiet sections.
  - Applies formatting to deploy, stop, clone, and setup-keys.
  - Appends a newline to reports and JSON, preserving existing newlines.
  - Preserves interactive Docker progress in the main command runner.
  - Adds output-formatting tests and protects concurrent writes.

## Checklist


internal/output/views/project_list_test.go
```diff
@@ -157,7 +157,7 @@ name-of-other-project
   when the content is long enough to span multiple lines.
 
 `
-		assert.Equal(t, want, outBuf.String())
+		assert.Equal(t, want+"\n", outBuf.String())
```
muchzill4, line 160: We should not be adding `\n` to every test. It's noisy and provides little value.

internal/operation/sequence.go
```diff
@@ -13,16 +13,18 @@ func NewSequence(operations ...Operation) Sequence {
 }
 
 func (s Sequence) Run(cmdOutput io.Writer) error {
```
muchzill4, line 15: We should aim to deprecate operations.

internal/deploy/docker/command.go
```diff
@@ -32,6 +33,13 @@ func RunComposeCommand(ctx context.Context, output io.Writer, host Host, scope p
 }
 
 func run(cmd *exec.Cmd, output io.Writer) error {
+	if section, ok := output.(*term.SectionPrinter); ok {
+		var err error
+		output, err = section.SubprocessOutput()
+		if err != nil {
+			return err
+		}
+	}
```
muchzill4, line 42: Can you explain what this does? Alarm bells ringing because I see a try-cast and a call (`SubprocessOutput`) the `run` function should not ever care about.

`run` should be given a writer and just write. It should not care about presentation.

internal/output/term/term_test.go
```diff
@@ -78,29 +79,83 @@ func TestWrapText(t *testing.T) {
 	})
 }
 
-func TestPrintHeader(t *testing.T) {
-	t.Run("renders header with padding", func(t *testing.T) {
+func TestCommandOutputFinish(t *testing.T) {
+	for _, message := range []string{"message", "message\n", "message\n\n", "message\n\n\n"} {
```
muchzill4, line 83: Why do we need to assert that it appends a newline after all those trailing `\n`? Is there anything specific about `\n\n\n` or `\n\n`?

e2e/setup_keys_test.go
```diff
@@ -39,7 +39,7 @@ func TestSetupKeysJourney(t *testing.T) {
 	setupOut, err := cmd.CombinedOutput()
 	require.NoError(t, err, "topo failed: %s", setupOut)
 	assert.Contains(t, string(setupOut), "Generate SSH key pair")
-	assert.Contains(t, string(setupOut), "Transfer public key")
+	assert.NotContains(t, string(setupOut), "Transfer public key")
```
muchzill4, line 42: Why this has changed?

internal/output/term/header.go
```diff
@@ -4,15 +4,70 @@ import (
...
-func PrintHeader(w io.Writer, description string) error {
-	header := Header(description, IsTTY(w))
-	if header == "" {
-		return nil
+type SectionPrinter struct {
+	*CommandOutput
+	pending string
+}
+
+type CommandOutput struct {
```
muchzill4, line 15: I think most of this new code doesn't belong in `header.go`.

muchzill4 (commented):
Must admit I'm not a fan of swapping `io.Writer` for `*term.CommandOutput` everywhere. `io.Writer` is more conventional - there's no need to remember to use awkward constructs such as `Subprocess` or `Finish`. I'd much rather if we kept using `io.Writer`.

We can either have `PrintFirstHeader` and `PrintNthHeader`. It's simple, but likely hard to track which header we're printing across collaborating functions.

Alternatively, we could have a custom `io.Writer`, such that "swallows" leading newlines.

yejseo01:
> Must admit I'm not a fan of swapping `io.Writer` for `*term.CommandOutput` everywhere. `io.Writer` is more conventional - there's no need to remember to use awkward constructs such as `Subprocess` or `Finish`. I'd much rather if we kept using `io.Writer`.
> 
> We can either have `PrintFirstHeader` and `PrintNthHeader`. It's simple, but likely hard to track which header we're printing across collaborating functions.

With this approach, we still have problems where if there are no logs from that header, the header will still print.

> 
> Alternatively, we could have a custom `io.Writer`, such that "swallows" leading newlines.

I'll proceed in this direction

e2e/setup_keys_test.go
yejseo01: because the logs of "Transfer public key" is empty.

internal/output/term/term_test.go
yejseo01: so, I established that all logs end with a newline appended regardless of the logs. that means if the logs include several newlines, we would respect that and append a newline.

I'm also open to ending with 1 newline at all times 
so it would be: 
```
"messsage" -> "message\n"
"messsage\n" -> "message\n"
"messsage\n\n" -> "message\n"
```

internal/deploy/docker/command.go
yejseo01: so we can do without this part of the code. it''s just the docker doesn't print in the same format without it. Without this part of the code, it would print like this: 
```
┌─ Build images ────────────────────────────────────────
 Image topo-welcome-app Building 
#1 [internal] load local bake definitions
#1 reading from stdin 664B done
#1 DONE 0.0s

#2 [internal] load build definition from Dockerfile
#2 transferring dockerfile: 223B done
#2 DONE 0.0s

#3 [internal] load metadata for docker.io/library/nginx:alpine
#3 DONE 0.6s

#4 [internal] load .dockerignore
#4 transferring context: 2B done
#4 DONE 0.0s

#5 [1/3] FROM docker.io/library/nginx:alpine@sha256:72ba65eb42c10344912a84ff42408db7d34f2feb642204570ab8fc5ffd29f1d3
#5 resolve docker.io/library/nginx:alpine@sha256:72ba65eb42c10344912a84ff42408db7d34f2feb642204570ab8fc5ffd29f1d3 0.0s done
#5 DONE 0.0s

#6 [internal] load build context
#6 transferring context: 59B done
#6 DONE 0.0s

#7 [2/3] COPY src/index.html /usr/share/nginx/html/index.html
#7 CACHED

#8 [3/3] RUN sed -i "s|{{GREETING_NAME}}|World|" /usr/share/nginx/html/index.html
#8 CACHED

#9 exporting to image
#9 exporting layers done
#9 exporting manifest sha256:20af9712d7d284873b91691546e8db3e85fbbddd3f1d79340789f64a2291812c done
#9 exporting config sha256:d21e589c04934308d6777b1fb0d4f9c003555df7d6d52715b0a89180524e9e3e done
#9 exporting attestation manifest sha256:8b3256452492671cabf8b0f3151c204d3d183315229d7593a358d9039fb50458 0.0s done
#9 exporting manifest list sha256:29729d9ec7d0f2061f91de7e524a6a9d32dd6efc3817aac91df7d99c86c6c6b7 done
#9 naming to docker.io/library/topo-welcome-app:latest done
#9 unpacking to docker.io/library/topo-welcome-app:latest
#9 unpacking to docker.io/library/topo-welcome-app:latest done
#9 DONE 0.1s

#10 resolving provenance for metadata file
#10 DONE 0.0s
 Image topo-welcome-app Built 
```

It used to print like this:
```
yejseo01@CF7PJYX5G9 topo-welcome % ../../topo/topo deploy --target pi5-grimes
┌─ Build images ────────────────────────────────────────
[+] Building 1.8s (10/10) FINISHED                                                                                                                                                                                                                                                    
 => [internal] load local bake definitions                                                                                                                                                                                                                                       0.0s
 => => reading from stdin 664B                                                                                                                                                                                                                                                   0.0s
 => [internal] load build definition from Dockerfile                                                                                                                                                                                                                             0.0s
 => => transferring dockerfile: 223B                                                                                                                                                                                                                                             0.0s
 => [internal] load metadata for docker.io/library/nginx:alpine                                                                                                                                                                                                                  1.5s
 => [internal] load .dockerignore                                                                                                                                                                                                                                                0.0s
 => => transferring context: 2B                                                                                                                                                                                                                                                  0.0s
 => [internal] load build context                                                                                                                                                                                                                                                0.0s
 => => transferring context: 59B                                                                                                                                                                                                                                                 0.0s
 => [1/3] FROM docker.io/library/nginx:alpine@sha256:72ba65eb42c10344912a84ff42408db7d34f2feb642204570ab8fc5ffd29f1d3                                                                                                                                                            0.0s
 => => resolve docker.io/library/nginx:alpine@sha256:72ba65eb42c10344912a84ff42408db7d34f2feb642204570ab8fc5ffd29f1d3                                                                                                                                                            0.0s
 => CACHED [2/3] COPY src/index.html /usr/share/nginx/html/index.html                                                                                                                                                                                                            0.0s
 => CACHED [3/3] RUN sed -i "s|{{GREETING_NAME}}|World|" /usr/share/nginx/html/index.html                                                                                                                                                                                        0.0s
 => exporting to image                                                                                                                                                                                                                                                           0.0s
 => => exporting layers                                                                                                                                                                                                                                                          0.0s
 => => exporting manifest sha256:20af9712d7d284873b91691546e8db3e85fbbddd3f1d79340789f64a2291812c                                                                                                                                                                                0.0s
 => => exporting config sha256:d21e589c04934308d6777b1fb0d4f9c003555df7d6d52715b0a89180524e9e3e                                                                                                                                                                                  0.0s
 => => exporting attestation manifest sha256:84b35dc005cf6531f894ec39ac25c6fe99e3523a26c7f17ce83f75a2be56e429                                                                                                                                                                    0.0s
 => => exporting manifest list sha256:d1c74b350d7638395299e725db15db9a0e156597e0690cb5a7dd269ccc78d038                                                                                                                                                                           0.0s
 => => naming to docker.io/library/topo-welcome-app:latest                                                                                                                                                                                                                       0.0s
 => => unpacking to docker.io/library/topo-welcome-app:latest                                                                                                                                                                                                                    0.0s
 => resolving provenance for metadata file                                                                                                                                                                                                                                       0.0s
```
the reason behind this is because docker detects that pipe and choose plain progress output since it's not detected as a terminal.

We don't need it, but the format would change. I put it in this version to open discussion

muchzill4:
> With this approach, we still have problems where if there are no logs from that header, the header will still print.

I think that's ok. 👍 The current header visual formatting makes us expect output below it, but we could re-think styling it differently, so that it could be treated as both "step title" (for deploy) and "section title" (for health).

yejseo01:
> > With this approach, we still have problems where if there are no logs from that header, the header will still print.
> 
> I think that's ok. 👍 The current header visual formatting makes us expect output below it, but we could re-think styling it differently, so that it could be treated as both "step title" (for deploy) and "section title" (for health).

Ah in that case, I'll proceed with the simpler `PrintFirstHeader` and `PrintNthHeader`

yejseo01:
https://github.com/arm/topo/pull/467

PR #462, opened 2026-09-14
feat: add `--env-file` to `ps`, `deploy` and `stop`, read env vars from `.env.topo` by default

awphi:
## Changes

This is the first change in moving to an env-file based approach for configuring Topo projects.

- Add `--env-file` to `ps`, `deploy` and `stop` - all the commands that either `project.Read` the compose file via compose-go or handle it with `docker|podman compose` invocations. This serves the same function as compose `--env-file` which allows the [interpolation](https://docs.docker.com/reference/compose-file/interpolation/) of env vars specified in the file(s) anywhere in the compose project.
- By default, the env files we read are `<compose.yaml dir>/.env` and `<compose.yaml dir>/.env.topo`, mimicking `docker compose`
- Any files specified via `--env-file` are resolved relative to cwd, rather than the compose file(s). This mimicks docker compose once again.
- Any `--env-file` specified that does not exist we error on. However, if none are specified then we just _try_ to load them silently. Once again, you guessed it, this mimicks docker compose.

Breaking change to follow which moves `topo configure`/`topo clone` over to writing to `.env.topo` (or a user-specified env file) to complete this. Likely will land as 2-3 PRs.

## Checklist


cmd/topo/root.go
```diff
@@ -180,6 +181,34 @@ func resolveOutput(cmd *cobra.Command) term.Format {
 	return term.Plain
 }
 
+const envFileFlag = "env-file"
+
+func addEnvFileFlag(cmd *cobra.Command) {
+	cmd.Flags().StringArray(
+		envFileFlag, []string{".env", env.DefaultFilename},
+		"path to an env file to source values for compose interpolation. can be specified multiple times",
```
muchzill4: - Should also specify what the defaults are in the help text unless cobra does it by default
- Is this additive or does it override defaults?

internal/env/file.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+const DefaultFilename = ".env.topo"
+
+func ResolveFiles(root string, files []string, allowMissing bool) ([]string, error) {
```
muchzill4:
```suggestion
func ResolveFiles(root string, files []string, skipMissing bool) ([]string, error) {
```

`allowMissing` sounds odd.

internal/deploy/podman/command_test.go
```diff
@@ -30,11 +30,11 @@ func TestCommand(t *testing.T) {
 
 func TestComposeCommand(t *testing.T) {
 	t.Run("sets args and env", func(t *testing.T) {
-		scope := project.Scope{ComposeFile: "compose.yaml", Env: []string{"FOO=BAR"}}
+		scope := project.Scope{ComposeFile: "compose.yaml", Env: []string{"FOO=BAR"}, EnvFiles: []string{".foobar.env", ".env"}}
 		command, err := podman.ComposeCommand(context.Background(), podman.NewSocket("tcp://127.0.0.1:12345"), scope, "up", "-d")
 
 		require.NoError(t, err)
-		assert.Equal(t, []string{"podman", "compose", "-f", "compose.yaml", "up", "-d"}, command.Args)
+		assert.Equal(t, []string{"podman", "compose", "-f", "compose.yaml", "--env-file", ".foobar.env", "--env-file", ".env", "up", "-d"}, command.Args)
```
muchzill4: Did you check if this actually computes? I hope it does, but who knows..

internal/project/scope.go
```diff
@@ -1,13 +1,16 @@
...
+)
 
 type Scope struct {
 	ComposeFile string
 	Env         []string
+	EnvFiles    []string
 }
 
-func LoadScope(composeFile string, target string) (Scope, error) {
+func LoadScope(composeFile string, target string, envFiles []string) (Scope, error) {
```
muchzill4: I start to wonder if this is a misnomer. It doesn't `Load` anything. All it does is build some target specific env variables. 🤔 

`BuildScope`?

internal/deploy/docker/command_test.go
```diff
@@ -25,11 +25,11 @@ func TestComposeCommand(t *testing.T) {
 	t.Run("builds docker compose command for remote host", func(t *testing.T) {
```
muchzill4: Wasn't paying attention and this test has ballooned in responsibility. Description no longer accurate. I wonder if we can split it into units. Same with mirror in podman.

internal/project/read_test.go
muchzill4:
```suggestion
	t.Run("inherits environment variables from custom dotenv file", func(t *testing.T) {
```

🤔

cmd/topo/root.go
awphi: - Should also specify what the defaults are in the help text unless cobra does it by default

Cobra adds it, can flesh out if you think it's unclear:

```
Flags:
      --env-file stringArray     path to an env file to source values for compose interpolation. can be specified multiple times (default [.env,.env.topo])
```

- Is this additive or does it override defaults?

Overrides the defaults, matching the behaviour `docker compose`

internal/deploy/podman/command_test.go
awphi: Computes as is in is executable? Didn't test this exact command but lots of `topo deploy --env-file ...` didn't reveal any issues with it

cmd/topo/root.go
muchzill4: Wait, there's a dot there, yet it starts with lowercase.

internal/deploy/podman/command_test.go
muchzill4: I meant does `podman + docker-compose` work as intended. Again, I assume yes, but worth checking..

internal/project/read_test.go
awphi: https://github.com/arm/topo/pull/462/commits/060889966c2b5e97dbd689dbc2bff8e9f6c30814

internal/deploy/docker/command_test.go
awphi: https://github.com/arm/topo/pull/462/commits/962ca351aea705e7ef33fbddb1cfc73bf9175c61

internal/project/scope.go
awphi: Intention was to but env file resolution in here but that felt awkward. Also wanted to capture it does hostname resolution via `ssh -G` and felt `Load` implied a bigger cost than something like `New`. `Build` seems like a good compromise: https://github.com/arm/topo/pull/462/commits/b86c185f910067e5ba3e0c2b8acbcf5745c7b2a9

internal/env/file.go
awphi: https://github.com/arm/topo/pull/462/commits/b29951c3faf196d6d65575a25092f456b8def016

internal/deploy/podman/command_test.go
awphi: Ah right, good point. Seems to be working fine for me on podman 6.1.0 w/ docker-compose 5.5.0!

cmd/topo/root.go
awphi: argh, we don't sentence case our help strings but this very much feels like a two sentence description. how do you feel about a semi-colon? https://github.com/arm/topo/pull/462/commits/be2dfc7bb8b72183e9cab57a97e5d8bbc62738d5

internal/deploy/podman/command_test.go
muchzill4: Devils advocate - do we need to say can be specified multiple times? 🤔 If it shares `docker compose` semantics..

muchzill4 (approved):

internal/deploy/podman/command_test.go
awphi: Works for me! It is dead https://github.com/arm/topo/pull/462/commits/d0be5a97a8736e35c4c773a5a5e02154075221e2

PR #465, opened 2026-09-15
feat: experimental env-file migration

awphi:
## Changes

Adds a `TOPO_EXPERIMENTAL_FEATURES`-gated `--migrate` flag to `topo configure`. This migrates Topo Projects from in-file storage of parameter values like this: 
```yaml
# compose.yaml
services:
  welcome-service:
    platform: linux/arm64
    build:
      context: .
      args: 
        ARG: value
```

to this:
```yaml
# compose.yaml
services:
  welcome-service:
    platform: linux/arm64
    build:
      context: .
      args: 
        ARG: ${ARG}
```
```text
# .env.topo
ARG=value
```

- It is provided as a convenience which we will point to from a check in `Configure` in a later PR. This is when we will de-experimental it too + add any necessary docs around the new `configure` mechanism too.
- It uses existing parameterization primitives like `ApplyParameterValues` which we could otherwise delete :sob: - hence the purely additive change. For this reason it's temporary and I've made a ticket to remove this flag and its orphaned primitives in a future release.
- A check for `.env.topo` before allowing a migration is included as a failsafe for running `topo configure --migrate` twice in a row.
- It's purposefully stupid i.e. if you have an arg like ARG: ${SOME_ENV_VAR} where ARG is a topo parameter - your env file will contain ARG=${SOME_ENV_VAR}. This is because build args could be complex constructions of multiple interpolations like ARG: ${ARG?required}-static-string-${SOME_OTHER_ENV_VAR:-default value} and trying to get clever about this sounds like a recipe for disaster. `docker-compose` env var interpolation can go arbitrarily deep so this keeps the migrated project faithful to the source project behaviourally

**Keen to hear opinions on if the juice is worth the squeeze here in terms of adding a migration at all.**

## Checklist


muchzill4:
I'm not sure migration is worth it. However, it'd be nice to provide a warning when user runs `configure` and the produced configuration won't be relevant, because project isn't leveraging defined topo parameters. This has a potential to be universally useful.

awphi:
> I'm not sure migration is worth it. However, it'd be nice to provide a warning when user runs `configure` and the produced configuration won't be relevant, because project isn't leveraging defined topo parameters. This has a potential to be universally useful.

Agreed. I had this warning down as something to include regardless of migration. Intention was for a warning along the lines of "argh it's broken - go run `topo configure --migrate-to-env`" if _none_ of your configured parameters were used and give a less prescriptive warning if only _some_ were used. However +300 LoC and the inability to delete a couple more hundred is kind of a shame, especially when we can write a pretty good warning on how to fix (or how a user's agent can fix it 😅 )

awphi:
@muchzill4 for context: after some further debate we concluded that this change is probably the least work of all the options and is more than good enough for a best-effort migration we expect you to run once.

Planning to cordon off this behaviour in a `migrate` package under `project` in a later PR such that deleting it will be easy.

internal/env/file.go
```diff
@@ -26,3 +42,30 @@ func ResolveFiles(root string, files []string, skipMissing bool) ([]string, erro
 	}
 	return paths, nil
 }
+
+func SaveFile(path string, values map[string]string) error {
```
th3james:
```suggestion
func WriteFile(path string, values map[string]string) error {
```
"Save" to me suggests something pre-existing we're updating. `WriteFile` is consistent with Go

internal/project/configure.go
```diff
@@ -18,10 +21,57 @@ func Configure(composeFilePath string, resolver parameter.Resolver) error {
...
 	}
 
-	return applyParameterValues(composeFilePath, values)
+	return applyParameterValuesToComposeFile(composeFilePath, values)
 }
 
-func applyParameterValues(composeFilePath string, values parameter.Values) error {
+func MigrateToEnv(composeFilePath string) error {
+	projectDir := filepath.Dir(composeFilePath)
+	envFile := filepath.Join(projectDir, env.DefaultFilename)
```
th3james:
```suggestion
	envFilePath := filepath.Join(projectDir, env.DefaultFilename)
```
nitpick, but I prefer distinguishing from paths and files. This is particularly beneficial at call sites like this: 
```go
err = env.SaveFile(envFilepath, values)
```
because it really clarifies the behaviour of the function, whereas `SaveFile(envFile, values)` could be interpreted as writing to a file handle.

awphi: https://github.com/arm/topo/pull/465/commits/f83c2281b0c069cc75ba03313db799fa24cfd149

internal/env/file.go
awphi: https://github.com/arm/topo/pull/465/commits/f83c2281b0c069cc75ba03313db799fa24cfd149

th3james (approved):

PR #467, opened 2026-09-15
fix: consistent newlines for cli outputs

yejseo01:
## Changes

  - Added distinct first and subsequent header functions.
  - Removed the leading blank line before first headers.
  - Updated all header callers.
  - Replaced ┌─ with the neutral ── divider.
  - Updated and passed affected tests.

## Checklist


internal/operation/sequence.go
```diff
@@ -13,12 +13,17 @@ func NewSequence(operations ...Operation) Sequence {
...
 		if cmdOutput != nil {
-			err := term.PrintHeader(cmdOutput, op.Description())
+			description := op.Description()
+			err := printHeader(cmdOutput, description)
 			if err != nil {
 				return err
 			}
+			if description != "" {
+				printHeader = term.PrintNthHeader
+			}
```
muchzill4, line 26: Slightly confused about this. Do we ever have empty descriptions? It'd be odd to print `-- -----------`, no?

I think we should printNthHeader for indices >0.

muchzill4 (approved):
Nice!

PR #468, opened 2026-09-15
refactor: make health checks graph-backed

muchzill4:
## Changes

- Centralise host and target health dependencies in a shared graph.
    - Evaluates prerequisites once and reuse results across health checks.
    - It establishes a foundation which will enable grouping checks by functionality and evaluating them cheaply in one shot.
- Update the health command for graph-backed reporting.

## Checklist


internal/health/dependency_graph.go
```diff
@@ -0,0 +1,133 @@
...
+	Target            *ssh.Destination
+	SkipVersionChecks bool
+	AcceptHostKeys    bool
+}
+
+type DependencyGraph struct {
+	Registry *DependencyRegistry
+	Host     []DependencyID
+	Target   []DependencyID
+}
```
awphi: Are the target and host dependencies not two distinct graphs, or do we explicitly want to keep this avenue open?

internal/health/dependency_graph.go
```diff
@@ -0,0 +1,133 @@
...
+	for _, id := range dependencies {
+		dependency := g.Registry.dependency(id).dependency
+		if !prerequisitesFulfilled(ctx, g.Registry, dependency.Prerequisites) {
+			continue
+		}
+		statuses = append(statuses, DependencyStatus{
+			ID:     dependency.ID,
+			Label:  dependency.Label,
+			Result: g.Registry.Check(ctx, id),
+		})
```
awphi: Does it make sense to require callers of `registry.Check` to have to check prereqs manually? Can't think of a scenario where you'd want to run the check without validating its prereqs first?

I guess the thing I'm getting at is that it's a little confusing to have `DependencyGraph` and `DependencyRegistry` and have evaluation responsibilities split between the two. Feels like `DependencyGraph` adds very little and we should just have two `DependencyRegistry`s (host/target) that can be evaluated separately - much like `DependencyGraph.Evaluate` does atm

awphi (commented):
Some Qs but do tell me to go away if there's any argument against what I suggest that I've overlooked!

internal/health/dependency_graph.go
muchzill4: https://github.com/arm/topo/pull/468/commits/2ae4854ecc7132678b8ca6ed748633b11b5c3ad4

internal/health/dependency_graph.go
muchzill4: https://github.com/arm/topo/pull/468/commits/f2073af1d83724defe129a157c0b2946f40badfa

awphi (approved):

PR #470, opened 2026-09-16
refactor: report missing target as a connectivity warning

muchzill4:
## Changes

- Represent a missing target as a Connectivity dependency warning with a corrective hint.
- Always include target health results, moving the missing-target message out of the view layer.
- Preserve target report JSON fields when a destination is specified.

Screenshot 2026-09-16 at 17 11 27

## Checklist


internal/health/dependencies.go
```diff
@@ -222,12 +222,20 @@ func NewDependencyOnDockerCompose(r runner.Runner, prerequisites ...DependencyID
...
+func NewConnectivityDependency(target *ssh.Destination, acceptNewHostKeys bool, missingTargetFixMessage string) Dependency {
 	return Dependency{
 		ID:    DependencyIDConnectivity,
 		Label: "Connectivity",
 		Check: func(ctx context.Context) DependencyCheckResult {
+			if target == nil {
+				failure := &DependencyCheckFailure{Severity: SeverityWarning, Message: "target not specified"}
+				if missingTargetFixMessage != "" {
+					failure.Fix = &Fix{Description: missingTargetFixMessage}
+				}
```
awphi, line 234: Could we just hard-code the fix message here instead of passing it down? Not sure it needs any context from anything up the stack

internal/health/health.go
```diff
@@ -40,15 +40,14 @@ func Check(ctx context.Context, options HealthCheckOptions) HealthReport {
...
 	targetReport := TargetReport{
-		Destination:  options.Target.String(),
-		IsLocalhost:  options.Target.IsPlainLocalhost(),
 		Dependencies: toDependencyReports(evaluatedHealthCheck.Target),
 	}
+	if options.Target != nil {
+		// Legacy JSON support
+		targetReport.Destination = options.Target.String()
+		targetReport.IsLocalhost = options.Target.IsPlainLocalhost()
+	}
```
awphi, line 50: Pondering on if this makes this a breaking change or not. We've changed the optionality of the `target` field, making it always returned by `topo health`. It's explicitly not breaking under our policy:

> Making a previously optional/nullable field required/non-nullable

but feels like it wouldn't have been unreasonable for a consumer to use the presence of `target` in the output to determine if a target was specified.

awphi: In the same vein, I question the utility of returning the `target` object when once isn't specified:

```
  "target": {
    "destination": "",
    "isLocalhost": false,
    "connectivity": {
      "name": "Connectivity",
      "status": "warning",
      "value": "target not specified",
      "fix": {
        "description": "provide --target or set TOPO_TARGET to check target health"
      }
    },
    "dependencies": [],
    "processingDomainDriver": {
      "name": "Processing Domain Driver (remoteproc)",
      "status": "",
      "value": ""
    }
  }
```

I suppose the content of `connectivity` is useful but the rest is thoroughly useless.

internal/health/dependencies.go
muchzill4: AFAIR the original rationale behind this (this is not new here!) was that cli layer owns the missing target fix message. In other words, the cli knows that we need to add `--target` kwarg. Granted, this historically isn't consistent with something like shim, that hardcodes the command.

internal/health/health.go
muchzill4: Nice catch! https://github.com/arm/topo/pull/470/commits/ec13efb3a32d71eaf660c65a7117b2fef9dd5bca

awphi (approved):
Ell gee tee em

PR #471, opened 2026-09-17
refactor: compose health check prerequisites from registered nodes

muchzill4:
## Changes

- Register health dependencies with direct prerequisite-node references instead of dependency ID lookups.
- Assemble host and target health dependencies inline in the health check constructor.
- Preserve legacy dependency IDs only where required for JSON output.
- Update health dependency and evaluation tests.

## Checklist


internal/health/dependencies.go
```diff
@@ -26,10 +26,10 @@ const (
 )
 
 type Dependency struct {
-	ID            DependencyID
-	Label         string
-	Check         DependencyCheckFn
-	Prerequisites []DependencyID
+	Label string
+	Check DependencyCheckFn
+	// Used to maintain legacy JSON ouptut
```
Luke-Parkin:
```suggestion
	// Used to maintain legacy JSON output
```

internal/health/dependency_registry.go
```diff
@@ -2,55 +2,58 @@ package health
...
 	})
-	return dependency.result
+	return node.result
 }
 
-func (r *DependencyRegistry) dependency(id DependencyID) *dependencyNode {
-	dependency, exists := r.dependencies[id]
-	if !exists {
-		panic(fmt.Sprintf("health dependency not registered: %q", id))
+func (r *DependencyRegistry) dependency(node *DependencyNode) *DependencyNode {
```
th3james: I find this confusing it its current form. I think the call sites for this function become much easier to understand with a simpler, more explicit:
```suggestion
func (r *DependencyRegistry) assertRegistered(node *DependencyNode) {
```
and ditching the chaining in favour of:
internal/health/health_check.go:83-84
```go
	for _, dependency := range dependencies {
		h.Registry.assertRegistered(dependency)
```

internal/health/dependency_registry.go
```diff
@@ -2,55 +2,58 @@ package health
...
+func (r *DependencyRegistry) Check(ctx context.Context, node *DependencyNode) (DependencyCheckResult, bool) {
+	r.dependency(node)
+	for _, prerequisite := range node.prerequisites {
+		prerequisiteResult, hasUnmetPrerequisites := r.Check(ctx, prerequisite)
 		if hasUnmetPrerequisites || prerequisiteResult.Failure != nil {
 			return DependencyCheckResult{}, true
 		}
 	}
-	return r.checkDependency(ctx, dependency), false
+	return r.checkDependency(ctx, node), false
```
th3james: nitpick but IMHO this boolean is back to front. Go lacks a mechanism for attributing a name to the meaning of the return value, and I find it a bit surprising that `true` means "I did no work".

I'd invert it so the call sites become:
internal/health/health_check.go:86-89
```go
		result, hasPrerequisites := h.Registry.Check(ctx, node)
		if !hasPrerequisites {
			continue
		}
```
or even
```go
		result, checked := h.Registry.Check(ctx, node)
		if !checked {
			continue
		}
```

awphi (approved):
This makes a lot more sense to me now. I would even consider renaming `DependencyRegistry` -> `DependencyGraph`, although I guess it's really a multigraph of disparate components but 🤷

th3james (approved):

internal/health/dependency_registry.go
muchzill4: https://github.com/arm/topo/pull/471/commits/bdc0d6611de0bfdcb35afbb2410d8aeee533d934

internal/health/dependency_registry.go
muchzill4: https://github.com/arm/topo/pull/471/commits/c485ef4da8055d94c8f20f670cb29f9a3c11d6ab

PR #472, opened 2026-09-17
feat: experimental --migrate-to-env support for `topo clone`

awphi:
## Changes

When testing the new `topo configure` flow (which still isn't implemented here) I realized we needed to allow users to continue to `topo clone` legacy-style projects as otherwise we'd attempt to configure them, see no referenced env vars and bail out. 

- Adds a `TOPO_EXPERIMENTAL_FEATURES`-gated `--migrate-to-env` flag to `topo clone`, to match configure
- Performs the migration _after_ cloning and paramaterizing the project with the old flow. When we change `Configure` to use the new flow, we must perform the migration first. This change exists just to get the flag + plumbing in place now. 

Open to other approaches. I considered skipping configuration altogether if a legacy project is detected and requiring the user to `topo configure --migrate-to-env` manually - felt like too much manual effort and gives the user an unconfigured project on disk which may not be deployable.

## Checklist


muchzill4:
How do I know if I need to use `--migrate-to-env` w/ `topo clone`? This feels a bit like a chicken and egg problem.

Shouldn't it always auto-migrate? 🤔

awphi:
> How do I know if I need to use `--migrate-to-env` w/ `topo clone`? This feels a bit like a chicken and egg problem.
> 
> Shouldn't it always auto-migrate? 🤔

Plan is you run `topo clone legacy-project`, it attempts to configure it, spots it's a legacy projects, bails out and tells you to re-run with this flag.

Open to making both `topo clone` and `topo configure` auto-migrate when they see fit, just feels a bit less intentional but maybe the convenience is worth it?

muchzill4 (approved):

PR #475, opened 2026-09-17
feat: section health output by checked functionality

muchzill4:
## Changes

- Split health checks into deployment and project discovery readiness sections.
   - Treat `--target` omission differently for each section.
- Preserve legacy JSON health-report output.

Screenshot 2026-09-17 at 15 30 36

## Checklist


internal/health/health_check.go
```diff
@@ -39,14 +49,40 @@ func NewHealthCheck(options HealthCheckOptions) HealthCheck {
...
+			options.AcceptHostKeys,
+			"target not specified",
+			SeverityError,
+			options.MissingTargetFixMessage,
+		))
+		projectDiscoveryConnectivity := registry.Register(NewConnectivityDependency(
+			nil,
+			options.AcceptHostKeys,
+			"target not specified; cannot calculate project compatibility",
+			SeverityWarning,
```
Luke-Parkin, line 70: Info? It's entirely reasonable to not set --target, no?
Not sure.

internal/output/term/header.go
```diff
@@ -33,8 +34,8 @@ func Header(description string, isTTY bool) string {
 	prefix := "── "
 	suffix := " "
 
-	descriptionWidth := len(description)
-	barWidth := max(totalWidth-len(prefix)-descriptionWidth-len(suffix), 0)
+	descriptionWidth := utf8.RuneCountInString(description)
```
Luke-Parkin, line 37: What's the difference here? Why this instead of len()?

(I see the test below checks, what's the diff?)

internal/output/term/term_test.go
```diff
@@ -100,13 +100,17 @@ func TestPrintHeader(t *testing.T) {
...
-		prefix := "── "
-		suffix := " "
-		barWidth := totalWidth - len(prefix) - len("Hello") - len(suffix)
-		expected := "\n" + prefix + "Hello" + suffix + strings.Repeat("─", barWidth) + "\n"
+		want := "\n── Hello ───────────────────────────────────────────────────\n"
+		assert.Equal(t, want, buf.String())
+	})
 
-		assert.Equal(t, expected, buf.String())
+	t.Run("counts unicode descriptions by runes", func(t *testing.T) {
```
Luke-Parkin: Not sure this adequately describes the test?

It's counting the Unicode descriptions as part of it, but comparing those strings doesnt a exclusively verify that, does it?

internal/output/views/health.go
```diff
@@ -117,25 +205,25 @@ type jsonFix struct {
 	Command     string `json:"command,omitempty"`
 }
 
-func toJSONHealthReport(report HealthReport) jsonHealthReport {
+func toJSONHealthReport(report legacyHealthReport) jsonHealthReport {
```
Luke-Parkin, line 208: Are we migrating this in the future?

Luke-Parkin (commented):
LGTM just some Q's and one thing to ponder on.

internal/output/views/health.go
muchzill4: This is here only to avoid breaking change with json output. Once I'm done `topo health --output json` will have different shape.

internal/output/term/header.go
muchzill4: | Symbol | `len` bytes | Runes | Terminal cells |
|---|---:|---:|---:|
| `✓` | 3 | 1 | 1 |
| `✗` | 3 | 1 | 1 |
| `─` | 3 | 1 | 1 |
| `!` | 1 | 1 | 1 |

internal/output/term/term_test.go
muchzill4: https://github.com/arm/topo/pull/475/commits/3fbed0e0d03af40f2b237655e3a843c05c96f54f

internal/health/health_check.go
muchzill4: I thought warnings were for non blocking issues that make topo less functional? 🤔 Can be swayed either way here. I can't find that document anymore.

internal/output/term/header.go
Luke-Parkin: Makes sens

internal/health/health_check.go
Luke-Parkin: fair

Luke-Parkin (approved):

PR #476, opened 2026-09-17
feat!: env-file based configure/clone

awphi:
> Back again with another big PR :( Majority of the new code is tests/migration support. We can do a big ol' cleanup in a month or two once a few versions with the migration capability have gone out.

## Changes

- De-experimentalizes `--migrate-to-env` on `topo clone` and `topo configure`
- **Changes the `project.Configure` flow to read from/write to the default `.env.topo` file for current values/setting new values.**
  - Documents this new flow with a small migration guide + update in-docs examples
  - Moves the `topo clone` dotenv migration to happen before configuration now that it uses this new flow
  - Prompts users to use the migration flag when attempting to clone a legacy project
- Fixes an issue with the migration path where args like `FOOBAR: 2` would be set to `FOOBAR: !!int ${FOOBAR}` by ensuring migrated value nodes are explicitly marked as strings

Not blocked by (thanks to `--migrate-to-env`), but would be good to merge after: https://github.com/arm/topo-project-catalog/pull/15

## Checklist


internal/project/parameters_test.go
awphi, line 1: This is mostly a duplicate of `project/definition.go` as the intention is this thinner module stays in the code as  the intended way to read param definitions from a root compose file and `project/definition.go` remains only for migration. It will be moved into a separate `migration` package in a follow-up.

internal/project/parameters.go
```diff
@@ -0,0 +1,61 @@
...
+		values := []string{}
+		if val, ok := currentValues[name]; ok {
+			values = append(values, val)
+		}
+		definitions = append(definitions, parameter.Definition{
+			Name:          name,
+			Description:   param.Description,
+			Required:      param.Required,
+			Example:       param.Example,
+			CurrentValues: values,
```
awphi, line 41: I will remove this in a follow-up, replacing it with a second arg in the parameter resolver chains containing current values. Another advantage of dotenv is params can only ever have one value at a time - not multiple!

cmd/topo/configure.go
awphi, line 1: Third and final thing I'll be adding in a follow-up; ability to configure the dotenv file `configure`/`clone` write to via a flag

internal/project/parameters.go
awphi: #495

internal/project/parameters_test.go
awphi: #493

docs/project-specification/02-project-configuration.md
```diff
@@ -59,6 +61,35 @@ RUN test -n "$GREETING" || (echo "ERROR: GREETING project parameter is required"
...
+```
+
+Topo saves the value in `.env.topo`. The Compose file continues to reference `${GREETING:-Hello, World}`.
+To use the saved values with Docker Compose directly, specify the environment file:
+
+```sh
+docker compose --env-file .env.topo up --build
+```
+
+Parameter names must match their environment variable references. If the project defines parameters but none are referenced, `topo configure` rejects the project as using the legacy format.
```
muchzill4: Is this true? We don't reject, but warn?

https://github.com/arm/topo/pull/476/changes#diff-b54ec861a180f629acaf06a84f372f7ce25249ffa27395bb423940de98464399R140

internal/project/configure.go
```diff
@@ -3,16 +3,63 @@ package project
...
+
+	buildArgs := make(map[string]any)
+	for name, values := range project.currentParameterValues {
+		buildArgs[name] = strings.Join(values, "\n")
+	}
+	references := template.ExtractVariables(buildArgs, nil)
+	hasParameterValues := false
+	for _, param := range project.Metadata.Parameters {
+		if _, referenced := references[param.Name]; referenced {
+			return false, nil
```
muchzill4: Potentially false positive for partially migrated projects.

internal/project/clone_test.go
```diff
@@ -55,7 +55,7 @@ services:
 		assert.FileExists(t, composeFilePath)
 	})
 
-	t.Run("preserves current build arg values", func(t *testing.T) {
+	t.Run("preserves current env values and leaves Compose unchanged", func(t *testing.T) {
```
muchzill4:
```suggestion
	t.Run("preserves current env values and leaves compose file unchanged", func(t *testing.T) {
```

e2e/configure_test.go
```diff
@@ -13,60 +14,63 @@ import (
 func TestConfigure(t *testing.T) {
 	topo := buildBinary(t)
 
-	t.Run("updates compose yaml parameters", func(t *testing.T) {
+	t.Run("writes parameters to env without changing compose yaml", func(t *testing.T) {
```
muchzill4: Perhaps we can drop the "without changing compose yaml" bit from this test. It guards against something we're dropping anyway. In the new reality it makes no sense to guard against it.

e2e/configure_test.go
```diff
@@ -13,60 +14,63 @@ import (
...
 		cmd := exec.Command(topo, "configure", "GREETING_NAME=Yml")
 		cmd.Dir = projectDir
 		out, err := cmd.CombinedOutput()
 
 		require.NoErrorf(t, err, "configure failed: %s", out)
 		assert.Empty(t, string(out))
-		want := configurableCompose("Yml")
-		got := testutil.RequireReadFile(t, composePath)
-		assert.YAMLEq(t, want, got)
+		assert.Equal(t, original, testutil.RequireReadFile(t, composePath))
```
muchzill4: Unnecessary assertion imo.

e2e/configure_test.go
```diff
@@ -13,60 +14,63 @@ import (
...
+		cmd := exec.Command(topo, "configure", "-f", "custom-compose.yaml", "CUSTOM_NAME=Custom")
 		cmd.Dir = projectDir
 		out, err := cmd.CombinedOutput()
 
 		require.NoErrorf(t, err, "configure failed: %s", out)
 		assert.Empty(t, string(out))
-		want := configurableCompose("Custom")
-		got := testutil.RequireReadFile(t, customComposePath)
-		assert.YAMLEq(t, want, got)
+		assert.Equal(t, customCompose, testutil.RequireReadFile(t, customComposePath))
```
muchzill4: Could drop this and the one below.

e2e/configure_test.go
```diff
@@ -76,21 +80,22 @@ func TestConfigure(t *testing.T) {
 		assert.Contains(t, string(out), "unknown parameter: UNKNOWN")
 		got := testutil.RequireReadFile(t, composePath)
 		assert.Equal(t, original, got)
```
muchzill4: Could drop the "changing compose" bit too.

internal/compose/yaml.go
```diff
@@ -94,6 +94,7 @@ func applyArgsMappingNode(args *yaml.Node, values map[string]string, used map[st
 		for param, value := range values {
 			if key.Value == param {
 				argValue.Value = value
+				argValue.Tag = "!!str"
```
muchzill4: What does this do?

internal/project/clone_test.go
```diff
@@ -101,12 +104,33 @@ x-topo:
...
+        GREETING: hello
+x-topo:
+  parameters:
+    GREETING: {}
+`)
+
+		err := project.Clone(t.Output(), destDir, source, parameter.NewStrictResolverChain(), false)
+
+		require.Error(t, err)
+		assert.ErrorContains(t, err, "Try cloning again with '--migrate-to-env'")
```
muchzill4: Alarm bells: unidiomatic golang with uppercase in errors. Also, not a fan of surfacing cli layer argument downstack, but we suffer from that throughout already.

internal/project/configure.go
```diff
@@ -21,7 +68,11 @@ func Configure(composeFilePath string, resolver parameter.Resolver) error {
 		return nil
 	}
 
-	return applyParameterValuesToComposeFile(composeFilePath, values)
+	if currentValues == nil {
+		currentValues = make(map[string]string)
+	}
+	maps.Copy(currentValues, values)
```
muchzill4: Why is this gymnastics required?

muchzill4 (commented):
Titanic effort!

docs/project-specification/02-project-configuration.md
awphi: That's the fallback behaviour. If _none_ are referenced then we do reject https://github.com/arm/topo/pull/476/changes#diff-b54ec861a180f629acaf06a84f372f7ce25249ffa27395bb423940de98464399R19

internal/project/configure.go
awphi: I guess we could make this stricter and only return `false` if there are no build args matching your parameter names that reference literal values. My thinking was to make this conservative to only _block_ the use of projects that we're sure don't use the new env file config and rely on the warnings to flag if you've partially migrated: https://github.com/arm/topo/pull/476/changes#diff-b54ec861a180f629acaf06a84f372f7ce25249ffa27395bb423940de98464399R19. WDYT?

internal/project/configure.go
awphi: It's not anymore 🤦 https://github.com/arm/topo/pull/476/commits/85d8c3eff32a5470c009a0815f22b82473d48dd5

internal/project/clone_test.go
awphi: https://github.com/arm/topo/pull/476/commits/051fbc495d3d801af218bc855184b2a97a8fab09 - hoisted with a bastion error. not a huge fan but smallest thing that can be done easily without breaking up `Clone`

internal/compose/yaml.go
awphi: https://github.com/arm/topo/pull/476/commits/2ef337b1d41ff67d5c4ff485b92fbcb45e006388 - the tests for this were in a non-obvious place

In English though; given the yaml build arg`FOOBAR: true`, if we attempted to override `FOOBAR` with a string like `"value"`, the current implementation would set it to `FOOBAR: !!bool "value"` which compose cannot parse. We always set values to strings so this just ensures the correct type tag.

internal/project/clone_test.go
awphi: https://github.com/arm/topo/pull/476/commits/8b12d354fcd4a99f6632bb1279114042d5cee6b1

e2e/configure_test.go
awphi: https://github.com/arm/topo/pull/476/commits/e8493a4db76b02a35801cc6142e69fb6df7bbd11

e2e/configure_test.go
awphi: https://github.com/arm/topo/pull/476/commits/e8493a4db76b02a35801cc6142e69fb6df7bbd11

e2e/configure_test.go
awphi: https://github.com/arm/topo/pull/476/commits/e8493a4db76b02a35801cc6142e69fb6df7bbd11

e2e/configure_test.go
awphi: https://github.com/arm/topo/pull/476/commits/e8493a4db76b02a35801cc6142e69fb6df7bbd11

internal/compose/yaml.go
muchzill4: Thanks for explaining. 👍

muchzill4 (approved):
🦾

PR #480, opened 2026-09-18
fix: verify Docker is reachable on remote targets

muchzill4:
## Changes

- Run `docker info` through the remote Docker connection during health checks, so unreachable or unavailable remote daemons are reported.
- Consolidate command failure formatting into a shared error type.

## Checklist


internal/health/health_check.go
```diff
@@ -85,8 +85,8 @@ func NewHealthCheck(options HealthCheckOptions) HealthCheck {
 		projectDiscoveryTargetDependencies = append(projectDiscoveryTargetDependencies, dependencyConnectivity)
 	}
 	if options.Target != nil {
+		dependencyDocker := registry.Register(NewDependencyOnRemoteDocker(*options.Target), targetPrerequisites...)
```
awphi, line 88: This also the pre-req of docker CLI being available on the host

muchzill4: What?

awphi: Sorry typo: This also **has** the pre-req of docker CLI being available on the host

This check can't function without `docker` on the host and thus it should probably have a prerequisite as such?

awphi (approved):

PR #481, opened 2026-09-18
fix: detect Docker Compose when installed as plugin

federicobozzini:
## Changes

- docker compose can be installed as a docker plugin, this PR fixes the docker compose detection
- breaking change since it changes the json returned by the `topo health` command

## Checklist


awphi (commented):
Don't think is breaking - we don't count the values of our json objects as part of the contract

Fortunately it looks [this regression](https://github.com/arm/topo/commit/9a0c65d0845a480240bc1fb1fc0a1dbde1ca66c5) only got released in 13.0.0 so we can merge and do a quick release to patch this

federicobozzini:
I think you are right, changed to non-breaking.

e2e/health_test.go
```diff
@@ -38,6 +38,7 @@ func TestHealthCheck(t *testing.T) {
 
 		assert.Contains(t, out, " ✓ OpenSSH (ssh)")
 		assert.Contains(t, out, " ✓ Container Engine (docker)")
+		assert.Contains(t, out, " ✓ Docker Compose (docker compose)")
```
muchzill4, line 41:
```suggestion
		assert.Contains(t, out, " ✓ Docker Compose")
```

How about this? It's kinda redundant at this point. 🤷

awphi (approved):

e2e/health_test.go
federicobozzini: I prefer the simmetry with the other checks and the fact that if we were to change to `docker-compose` the test would be explicit.

PR #482, opened 2026-09-18
fix: improve messaging when it's local docker that fails remote check

muchzill4:
## Changes

I have realised that we say "on the target", while the issue is lack of "host's" `docker`.

- Improve messaging when `docker` is unavailable locally

## Checklist


federicobozzini (approved):

awphi (approved):
Looks good! Could do with a test though?

PR #483, opened 2026-09-18
fix: restore conditional -f in post-deploy ps command

th3james:
## Changes

- Fixes a regression where the post-deploy message always rendered the `-f` argument, even when it was implicit. This was caused by a change in compose file resolution: `getComposeFileName` was returning an absolute path when no flag was specified. This was then compared with the default file _name_  (not path), and deemed to be a user specified compose file.
- `getComposeFileName` is now `resolveComposeFilePath`, and it now consistently returns an absolute path in all non-error conditions.
- Success message rendering is pushed up the stack closer to the CLI interface so `cmd.Flag(composeFileFlag)` is in scope. This avoids drilling CLI interface details into the unit layer, and means that the printed message command is consistent with what the user specified.
- Does some renaming to be more explicit about the fact that variables are FilePaths, not Files or FileNames (confusing these caused this bug)

## Checklist


cmd/topo/deploy.go
```diff
@@ -53,23 +56,39 @@ By default, Topo uses compose.yaml in the current working directory, then compos
...
+		}
+		composeFileFlagValue := cmd.Flag(composeFileFlag)
+		return post_deploy.PrintDeploySuccess(
+			os.Stdout,
+			scope,
+			defaultDeploySuccessMessage(
+				strings.TrimSpace(composeFileFlagValue.Value.String()),
+				composeFileFlagValue.Changed,
+			),
+		)
```
awphi, line 91: Not a great fan of hoisting this up away from all the other header printing, makes it harder to see the order of operations at a glance.

WDYT about passing in a `DefaultSuccessMessage` into `docker|podman.Deploy` via `docker|podman.DeployOptions`?

internal/deploy/podman/deploy.go
```diff
@@ -83,17 +82,14 @@ func Deploy(ctx context.Context, output io.Writer, scope project.Scope, options
...
+	if tunnel == nil {
+		return nil
 	}
-
-	if err := term.PrintNthHeader(output, "Deployment Success"); err != nil {
+	if err := closeRemoteTunnel(tunnel); err != nil {
 		return err
 	}
-	return post_deploy.PrintDeploySuccess(output, scope, post_deploy.DefaultMessage(scope.ComposeFile))
+	tunnel = nil
```
awphi, line 91: Does nil-ing out the tunnel do anything here? It goes out of scope after function execution so should be gc'ed regardless?

th3james: It doesn't affect correctness, but it prevents evaluation of code inside the conditional in the defer block:
internal/deploy/podman/deploy.go:62-66
```go
		defer func() {
			if tunnel != nil {
				deployErr = errors.Join(deployErr, closeRemoteTunnel(tunnel))
			}
		}()
```
`closeRemoteTunnel` is actually idempotent and won't return an error, but setting `tunnel = nil` signals to the defer block that this work has been performed and need not be re-attempted

cmd/topo/deploy.go
th3james: Hmm, I might ponder this over the weekend. My initial thought was that this isn't really part of deploy, it's... post-deploy. It doesn't have a dependency on any of the deploy mechanisms, it's just reading a value from the project. Feels a bit awkward injecting those details down. But I may change my mind come Monday...

th3james: https://github.com/arm/topo/pull/483/commits/21871eef894acceed9f83654e59f5ce4149f49c1

awphi (approved):

PR #485, opened 2026-09-18
fix: support registries with engine-assigned ports

awphi:
## Changes

- Adds a small tweak to our registry container port detection to grab the external port 5000 is actually bound to, rather than simply the one that was requested. For topo-launched registries this is functionally identical but allows users (and our tests) to launch registries manually with engine-assigned ports.
- Adds a test fixture to launch a test registry with an auto-assigned port, making use of the behaviour described above. This avoids potential races ([example](https://github.com/arm/topo/actions/runs/35360096668/job/105648950496)) between us finding a free port and actually binding something to it, where some other process can swoop in and steal it.

This doesn't eliminate all port races but does (hopefully) solve a big chunk of them

## Checklist


th3james (approved):
Very nice

PR #486, opened 2026-09-18
test: health graph assembly

muchzill4:
## Changes

- Separate production check creation from graph assembly
- Add coverage for missing graph rules

## Checklist

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


internal/health/health_check.go
```diff
@@ -14,17 +14,82 @@ type HealthCheckOptions struct {
...
 	ProjectDiscovery ReadinessCheck
 }
 
+func NewHealthCheck(options HealthCheckOptions) HealthCheck {
+	return AssembleHealthCheck(options.Target, newProductionChecks(options))
+}
+
+func AssembleHealthCheck(target *ssh.Destination, checks Checks) HealthCheck {
+	registry := NewDependencyRegistry()
+	hostNodes := registerHostChecks(registry, checks)
```
Luke-Parkin: This makes more sense than before

internal/health/health_check.go
```diff
@@ -14,17 +14,82 @@ type HealthCheckOptions struct {
 	AcceptHostKeys          bool
 }
 
+type Checks struct {
```
Luke-Parkin, line 17: I would think that having this shared list would mean you can deduplicate across target and host checks, but I can see there's some checks that are 'TargetDocker' vs 'Docker'

And also others like 'Lscpu' that are target only?

Is it worth splitting this into target/host if there's no deduplication gained from keeping them together? Or Atleast tagging all of the Target ones so they start with Target perhaps?

internal/health/health_check.go
```diff
@@ -41,107 +106,100 @@ type EvaluatedDependency struct {
...
+	deployment []*DependencyNode
+	discovery  []*DependencyNode
 }
 
-func (h ReadinessCheck) Evaluate(ctx context.Context) EvaluatedReadinessCheck {
-	return EvaluatedReadinessCheck{
-		Host:   h.evaluateDependencies(ctx, h.Host),
-		Target: h.evaluateDependencies(ctx, h.Target),
+func registerHostChecks(registry *DependencyRegistry, checks Checks) hostNodes {
+	topo := registry.Register(checks.Topo)
```
Luke-Parkin, line 170: If you had all of the hostChecks in a seperate struct, would that mean you could here do something like

```
for _, element := range deploymentChecksOnHost {
     registered_deployment_checks = append(registered_checks, registry.Register(element)
}
```
then the same for discovery.
Would mean the src of truth is the structure at the top meaning it's a bit more data driven.

Luke-Parkin (approved):
This is all round improvements, I've dropped a suggestion, but I'm happy to approve for for now, will leave it up to your judgement on whether you want to implement.

internal/health/health_check.go
muchzill4: I liked the idea of logical grouping, however looping is sadly not possible, because of specific prerequisites. https://github.com/arm/topo/pull/486/commits/2a45a4bae1fef20ead272f935dcc35a2a88fa867

internal/health/health_check.go
muchzill4: https://github.com/arm/topo/pull/486/commits/2a45a4bae1fef20ead272f935dcc35a2a88fa867

PR #488, opened 2026-09-18
refactor: model absence of `--target` as shared prerequisite

muchzill4:
## Changes

- Introduce a simpler model of shared (but different) `TargetSpecified` dependency, making graph assembly simpler

## Checklist

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


internal/health/dependencies.go
```diff
@@ -198,37 +200,38 @@ func NewDependencyOnDockerCompose(r runner.Runner) Dependency {
...
 
 type ConnectivityOperations struct {
-	Authenticate    func(context.Context) error
-	KnownHostsEntry func() (string, error)
+	Authenticate    func(context.Context, ssh.Destination) error
+	KnownHostsEntry func(ssh.Destination) (string, error)
 }
 
-func NewConnectivityDependency(target ssh.Destination, operations ConnectivityOperations) Dependency {
+func NewConnectivityDependency(target *ssh.Destination, operations ConnectivityOperations) Dependency {
```
awphi: Not something that was introduced but is built on here - passing `target ssh.Destination` and this `operations ConnectivityOperations` together rubs me the wrong way. Think it's just the injection of abstract behaviour that's a bit too much DI-like + an extra level of indirection that's a little hard to follow.

Similar issue with `probeInfo` on `NewDependencyOnRemoteDocker`.

internal/health/dependencies.go
```diff
@@ -198,37 +200,38 @@ func NewDependencyOnDockerCompose(r runner.Runner) Dependency {
...
 }
 
-type MissingTargetOptions struct {
-	Message    string
-	Severity   CheckSeverity
-	FixMessage string
-}
-
-func NewMissingTargetDependency(options MissingTargetOptions) Dependency {
+func NewTargetSpecifiedDependency(target *ssh.Destination, fixMessage string) Dependency {
```
awphi: Is there any scenario where you want to use target specification as a prerequisite and not target connectivity? Trying to work out why this needs to exist as a separate node rather than folded into the existing one

internal/health/health_check.go
```diff
@@ -119,54 +118,41 @@ type EvaluatedDependency struct {
 
 func newProductionChecks(options HealthCheckOptions) Checks {
 	localRunner := runner.NewLocal()
-	checks := Checks{
+	var targetRunner runner.Runner
+	if options.Target != nil {
+		targetRunner = runner.For(*options.Target)
+	}
```
awphi: It's a shame we now allow the modelling of invalid states (like a `nil` runner here, nil-able `ssh.Desination`s being passed to checks etc.) that are only enforced at runtime via panics. Feels a little like we're forcing a CLI-level validationy check where it doesn't belong.

Luke-Parkin (approved):
Looks nice!

internal/health/health_check.go
```diff
@@ -119,54 +118,41 @@ type EvaluatedDependency struct {
 
 func newProductionChecks(options HealthCheckOptions) Checks {
```
awphi: Not something you changed here but 'production' seems out place?

internal/health/dependencies.go
muchzill4: I share your concern - happy to revisit this separately. 👍 

The original aim was to check conditional logic of checks, which w/o DI would require setting up complex state. In this case, it'd be an ssh target that can yield various errors. Alternatively we keep this untested, which is also a valid trade-off, as cost of test complexity might be too much.

awphi: I think my main concern is the dependency we're injecting doesn't feel quite right or complete, we could consider merging `target` and `operations` under a single struct. Agreed can be revisited later though.

internal/health/health_check.go
muchzill4: Any suggestions?

awphi: Just `newChecks` I guess. It's just a simple constructor and there's no alternative to the 'production' checks

internal/health/dependencies.go
muchzill4: It'd be useful for `--target localhost`, where connectivity is not applicable. Also connectivity has been doing a bit too much. Having said that, I think you're right that this probably doesn't belong here at all and shouldn't be modelled as a dependency, but graph assembly branch. Changing approach could also fix https://github.com/arm/topo/pull/488#discussion_r4060909452.

internal/health/health_check.go
muchzill4: https://github.com/arm/topo/pull/488/changes/95c2109136556b7433dab613b9eaa273fc2393d5

internal/health/dependencies.go
muchzill4: https://github.com/arm/topo/pull/488/changes/95c2109136556b7433dab613b9eaa273fc2393d5

internal/health/health_check.go
muchzill4: https://github.com/arm/topo/pull/488/changes/95c2109136556b7433dab613b9eaa273fc2393d5

internal/health/health.go
```diff
@@ -25,31 +28,64 @@ type TargetDetails struct {
...
+		targetStatus.Fix = &Fix{Description: fixMessage}
 	}
+	return targetStatus
+}
+
+func removeSuccessfulLocalhostConnectivityReports(reports []DependencyReport, target *TargetDetails) []DependencyReport {
+	return slices.DeleteFunc(reports, func(report DependencyReport) bool {
+		return report.Status == CheckStatusOK && report.ID == DependencyIDConnectivity && target.IsLocalhost
+	})
 }
```
awphi: This feels a little odd. Why would we add the connectivity check to the graph at all for `--target localhost`?

muchzill4: Good point. 🤦 https://github.com/arm/topo/pull/488/commits/e61fdb7585b8359992658853ea32543cf1b8c8d7

awphi (approved):

PR #490, opened 2026-09-20
refactor: model health dependency evaluation outcomes

muchzill4:
## Journey

The end goal is for `topo health` to distinguish a failed check from a check that could not run because another dependency failed.

This PR establishes the evaluation vocabulary needed for that journey: a dependency can execute, be blocked by a prerequisite, or be omitted because a condition does not apply. It also makes remoteproc availability a condition for its runtime checks, rather than treating it as a prerequisite.

## Changes

- Replace the checked/not-checked result with explicit executed, blocked, and omitted evaluation states.
- Record the failed prerequisites that block a dependency.
- Separate requirements that prevent execution from conditions that make a check inapplicable.
- Make remoteproc runtime checks conditional on remoteproc availability.
- Cover evaluation states, prerequisite ordering, and conditional checks with unit tests.

## Checklist


internal/health/dependency_registry.go
```diff
@@ -25,34 +46,55 @@ func NewDependencyRegistry() *DependencyRegistry {
...
-func (r *DependencyRegistry) checkDependency(ctx context.Context, node *DependencyNode) DependencyCheckResult {
-	node.once.Do(func() {
-		if node.dependency.Check != nil {
-			node.result = node.dependency.Check(ctx)
+	blockedBy := make([]*DependencyNode, 0, len(node.requirements.Prerequisites))
+	for _, prerequisite := range r.dependencies {
+		isPrerequisite := slices.Contains(node.requirements.Prerequisites, prerequisite)
+		if isPrerequisite && !r.succeeded(ctx, prerequisite) {
+			blockedBy = append(blockedBy, prerequisite)
 		}
```
awphi:
```suggestion
	for _, prerequisite := range node.requirements.Prerequisites {
		if !r.succeeded(ctx, prerequisite) {
			blockedBy = append(blockedBy, prerequisite)
}
```

No different and a bit faster afaict?

internal/health/dependency_registry.go
```diff
@@ -7,10 +7,31 @@ import (
...
+	Conditions    []*DependencyNode
+}
+
+type EvaluationState uint8
+
+const (
+	EvaluationExecuted EvaluationState = iota
+	// EvaluationBlocked means an unsuccessful prerequisite prevented execution.
+	EvaluationBlocked
+	// EvaluationOmitted means an unsuccessful condition disabled the check.
```
awphi:
```suggestion
	// EvaluationOmitted means an unmet condition disabled the check.
```

or unfulfilled or similar

awphi (approved):
LGTM w/ nits!

internal/health/dependency_registry.go
muchzill4: Different, because it'll lose the registration order. There's an explicit test for this.

internal/health/dependency_registry.go
muchzill4: I assume this applies to both comments?

awphi: Now that I have a better grip on this code I'd actually leave as is tbh. I found 'unsuccessful condition' a little awkward at first but it matches vocab used elsewhere so is probably okay.

